PR #20, opened 2025-02-25
support artifacts

zerob13:
Implemented basic artifact support. #8 

Currently supports the following artifact types:

- application/vnd.ant.code
- text/markdown
- text/html
- image/svg+xml
- application/vnd.ant.mermaid


deepinfect:
TL;Dl 
LGTM

PR #125, opened 2025-03-13
add react artifact support

zerob13:
## Pull Request Description (中文)

**你的功能请求是否与某个问题有关？请描述一下。**
使用cdn支持了react控件的渲染

**请描述你希望的解决方案**
通过直接诶导入相关的js和css库，在iframe里面单独跑了一个react组件

**桌面应用程序的 UI/UX 更改**
image


yharby:
Can we predefine something like [maplibre](https://github.com/maplibre/maplibre-gl-js) or ArcGIS Maps SDK that can interacts with dify or any other backend LLM/agent ? 
Thanks @zerob13

zerob13:
> Can we predefine something like [maplibre](https://github.com/maplibre/maplibre-gl-js) or ArcGIS Maps SDK that can interacts with dify or any other backend LLM/agent ? Thanks @zerob13

Thanks for your feedback!

At the moment, it's quite difficult to implement this capability with the current React artifact. The dependencies are injected globally via <script> tags, and we don't yet have a fully modular ESM-based React environment that allows third-party integrations. Even if we did, predefining too many libraries wouldn't be ideal, as they would consume a lot of tokens.

In the future, support for such integrations might be possible, but it should ideally be triggered explicitly by the user's prompt rather than baked in by default. For now, you could try including your desired third-party libraries through prompts in the HTML artifact, as a workaround.

PR #241, opened 2025-04-18
add font size settings and display settings | 增加字体大小调节能力

zerob13:
## Pull Request Description (中文)

**你的功能请求是否与某个问题有关？请描述一下。**
增加了一些预设的字体大小，方便用户调整ui的字体大小

**请描述你希望的解决方案**
1、设置项增加了字体滑块
2、预设几个字体大小
3、整体ui字体缩放

**桌面应用程序的 UI/UX 更改**


eric8810 (approved):
LGTM

eric8810:
The issue is caused by the following line in `DisplaySettings.vue`:

```vue
34| @update:model-value="(val) => (fontSizeLevel = val[0])"
```

The error occurs because `val` is possibly `undefined`. To resolve this issue, you can ensure that `val` is defined before accessing its first element. Here's the corrected line of code:

```vue
34| @update:model-value="(val) => { if (val && val[0] !== undefined) fontSizeLevel = val[0]; }"
```

This fix ensures that the code checks if `val` is defined and has a valid first element before updating `fontSizeLevel`. Replace the problematic line with the corrected one, and it should resolve the issue.


please fix build check

PR #325, opened 2025-05-10
Save and restore window size & position

hutchisr:
## Pull Request Description

**Is your feature request related to a problem? Please describe.**
When opening the app, the window always resets to the same size and position, which is rather small.

**Describe the solution you'd like**
Application window restores previous size and position when reopened.

**UI/UX changes for Desktop Application**
When moving or resizing the application, the size and position are saved and restored to the last size/position when the application is re-opened.

**Platform Compatibility Notes**
Unsure if this has any significant compatibility issues, I don't think it does but I will test on Windows once I have a dev environment set up (currently tested on macOS)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "electron-log": "^5.3.3",
     "electron-store": "^8.2.0",
     "electron-updater": "^6.1.7",
+    "electron-window-state": "^5.0.3",
     "fflate": "^0.8.2",
     "file-type": "^20.5.0",
     "https-proxy-agent": "^7.0.6",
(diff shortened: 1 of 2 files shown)


zerob13 (changes requested):
In this PR, there is a lack of detection for the current screen size. If the user last used the program on an external monitor and then closed it, switched to the laptop's built-in screen, and reopened the program, the software might not function properly due to recorded incorrect window size and position information. Without handling this part, it is currently not advisable to include the capability to remember window positions and sizes.  

Such scenarios are quite common in environments where people commute with their laptops.

hutchisr:
Switched to using a package with this functionality. ensures the mainWindow is visible on a display and saves/restores fullscreen & maximized state

PR #452, opened 2025-06-04
chore: adjust model-configs of deepseek & ppio

neoragex2002:
chore: adjust model-configs of deepseek & ppio


neoragex2002:
按官方平台手册参数

PR #458, opened 2025-06-05
feat(dsearch): support self-driven retrieval with confidence scoring and incremental reflection evaluation

neoragex2002:
支持自驱式检索与信心评分，改用增量型反思评估


zerob13:
```
Error: src/main/presenter/mcpPresenter/inMemoryServers/deepResearchServer.ts(397,25): error TS6133: 'iteration' is declared but its value is never read.

```

PR #483, opened 2025-06-10
fix(ui): enforce strict tab-session binding so that a session can only appear in one tab at any time and in any interaction flow

neoragex2002:
fix(ui): enforce strict tab-session binding so that a session can only appear in one tab at any time and in any interaction flow

issue 469: https://github.com/ThinkInAIXYZ/deepchat/issues/469


zerob13:
LGTM

PR #497, opened 2025-06-11
Add TogetherAI Provider

noQuli:
I added TogetherAI  as a new provider. It now fetches available models and lets users pick one to generate responses. 
How I Did It:
1.I used TogetherAI TS SDK to fetch the list of models.
2.Since TogetherAI supports the OpenAI SDK but doesn’t have a specific endpoint for fetching models, I reused the logic from openAICompatibleProvider 
3.I overwrote the fetch function to make it work with TogetherAI’s SDK while keeping the same response format as OpenAI.

### DIFF
--- a/src/main/presenter/configPresenter/providers.ts
+++ b/src/main/presenter/configPresenter/providers.ts
@@ -200,7 +200,7 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
   {
     id: 'together',
     name: 'Together',
-    apiType: 'openai',
+    apiType: 'together',
     apiKey: '',
     baseUrl: 'https://api.together.xyz/v1',
     enable: false,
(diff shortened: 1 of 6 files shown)


src/main/presenter/configPresenter/providers.ts
```diff
@@ -197,21 +197,21 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
...
-  //     apiKey: 'https://api.together.ai/settings/api-keys',
-  //     docs: 'https://docs.together.ai/docs/introduction',
-  //     models: 'https://docs.together.ai/docs/chat-models',
-  //     defaultBaseUrl: 'https://api.tohgether.xyz'
-  //   }
-  // },
+  {
+    id: 'together',
+    name: 'Together',
+    apiType: 'openai',
```
zerob13: the `apiType` needs to be set to `together`, otherwise it will still only create an OpenAIcompatibleprovider, because the creation is determined based on `apiType`.

noQuli: Yes, you're right, i changed it already

PR #685, opened 2025-08-03
fix: resolve message editing and markdown rendering issues

yyhhyyyyyy:
- Fix message editing content loss with angle brackets
2ed39d89af62fbdfb3499fcd3c261c9e
92f4916e30fa1d5c76deacf3cb6661f5

- Remove unsupported typewriterEffect prop from NodeRenderer
015e714d8ec5f61b2bfb227109267a40

### DIFF
--- a/src/renderer/src/components/markdown/MarkdownRenderer.vue
+++ b/src/renderer/src/components/markdown/MarkdownRenderer.vue
@@ -5,7 +5,6 @@
       :custom-components="nodeComponents"
       :content="content"
       @copy="$emit('copy', $event)"
-      :typewriterEffect="true"
     />
   </div>
 </template>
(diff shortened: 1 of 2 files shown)


src/renderer/src/components/markdown/MarkdownRenderer.vue
```diff
@@ -5,7 +5,6 @@
       :custom-components="nodeComponents"
       :content="content"
       @copy="$emit('copy', $event)"
-      :typewriterEffect="true"
```
zerob13: 不要移除该功能

zerob13 (changes requested):

src/renderer/src/components/markdown/MarkdownRenderer.vue
yyhhyyyyyy: 已还原

PR #694, opened 2025-08-05
fix: resolve input focus loss issue after sending messages

yyhhyyyyyy:
ensure that the focus is in the input box on the conversation page.


src/renderer/src/components/ChatInput.vue
```diff
@@ -1261,7 +1299,8 @@ defineExpose({
       console.error('Failed to append mention:', error)
       return false
     }
-  }
+  },
+  restoreFocus
```
hllshiro, line 1304: 应该在上一行，写错位置了

hllshiro (changes requested):
合并错误

PR #813, opened 2025-08-29
fix: resolve race condition in floating button drag-and-drop detection

neoragex2002:
解决悬浮按钮拖动乱飞问题

## Summary by CodeRabbit

* **Bug Fixes**
  * Dragging the floating button is more reliable and responsive: drags start immediately at threshold, concurrent timers are prevented, and click events are suppressed during a drag to avoid accidental toggles.
  * Mouse interactions now ignore non-left-button presses to prevent unintended behavior.
  * Floating chat window positioning improved: smarter side-first placement, bounds clamping, and restored visibility after drag preserve consistent placement and on-screen layout.


yyhhyyyyyy (approved):
LGTM

PR #1024, opened 2025-10-18
fix import button bug and prevent backup overwriting during import

Dw9:
- Remove incorrect startBackup call from import button click handler
  - Cancel pending backup timer before import to prevent data loss
  - Add imported conversation count to import success message
  - Update interface to return count in importFromSync result
  - Count conversations from backup database before import operations
  - Update i18n files to display imported count in all languages

## Summary by CodeRabbit

* **New Features**
  * Import completion now reports the number of conversations imported.

* **UI**
  * Import dialog behavior streamlined so the dialog opens reliably and shows contextual success counts.

* **Bug Fixes**
  * Prevents backups from being overwritten during import operations.

* **Localization**
  * Updated import-complete messages in all supported languages to include a conversation count placeholder.

### DIFF
--- a/src/main/presenter/syncPresenter/index.ts
+++ b/src/main/presenter/syncPresenter/index.ts
@@ -183,22 +183,21 @@ export class SyncPresenter implements ISyncPresenter {
 
       let importedCount = 0
       try {
-        // Count conversations in backup db first (before any modifications)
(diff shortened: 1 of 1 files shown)


zerob13:
@codex review

PR #1509, opened 2026-04-22
feat(icon): implement asynchronous icon loading to improve startup pe…

zhangmo8:
…rformance

## Release Notes

* **Performance**
  * Improved application startup time through asynchronous icon loading instead of synchronous initialization.
  * Icons now preload in the background after app mount, reducing initial load delays.

* **Reliability**
  * Enhanced error handling and logging for icon loading failures.
  * Prevents repeated loading attempts when errors occur.

### DIFF
--- a/src/renderer/settings/App.vue
+++ b/src/renderer/settings/App.vue
@@ -82,6 +82,7 @@ import { useModelStore } from '@/stores/modelStore'
 import { useOllamaStore } from '@/stores/ollamaStore'
 import { useProviderDeeplinkImportStore } from '@/stores/providerDeeplinkImport'
 import { useMcpInstallDeeplinkHandler } from '../src/lib/storeInitializer'
+import { ensureIconsLoaded } from '../src/lib/iconLoader'
 import { useFontManager } from '../src/composables/useFontManager'
 import { markStartupInteractive } from '../src/lib/startupDeferred'
 import type {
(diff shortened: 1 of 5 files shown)


zerob13:
I found three actionable issues in this PR:

1. [P1] Register icon collections before child components mount
   `ensureIconsLoaded()` now runs from `onMounted()` in `src/renderer/src/App.vue`, so `WindowSideBar` and the settings navigation render `<Icon>` before the local `lucide` / `vscode-icons` collections are registered. In that state `@iconify/vue` starts a remote `loadIcons()` request, and a later `addCollection()` does not wake that pending load, so offline or firewalled installs can show blank chrome icons until the request times out. The same startup pattern appears in `src/renderer/settings/App.vue`.

2. [P2] Allow icon loading to retry after transient import errors
   In `src/renderer/src/lib/iconLoader.ts`, the `catch` branch still sets `state.isLoaded = true`. If the first dynamic import fails during HMR or just after an auto-update, every later `ensureIconsLoaded()` / `preloadIcons()` call will short-circuit, which leaves icons unavailable for the rest of that renderer session.

3. [P2] Avoid fetching the session list twice during startup
   `src/renderer/src/App.vue` now calls `sessionStore.fetchSessions()` during `onMounted()`, while `ChatTabView` already does `await sessionStore.fetchSessions()` in its startup path. Because `/` always redirects into `ChatTabView`, this guarantees a second full `listLightweight` IPC request on launch, and the request-id guard can discard the earlier result, turning the new parallel fetch into pure extra startup work.

PR #1510, opened 2026-04-22
refactor: optimize store initialization and loading in ChatTabView

zhangmo8:
## Summary by CodeRabbit

* **Performance Improvements**
  * Application startup now loads critical components concurrently rather than sequentially, reducing initialization time.
  * Background resource loading has been optimized to prevent blocking the user interface during startup.
  * Session management now loads asynchronously in the background without delaying app readiness.

### DIFF
--- a/src/renderer/src/views/ChatTabView.vue
+++ b/src/renderer/src/views/ChatTabView.vue
@@ -101,7 +101,7 @@ const initializeRouteFromFallbackState = async () => {
 onMounted(async () => {
   startupWorkloadStore?.connect()
   console.info('[Startup][Renderer] ChatTabView critical hydration begin')
-  let shouldRefreshAgentsInDeferred = true
+  let criticalLoadPromises: Promise<void> | null = null
 
   try {
     const startupClient = createStartupClient()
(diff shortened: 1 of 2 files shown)


src/renderer/src/views/ChatTabView.vue
```diff
@@ -120,31 +120,38 @@ onMounted(async () => {
     await pageRouter.initialize({
       activeSessionId: bootstrap.activeSessionId
     })
+
+    // Start loading agents, projects, models, and ollama immediately after router init
+    // Don't block on them, they load in background while we mark interactive
+    criticalLoadPromises = Promise.allSettled([
+      agentStore.fetchAgents(),
+      projectStore.fetchProjects(),
+      modelStore.initialize(),
```
zerob13, line 129: Starting `modelStore.initialize()` here races `void initAppStores()` in `App.vue`, which initializes `providerStore` asynchronously. If `providerStore.providers` is still empty when this runs, `modelStore.initialize()` marks itself initialized after refreshing zero providers, so `NewThreadPage.resolveModel()` and `ChatStatusBar` treat the empty model set as ready and the first session creation falls through to `No model available`.

src/renderer/src/views/ChatTabView.vue
```diff
@@ -120,31 +120,38 @@ onMounted(async () => {
...
     cancelDeferredHydration = scheduleStartupDeferredTask(async () => {
       console.info('[Startup][Renderer] ChatTabView deferred hydration begin')
-      await Promise.allSettled([
-        shouldRefreshAgentsInDeferred ? agentStore.fetchAgents() : Promise.resolve(),
-        projectStore.fetchProjects(),
-        modelStore.initialize(),
-        ollamaStore.initialize()
-      ])
+      // Wait for critical loads if they're still running
+      if (criticalLoadPromises) {
```
zerob13, line 152: When `getBootstrap()` throws, `criticalLoadPromises` stays `null`, so the deferred task below has nothing left to run. In that fallback startup path we never call `projectStore.fetchProjects()`, and `modelStore.initialize()` / `ollamaStore.initialize()` also stop unless some later manual action triggers them, leaving recent projects and model data missing for the whole renderer lifetime.

zerob13 (commented):
Posting two inline findings from the review.

PR #1559, opened 2026-04-29
feat(agent): enhance agent

zhangmo8:
- Updated subagent orchestrator schema to include new operations: 'run', 'list', 'info', 'log', 'wait', 'kill'.
- Added support for background task execution with runId tracking.
- Implemented validation for required fields based on operation type.
- Introduced MutableRunState to manage task states and statuses.
- Enhanced rendering of progress and final results for subagent runs.
- Added new subagent slots: Explorer, Implementer, Reviewer.
- Updated DeepChat agent settings to enable subagents by default.
- Improved tool presenter to handle new tool result structures and provide better error handling.
- Added tests for background run operations, ensuring correct behavior for list, info, and kill commands.

## Summary by CodeRabbit

* **New Features**
  * Subagents enabled by default with three roles (explorer, implementer, reviewer).
  * Background subagent lifecycle: start, list, info, wait, kill operations.
  * Standardized tool result envelope for consistent success/error/truncation semantics.

* **Improvements**
  * Parallel execution for read-only tools; mutating/write operations remain serialized.
  * Session-level tool-profile caching and richer system prompt with permission/verification sections.
  * UI warning for chat-model fallback capability; added localized warning strings.

* **Documentation**
  * New agent-refactor roadmap, spec, tasks, and tool-result envelope docs.

* **Tests**
  * Added tests for subagent lifecycle, concurrency/ordering, prompt ordering, and tool-result handling.


src/main/presenter/agentRuntimePresenter/index.ts
```diff
@@ -4038,18 +4188,80 @@ export class AgentRuntimePresenter implements IAgentImplementation {
...
     try {
-      return await this.toolPresenter.getAllToolDefinitions({
+      const profile = await this.resolveToolProfile(sessionId, projectDir, activeSkillNamesOverride)
+      const cachedProfile = this.toolProfileCache.get(sessionId)
+      if (
+        cachedProfile &&
+        cachedProfile.profile === profile.kind &&
+        cachedProfile.fingerprint === profile.fingerprint
+      ) {
+        return cachedProfile.tools
```
zerob13, line 4202: **[P1] Refresh the tool manager before returning cached tools** When session A hits this cache branch after session B has loaded tools for a different project, `toolPresenter.getAllToolDefinitions()` is skipped, so the singleton `AgentToolManager`/`AgentBashHandler` still points at B's workspace. Subsequent `exec` in A defaults to B's cwd, and file tools in A add B's `agentWorkspacePath` to the allowed directories, bypassing the per-session workspace/permission boundary. Re-sync the tool manager for the session on cache hits or remove workspace state from the singleton before returning cached definitions.

src/main/presenter/agentRuntimePresenter/index.ts
```diff
@@ -4038,18 +4188,80 @@ export class AgentRuntimePresenter implements IAgentImplementation {
...
+      fingerprint: JSON.stringify({
+        kind,
+        projectDir: normalizedProjectDir ?? '',
+        providerId: state?.providerId ?? '',
+        modelId: state?.modelId ?? '',
+        toolRegistryRevision: this.toolRegistryRevision,
+        disabledAgentTools: [...disabledAgentTools].sort((left, right) =>
+          left.localeCompare(right)
+        ),
+        activeSkillNames
```
zerob13, line 4250: **[P2] Include skill enablement in the tool-profile key** The cache key only records `activeSkillNames`, but `getAllToolDefinitions()` also adds/removes `skill_list`, `skill_view`, `skill_run`, and settings tools based on `configPresenter.getSkillsEnabled()` and related skill state. If the user toggles Skills while no skills are active, `activeSkillNames` stays `[]`, this fingerprint still matches, and the session keeps stale skill tool schemas/prompts until some unrelated key changes. Include the Skills enabled state in this fingerprint or clear the cache on that setting change.

src/renderer/src/components/settings/ModelConfigItem.vue
```diff
@@ -20,6 +20,12 @@
         class="h-4 w-4 shrink-0 text-orange-500"
         title="函数调用能力"
       />
+      <Icon
+        v-if="type === ModelType.Chat"
```
zerob13: **[P2] Gate weak-agent warning on native tool capability** Every normal text model is rendered with `type ?? ModelType.Chat`, so this condition fires for nearly all chat models, including models that advertise `functionCall` or new API endpoint metadata. That shows the legacy-fallback warning on capable models in the settings list, giving users a false downgrade signal. Base this icon on the absence of native tool capability instead of the generic model type.

zerob13 (commented):
Codex review findings.

PR #1778, opened 2026-06-17
feat(tape): complete manifest grounding, integrity, and lineage

yyhhyyyyyy:
- Ground tool-loop provider requests with real `tool_call` / `tool_result` tape refs instead of synthetic-only manifest refs.
- Make view manifest hashes deterministic and bind request traces to the same request sequence.
- Add manifest integrity verification, compact lineage ranges, and narrowed reconstruction anchor refs.
- Surface manifest integrity and lineage details in TraceDialog.


zerob13 (commented):
Reviewed PR #1778. I found one issue around tool-loop manifest provenance dropping assistant text that is sent alongside tool calls. Verification run locally: targeted Vitest suites for tape manifest/facts/service and TraceDialog all pass; `pnpm run typecheck` passes; `git diff --check origin/dev...HEAD` passes.

src/main/presenter/agentRuntimePresenter/tapeViewManifest.ts
```diff
@@ -250,18 +277,72 @@ export function buildExcludedRefs(
...
+        lastToolCallIndex.set(toolCall.id, index)
+      }
+    } else if (message.role === 'tool' && message.tool_call_id) {
+      lastToolResultIndex.set(message.tool_call_id, index)
+    }
+  })
+
+  const refs: DeepChatTapeViewEntryRef[] = []
+  messages.forEach((message, index) => {
+    if (message.role === 'assistant' && message.tool_calls?.length) {
```
zerob13, line 298: [P2] Assistant messages can include both prompt-visible `content`/`reasoning_content` and `tool_calls` in the same request message. This branch now emits one ref per tool call and then returns, so any assistant text in that same message has no manifest ref at all. The `promptHash` still covers the text, but replay/lineage can look fully grounded while part of the actual prompt is only implicit. Please add a synthetic assistant-message ref when content or reasoning is present, in addition to the grounded tool-call refs, or otherwise represent that assistant text in the manifest refs.

PR #1912, opened 2026-07-10
refactor: ux performance optimization

zhangmo8:
## Summary
- Optimize chat submit/input state handling so attachment and skill-chip sync do not leave stale drafts or inconsistent composer state.
- Optimize ChatPage scrolling by simplifying custom scroll handling and keeping only the intended behaviors: CSS-driven smooth scrolling, scroll-to-bottom on session open, and setting-controlled auto-scroll while generating.
- Optimize assistant message activity rendering by making tool-call block keys stable and unique across multi-stage tool updates.
- Polish chat interaction performance and visual hierarchy with shared motion/blur/z-index tokens across chat chrome, spotlight, side panels, overlays, and message controls.
- Add UX interaction audit notes for the current optimization pass.

## UI layout
BEFORE:
```text
Chat route transition
└─ fades entire ChatPage message tree during session switches

Permission/question active
└─ ChatInputBox unmounted

Active scroll
└─ full blur + frequent row measurements
```

AFTER:
```text
Chat route shell
└─ stable page wrapper without full message-tree fade

Permission/question active
└─ ChatInputBox kept mounted via v-show/inert

Active user scroll
└─ simpler native scrolling + CSS smooth behavior
```

## Tests
- `pnpm run format`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck:web`
- `pnpm exec vitest --config vitest.config.renderer.ts test/renderer/components/ChatInputBox.test.ts test/renderer/components/ChatPage.test.ts test/renderer/composables/useMessageWindow.test.ts`
- `pnpm exec vitest --config vitest.config.renderer.ts test/renderer/components/message/messageActivityGroups.test.ts test/renderer/components/message/MessageItemAssistant.test.ts --run`
- `git diff --check`

## Review
- Ran targeted agent review loops; addressed reported scroll-measurement, programmatic-scroll, and duplicate tool-call key issues before updating this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale chat input updates during internal editor synchronization and file-chip syncing.
  * Reduced message-list layout jumps with more reliable height measurement and improved measurement batching.
  * Improved cached model-icon loading state handling to minimize flicker and shifts.
* **User Experience**
  * Refined scrolling/overscroll behavior, gesture handling, and overlay/panel layering for chat and workspace.
  * Improved transition and blur rendering consistency (including macOS-specific behavior).
* **Documentation**
  * Added a Chinese-language UX interaction audit document.
* **Tests**
  * Updated/added unit tests covering editor syncing, measurement thresholds, and icon/loading and key uniqueness.

### DIFF
--- a/src/renderer/src/pages/ChatPage.vue
+++ b/src/renderer/src/pages/ChatPage.vue
@@ -5,11 +5,6 @@
       data-testid="chat-page"
       :data-generating="String(isGenerating)"
       class="message-list-container h-full w-full min-w-0 overflow-y-auto"
-      :class="{ 'dc-list-scrolling': isListScrolling }"
-      @scroll.passive="onScroll"
-      @wheel.passive="onWheel"
-      @touchstart.passive="markListScrolling"
-      @pointerdown.passive="markListScrolling"
     >
       <ChatTopBar
         class="chat-capture-hide"
@@ -36,7 +31,7 @@
           />
         </div>
       </div>
-      <div ref="messageSearchRoot" class="min-h-[calc(100%-242px)]" :style="messageSearchRootStyle">
+      <div ref="messageSearchRoot" class="min-h-[calc(100%-242px)]">
         <div
           v-if="messageStore.isLoadingHistory"
           class="pointer-events-none px-6 py-2 text-center text-xs text-muted-foreground"
@@ -45,10 +40,8 @@
         </div>
         <MessageList
           ref="messageListRef"
-          :messages="visibleDisplayMessages"
+          :messages="displayMessages"
           :all-messages-for-capture="displayMessages"
-          :before-spacer-height="messageWindowBeforeHeight"
-          :after-spacer-height="messageWindowAfterHeight"
           :conversation-id="props.sessionId"
           :ephemeral-rate-limit-block="ephemeralRateLimitBlock"
           :ephemeral-rate-limit-message-id="ephemeralRateLimitMessageId"
@@ -62,9 +55,7 @@
           @continue="onMessageContinue"
           @trace="onMessageTrace"
           @edit-save="onMessageEditSave"
-          @measure="onMessageMeasure"
         />
-        <div ref="bottomScrollAnchor" class="h-px w-full" aria-hidden="true" />
       </div>
       <TraceDialog :message-id="traceMessageId" @close="traceMessageId = null" />
       <MemoryTurnDialog :read-only="isReadOnlySession" />
@@ -93,7 +84,6 @@
             <div>
               <div
                 v-if="latestPlanSnapshot || activePendingInteraction"
-                ref="planFloatLayer"
                 class="pointer-events-none absolute inset-x-0 bottom-[calc(100%+0.75rem)] flex w-full flex-col items-end gap-2"
                 style="z-index: var(--dc-z-float)"
                 data-testid="agent-progress-float-layer"
@@ -258,7 +248,6 @@ import { useSessionStore } from '@/stores/ui/session'
 import { useMessageStore } from '@/stores/ui/message'
 import { usePendingInputStore } from '@/stores/ui/pendingInput'
 import { useAgentPlanStore } from '@/stores/ui/agentPlan'
-import { useSpotlightStore } from '@/stores/ui/spotlight'
 import { useModelStore } from '@/stores/modelStore'
 import { createSessionClient } from '@api/SessionClient'
 import { isManualCompactionCommand } from '@/components/chat/mentions/utils'
@@ -273,7 +262,6 @@ import {
 import { WORKSPACE_EVENTS } from '@/events'
 import { filterUnsupportedAudioAttachments } from '@/lib/audioInputSupport'
 import { useSpeechRecognition } from '@/components/chat/composables/useSpeechRecognition'
-import { useMessageWindow } from '@/composables/message/useMessageWindow'
 import { playChatInputHeroFlight } from '@/lib/chatInputHero'
 import { scheduleStartupDeferredTask } from '@/lib/startupDeferred'
 import type {
@@ -295,7 +283,6 @@ const sessionStore = useSessionStore()
 const messageStore = useMessageStore()
 const pendingInputStore = usePendingInputStore()
 const agentPlanStore = useAgentPlanStore()
-const spotlightStore = useSpotlightStore()
 const modelStore = useModelStore()
 const chatClient = createChatClient()
 const modelClient = createModelClient()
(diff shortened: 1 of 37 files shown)


src/renderer/src/pages/ChatPage.vue
```diff
@@ -5,8 +5,6 @@
       data-testid="chat-page"
       :data-generating="String(isGenerating)"
       class="message-list-container h-full w-full min-w-0 overflow-y-auto"
```
zerob13, line 7: **P1 — Restore a trigger for loading older history.** Removing the `scroll` binding also removes the only path that called `messageStore.loadOlderMessages()` (`onScroll` → `loadOlderMessagesAtTop`, both deleted below). Session restore still requests only `INITIAL_MESSAGE_RESTORE_COUNT` (40), and there is no sentinel/observer or explicit load-more action elsewhere. As a result, conversations longer than 40 messages can never expose their older history.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -38,10 +40,8 @@
         </div>
         <MessageList
           ref="messageListRef"
-          :messages="visibleDisplayMessages"
+          :messages="displayMessages"
```
zerob13: **P2 — This removes the performance bound the PR is meant to improve.** `displayMessages` mounts the entire loaded history instead of the previous window. The existing `bounds mounted message rows for long loaded histories` test now receives 300 rows instead of at most 90. Every row also still installs a `ResizeObserver` in `MessageListRow`, even though `@measure` was removed here, so long conversations pay that observer/rAF cost for no consumer. Please either retain windowing or remove the unused measurement machinery and explicitly re-evaluate the long-history performance contract.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -324,12 +324,24 @@ async function loadMessagesForSession(sessionId: string, count?: number) {
...
+    loadMessagesForSession(id, INITIAL_MESSAGE_RESTORE_COUNT),
+    pendingInputStore.loadPendingInputs(id)
+  ])
+
+  if (requestId !== sessionRestoreRequestId) {
+    return
+  }
+
+  applyRestoredSessionSummary(restoredSession)
+  scheduleScrollToBottom()
```
zerob13: **P1 — Preserve Spotlight message jumps during restore.** The Spotlight store still sets `pendingMessageJump` before selecting a message result's session, but this PR removes `useSpotlightStore` and `focusPendingSpotlightMessageJump`. Restore now always schedules a bottom scroll here, so selecting a message result lands at the conversation bottom, never highlights the target, and never clears the pending jump. The existing `consumes pending spotlight message jumps...` test fails on this head.

zerob13 (commented):
Reviewed the current head (`9c4e6717c`). The simplified scroll implementation removes several still-required behaviors, and the targeted ChatPage suite does not pass on this head. Inline comments cover the reproducible regressions.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -383,539 +368,43 @@ const displayMessageCache = new Map<
...
-      sessionRestoreScrollFrame = null
-
-      if (!applySessionRestoreBottomScroll(requestId, sessionId)) {
-        cancelSessionRestoreScrollSettle()
-        return
-      }
-
-      remainingFrames -= 1
-      scheduleNextFrame()
+      el.scrollTo({ top: el.scrollHeight })
```
zerob13: **P1 — The documented targeted test command fails on this call.** `pnpm exec vitest --config vitest.config.renderer.ts test/renderer/components/ChatPage.test.ts --run` produces `TypeError: el.scrollTo is not a function` from this line (46 uncaught exceptions in the current jsdom environment). After locally supplying only a `scrollTo` polyfill, five behavioral assertions still fail. Using the previous `el.scrollTop = Math.max(el.scrollHeight - el.clientHeight, 0)` path would keep the production behavior deterministic and avoid requiring a new test-environment API shim.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -1398,25 +758,16 @@ const traceMessageIds = computed(() =>
...
-
-    if (scrollMode.value === 'initial-bottom') {
-      scrollToBottom(true)
-      scrollMode.value = 'auto-follow'
-    } else if (scrollMode.value === 'auto-follow') {
-      scrollToBottom(false)
-    } else {
-      scheduleScrollMetricsRead()
+    if (isGenerating.value && uiSettingsStore.autoScrollEnabled) {
+      scheduleScrollToBottom()
```
zerob13: **P1 — Do not override the user's reading position during generation.** Every stream revision now schedules a bottom scroll whenever the setting is enabled. The removed implementation tracked `isNearBottom`/`shouldAutoFollow` and stopped following after an upward wheel/scroll gesture; with that state gone, a user who scrolls up to read earlier content is pulled back to the bottom on the next token update. Keep a minimal near-bottom/follow-state guard so auto-scroll can be disengaged by user intent.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -1442,34 +793,7 @@ async function refreshChatSearchHighlights() {
...
-  }
-
-  scrollMode.value = 'manual-jump'
-  markProgrammaticScroll()
-  container.scrollTop = Math.max(entry.top - Math.round(container.clientHeight / 3), 0)
-  syncMessageViewportMetrics(container)
-  await nextTick()
-  applyChatSearchHighlights(root, chatSearchQuery.value)
-  setActiveChatSearchResult(root, result, { behavior })
+  setActiveChatSearchResult(root, chatSearchResults.value[nextIndex], { scroll: false })
```
zerob13: **P1 — Search navigation no longer reveals off-screen matches.** `setActiveChatSearchResult` calls `scrollIntoView` unless `options.scroll === false`; this argument explicitly disables it. Since `revealChatSearchResult` and its fallback were also removed, next/previous navigation only changes an off-screen highlight and gives the user no visible result. Removing `{ scroll: false }` restores the helper's built-in scrolling now that all rows are mounted.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -2498,61 +1802,38 @@ onMounted(() => {
...
 </script>
 
 <style>
 .message-list-container {
-  scrollbar-gutter: stable both-edges;
-  will-change: scroll-position;
-  overscroll-behavior: contain;
-  overflow-anchor: none;
-  scroll-behavior: auto;
+  scroll-behavior: smooth;
```
zerob13: **P2 — Avoid smooth behavior for high-frequency auto-follow.** This CSS also governs the `scrollTo` issued for every streaming revision. Each update starts/restarts a smooth animation while `scrollHeight` keeps growing, so the viewport can visibly lag or crawl behind the generated content. Keep the container at `scroll-behavior: auto` and request smooth behavior only for one-shot, user-initiated navigation.

PR #1959, opened 2026-07-13
refactor(agent): split runtime and harden loop

zhangmo8:
## Summary

- Split `AgentRuntimePresenter` from a multi-thousand-line monolith into a thin facade plus focused services (session lifecycle, pending input, generation control, stream lifecycle, interaction resume, memory/compaction, turn preparation, etc.), preserving the public IPC/presenter contract.
- Harden native agent loop reliability: cumulative multi-round usage accounting, consistent terminal `runOutcome`/`stopReason`, no-progress repeated-tool-loop guard, MCP tool call cancellation via `AbortSignal`, and subagent run timeouts/concurrency/handoff guardrails.
- Add offline native-agent behavior eval harness (`pnpm run test:agent:eval`) and wire it into PR CI.

## Context / docs

- Architecture: `docs/architecture/agent-runtime-presenter-split/`, `docs/architecture/native-agent-evaluation/`, `docs/architecture/subagent-run-guardrails/`
- Issues: multi-round usage, repeated tool loop, terminal outcome consistency, MCP tool cancellation under `docs/issues/`

## Test plan

- [ ] `pnpm run test:agent:eval`
- [ ] Focused unit suites for new services and process guards (`agentRuntimePresenter/*`, `mcpClient`, `toolManager`, `subagentOrchestratorTool`)
- [ ] `pnpm run lint` / architecture cleanup guard (`lint:agent-cleanup`) if available in local workflow
- [ ] Smoke: start agent turn, multi-tool rounds, stop mid-MCP tool, permission pause/resume, subagent run timeout and capacity limits
- [ ] Confirm message metadata retains cumulative usage, provider-round count, tool-call count, and correct terminal stop reasons

## Summary by CodeRabbit

* **New Features**
  * Added configurable subagent run deadlines and a limit of three active runs per parent.
  * Introduced a deterministic native-agent evaluation harness (with a new eval test command) for scenario-based reporting.
  * Improved cancellation/abort handling across agent runs, tool execution, MCP calls, media/compaction, and standalone generation.
  * Enhanced run status/stop-reason and token/usage accounting across pauses/resumes, plus clearer termination for empty/provider/context/tool-call limits.
  * Added detection/termination for repeated no-progress tool loops and improved follow-up pending state tracking.

* **Documentation**
  * Added architecture and issue documentation for evaluation, guardrails, usage accounting, cancellation, and terminal outcomes.

* **Tests**
  * Added deterministic native-agent eval scenarios and expanded cancellation, orchestration, and accounting coverage.


src/main/presenter/agentRuntimePresenter/dispatch.ts
```diff
@@ -1302,14 +1314,23 @@ async function runToolCall(params: {
...
+        onToolCallStarted?.(completedToolCall.id)
+      }
+      const result = await toolExecution.execute(toolCall, {
         onProgress: applyProgressUpdate,
         signal: io.abortSignal,
         permissionMode: toolPermissionMode,
         activeSkillNames: controls?.getActiveSkillNames?.(),
         enabledSkillNames: controls?.getEnabledSkillNames?.()
       })
+      io.abortSignal.throwIfAborted()
```
zerob13: [P1] Persist a returned tool result before honoring late cancellation

A tool can finish its side effect and resolve while also aborting the parent signal. This post-await check discards the returned result, so `executeTools` never applies it and `afterRoundPersisted` never writes its `tool_call` / `tool_result` Tape facts. The new `persists the completed tool batch before a post-tool abort` test reproduces this and currently fails. Stage and commit a successfully returned result first, then let the outer batch settle the run as aborted.

src/main/presenter/agentRuntimePresenter/index.ts
```diff
@@ -2266,11 +2405,30 @@ export class AgentRuntimePresenter {
...
             }
           })
-          const execution = await this.executeDeferredToolCall(sessionId, messageId, toolCall)
+          let deferredToolCallCounted = false
+          const markDeferredToolCallStarted = () => {
+            if (deferredToolCallCounted) {
+              return
+            }
+            deferredToolCallCounted = true
+            resumeAccounting = incrementToolCallAccounting(resumeAccounting)
```
zerob13: [P1] Enforce the global cap before running a deferred call

At a post-call permission pause, the probe is already persisted as the 128th call (`process.test.ts` explicitly covers 127 → 128). Granting that interaction reaches this increment and `executeDeferredToolCall` without any cap check, so the side effect executes and metadata becomes 129. The loop guard only rejects a later provider-emitted batch and cannot protect this path. Reject or reserve the deferred execution before invoking it, and never persist a count above 128.

src/main/presenter/agentRuntimePresenter/dispatch.ts
```diff
@@ -145,6 +145,14 @@ type MutableToolBatchState = {
 const PARALLEL_READ_ONLY_AGENT_TOOLS = new Set(['read'])
 const USER_CANCELED_GENERATION_ERROR = 'common.error.userCanceledGeneration'
 
+function isToolCancellation(error: unknown, signal: AbortSignal): boolean {
+  const errorName =
+    error && typeof error === 'object' && 'name' in error
+      ? (error as { name?: unknown }).name
+      : undefined
+  return signal.aborted || errorName === 'AbortError' || errorName === 'CanceledError'
```
zerob13: [P1] Do not infer parent cancellation from an error name

`AbortError` is also used for model request timeouts (`BaseLLMProvider.createModelRequestTimeoutError`), and agent tools can invoke those provider paths. With a live `io.abortSignal`, this helper still classifies that tool-local timeout as parent cancellation; the error escapes the tool batch and `processStream` persists `runOutcome: aborted` / `runStopReason: user_stop`. Treat it as parent cancellation only when the run signal is aborted (or the thrown value is its exact reason); otherwise preserve it as a tool/provider error.

test/main/presenter/agentRuntimePresenter/acpCompatibilityAdapters.test.ts
```diff
@@ -231,8 +231,12 @@ describe('ACP compatibility adapters', () => {
     const blocks = JSON.parse(message?.content ?? '[]')
     const tapeRecord = harness.getAssistantTapeRecord()
 
-    expect(settlement).toEqual({ status: 'completed', stopReason: 'complete' })
-    expect(message?.status).toBe('sent')
+    expect(settlement).toEqual({
```
zerob13, line 234: [P1] Restore the ACP success/failure expectations

The settlement/status expectations for this `complete()` case and the `adapter.fail()` case below were swapped. `complete()` returns `completed`/`sent`, while `fail()` returns `error`/`error`; the surrounding event and Tape assertions already reflect those outcomes. The full main suite currently fails both tests. Restore the original success assertions here and the error assertions in the failure case.

zerob13 (changes requested):
Requesting changes for runtime correctness and a red repository test suite.

Validation at `9d90fb7b`:
- `pnpm run typecheck:node`: passed
- `pnpm run test:agent:eval`: passed
- `pnpm exec vitest --run --config vitest.config.ts test/main`: failed (10 tests in 3 files; 3,837 passed)

The green PR workflow currently runs the native-agent eval but not `test/main`, so it does not expose these failures. In addition to the inline lifecycle failure and swapped ACP assertions, six interaction tests fail because the new `deepchatMessagesTable.updateMetadata` call was not added to the presenter test fixture, and the deferred-subagent cancellation test still expects the old resolved-error contract.

PR #1963, opened 2026-07-14
fix(agent): multi-agent host isolation and session lifecycle

zhangmo8:
## Summary
- Enforce per-host-agent isolation for DeepChat runtime: MCP allow-lists, cross-session workspace allow dirs, transfer permission/skill reset, and cross-agent subagent policy resolution.
- Improve multi-agent UX and lifecycle safety: ACP tool progress no longer looks like permissions, Plugins Hub clarifies global vs agent-scoped pages, parent permission approvals inherit into subagents, and chat/session timeouts plus best-effort deletion avoid zombie sessions.
- i18n for Plugins Hub scope banners is complete across locales with native strings (zh/en/de/fr/ja/…); `pnpm run i18n` clean.

## Key changes
- **MCP allow-list runtime**: pass `enabledMcpServerIds` into catalog/fingerprint/call/deferred paths
- **Workspace isolation**: stop merging last-synced workdir across concurrent sessions
- **Transfer rebind**: clear permissions/plan/runtime skills; refilter active skills to target policy
- **Subagent target policy**: cross-agent children use target permissionMode/tools/systemPrompt/skill allow-list; keep parent workdir/model
- **Subagent approvals**: clone parent session permission approvals; clearer waiting progress with child sessionId
- **ACP projection**: tool progress emits reasoning only (no pseudo `tool_call_permission` blocks)
- **Plugins Hub**: scope banner for global plugins vs current-agent Skills/MCP policy (localized for en/de/fr/ja/ko/…)
- **Chat/Session**: remove fake stream lock; longer create/list timeouts; deletion continues row cleanup after partial failures

## Test plan
- [x] `pnpm run i18n` — no missing/invalid keys
- [x] `pnpm run format` / `pnpm run lint` (incl. architecture guards)
- [x] Focused vitest for agent runtime, assignment policy, chatService, acpContentMapper, deletion, subagent
- [ ] Manual: Agent A/B MCP allow-list catalog differs
- [ ] Manual: transfer A→B resets approvals and skills
- [ ] Manual: cross-agent subagent uses target policy
- [ ] Manual: Plugins Hub banners in zh-CN / en-US / de-DE / fr-FR / ja-JP
- [ ] Manual: ACP tool progress has no permission-style cards

## Summary by CodeRabbit

* **New Features**
  * Added clearer plugin scope messaging across supported languages.
  * Improved subagent handling with agent-specific policies, permissions, skills, and workspace behavior.
  * Added contextual progress guidance when subagent actions require attention.

* **Bug Fixes**
  * Strengthened isolation between agents and concurrent sessions for MCP tools, file access, permissions, and skills.
  * Prevented unavailable tools from executing and corrected permission-retry handling.
  * ACP tool progress now appears as reasoning instead of a misleading permission prompt.
  * Session deletion and stream cancellation now continue cleanup after partial failures.
* **Documentation**
  * Added architecture specifications and implementation plans for multi-agent and subagent isolation.


src/main/presenter/toolPresenter/agentTools/agentToolManager.ts
```diff
@@ -1260,8 +1260,9 @@ export class AgentToolManager {
       ordered.push(resolved)
     }
 
+    // Only trust the call-scoped workspace path. Do not merge the manager's last
```
zerob13, line 1262: [P1] `workspacePath` is not actually call-scoped here. Both execution and pre-check compute it as `dynamicWorkdir ?? this.agentWorkspacePath ?? ...` (currently lines 962–963 and 2023–2024). When a session has no `projectDir`, or its lookup fails, `this.agentWorkspacePath` is the last session passed to `syncContext`, so session A still whitelists session B's workspace. Use the per-session default when a `conversationId` is present and never fall back to manager state; add a two-session regression test.

src/main/presenter/index.ts
```diff
@@ -565,6 +565,11 @@ export class Presenter implements IPresenter {
         this.filePermissionService.clearConversation(sessionId)
         this.settingsPermissionService.clearConversation(sessionId)
       },
+      cloneSessionPermissions: (sourceSessionId, targetSessionId) => {
```
zerob13, line 564: [P1] The aggregate permission port still leaves MCP session grants behind. `McpPresenter.grantPermission(..., remember=false, conversationId)` stores them in `ToolManager.sessionPermissions`, but neither `clearSessionPermissions` above nor this clone touches that cache. A transfer therefore preserves an MCP approval under the same session ID, so a stricter target agent can call an allowed shared server without re-prompting. Include MCP clearing in this port (and define explicit clone semantics) and cover the transfer path.

src/main/presenter/sessionApplication/lifecycleCoordinator.ts
```diff
@@ -249,12 +253,14 @@ export class SessionLifecycleCoordinator implements SessionLifecyclePort {
...
           projectDir,
-          permissionMode: input.permissionMode,
+          permissionMode: runtimeConfig.permissionMode,
           generationSettings: runtimeConfig.generationSettings
         })
         if (runtimeConfig.activeSkills.length > 0) {
           await this.dependencies.skills.setActiveSkills(sessionId, runtimeConfig.activeSkills)
         }
+        // Inherit parent session approvals so child does not re-prompt for already-trusted work.
+        this.dependencies.permissions?.cloneSessionPermissions?.(parentSessionId, sessionId)
```
zerob13: [P1] This clones parent approvals into every child after resolving the target policy. For a cross-agent child, that defeats the target `permissionMode`: for example, external file paths approved in the parent are copied into the child's allowed directories and no target-agent prompt occurs. It also contradicts the isolation contract's conversation-scoped permission cache. At minimum, only inherit approvals for self-target children; cross-agent children must start with a clean cache.

src/main/presenter/sessionApplication/agentAssignmentPolicy.ts
```diff
@@ -117,14 +118,46 @@ export class SessionAgentAssignmentPolicy implements SessionAssignmentPolicyPort
...
-      targetAgentId: input.targetAgentId?.trim() ? descriptor.id : null,
+      targetAgentId,
       providerId: input.providerId,
       modelId: input.modelId,
-      generationSettings: input.generationSettings,
-      disabledAgentTools: normalizeDisabledAgentTools(input.disabledAgentTools),
-      activeSkills: normalizeActiveSkills(input.activeSkills)
+      permissionMode: normalizePermissionMode(agentConfig?.permissionMode),
+      generationSettings,
+      disabledAgentTools: normalizeDisabledAgentTools(agentConfig?.disabledAgentTools),
```
zerob13, line 156: [P1] This only filters the target tool catalog; it is not a runtime boundary. `dispatch.buildToolExecutionContext` still invokes a call when `toolDef` is undefined, and `ToolPresenter.getToolSource` then falls back to the process-global mapper. With concurrent sessions, a permissive session can map `exec`, after which a target child with `exec` disabled can still execute a hallucinated or stale `exec` call. Reject calls absent from the current session definitions (as the deferred path already does), or enforce `disabledAgentTools` again at execution.

src/main/routes/sessions/sessionService.ts
```diff
@@ -52,7 +55,7 @@ export class SessionService {
   ): Promise<SessionWithState> {
     return await this.deps.scheduler.timeout({
       task: this.deps.lifecycle.createSession(input, context.webContentsId),
-      ms: SESSION_OPERATION_TIMEOUT_MS,
+      ms: SESSION_CREATE_TIMEOUT_MS,
```
zerob13: [P1] This changes the create/list timeout contract without updating `test/main/routes/sessionService.test.ts`. `pnpm exec vitest run test/main/routes/sessionService.test.ts` currently fails: the test expects 5,000 ms but receives 60,000/15,000 ms. Update the assertions and test name so `test:main` is green.

src/main/routes/sessions/sessionService.ts
```diff
@@ -52,7 +55,7 @@ export class SessionService {
   ): Promise<SessionWithState> {
     return await this.deps.scheduler.timeout({
       task: this.deps.lifecycle.createSession(input, context.webContentsId),
```
zerob13: [P2] `Scheduler.timeout` only races promises; it does not cancel `lifecycle.createSession`. If ACP initialization crosses 60 seconds, the caller receives `TimeoutError` while the underlying task can later bind the window, publish the session, and start the initial turn. A retry can therefore create duplicate or unexpected sessions. Either avoid a route timeout around this mutating operation or thread an `AbortSignal` through creation and roll back on timeout.

src/main/routes/chat/chatService.ts
```diff
@@ -226,4 +203,25 @@ export class ChatService {
       ...result
     }
   }
+
+  private async bestEffortCancel(sessionId: string, reason: string): Promise<void> {
+    const cleanupResults = await Promise.allSettled([
+      Promise.resolve(this.deps.sessionPermissionPort.clearSessionPermissions(sessionId)),
```
zerob13: [P2] `clearSessionPermissions` returns `void` and may throw synchronously. Its result is evaluated before `Promise.resolve`, so a throw prevents `Promise.allSettled` from being created, skips `cancelGeneration`, and replaces the original send `TimeoutError`. Wrap this call in `Promise.resolve().then(() => ...)`, as `stopStream` already does.

zerob13 (changes requested):
Requesting changes for runtime isolation gaps and a confirmed test failure. I verified the branch at ef44dc2; typecheck, lint, format check, i18n, and the PluginsHub renderer test pass, while the main test suite fails in sessionService.test.ts.

src/main/presenter/toolPresenter/agentTools/agentToolManager.ts
zhangmo8: Fixed in 563015c. Conversation-scoped execution and pre-check now use the isolated default workspace when workdir resolution is empty or fails, never the manager fallback. Added a two-session regression in agentToolManagerRead.test.ts.

src/main/presenter/index.ts
zhangmo8: Fixed in 563015c. The aggregate permission port now clears ToolManager MCP session grants. Clone semantics explicitly clear target MCP grants and copy only command/file/settings approvals; transfer and MCP clear seams are covered by tests.

src/main/presenter/sessionApplication/lifecycleCoordinator.ts
zhangmo8: Fixed in 563015c. Lifecycle now clones approvals only when the resolved child agent matches parentAgentId. Cross-agent and ACP children start clean; self-target and cross-agent behavior is covered by lifecycle tests.

src/main/presenter/sessionApplication/agentAssignmentPolicy.ts
zhangmo8: Fixed in 563015c. Dispatch now rejects and stages an error for any call missing from the current session tool definitions before ToolPresenter invocation. Added a regression proving a filtered exec call is not executed.

src/main/routes/sessions/sessionService.ts
zhangmo8: Fixed in 563015c. Updated the test name and timeout assertions: list uses 15 seconds, ordinary reads use 5 seconds, and create is no longer raced through Scheduler.timeout. sessionService tests pass.

src/main/routes/sessions/sessionService.ts
zhangmo8: Fixed in 563015c by removing the non-cancelling route timeout around lifecycle.createSession. The mutating operation now completes through the lifecycle owner, avoiding late publication after a route timeout.

src/main/routes/chat/chatService.ts
zhangmo8: Fixed in 563015c. Both cleanup calls are now deferred through Promise.resolve().then, so synchronous permission-clear failures cannot skip cancelGeneration or replace the original send TimeoutError. Added regression coverage.

zhangmo8: Fixed in 563015c. Both cleanup calls are now deferred through Promise.resolve().then, so synchronous permission-clear failures cannot skip cancelGeneration or replace the original send TimeoutError. Added regression coverage.

PR #1965, opened 2026-07-14
feat(grok): add oauth device login

zhangmo8:
## Summary

- add xAI device-code OAuth for SuperGrok and X Premium+ accounts
- store credentials outside provider records and coordinate token refreshes
- retain API-key fallback while restricting OAuth bearer use to trusted x.ai endpoints
- add localized UI copy for all supported locales and main/renderer regression tests

## UI

BEFORE

    Grok provider settings
    ├── API URL
    ├── API Key
    └── Verify / Refresh models

AFTER

    Grok provider settings
    ├── API URL
    ├── xAI Grok OAuth
    │   ├── connection status / account
    │   ├── device code + verification link
    │   └── sign in / reconnect / sign out
    ├── API Key fallback
    └── Verify / Refresh models

## Validation

- pnpm run format
- pnpm run i18n
- pnpm run lint
- pnpm run typecheck
- pnpm exec vitest run test/main/presenter/xaiGrokAuth.test.ts test/renderer/components/GrokOAuth.test.ts test/renderer/components/ProviderApiConfig.test.ts

## Summary by CodeRabbit

* **New Features**
  * Added first-class xAI Grok OAuth sign-in using device authorization.
  * Added Grok connection status, verification, cancellation, reconnection, and sign-out controls in provider settings.
  * OAuth credentials are securely stored and refreshed automatically.
  * Existing Grok API-key authentication remains available as a fallback.
  * Added localized settings text across supported languages.
* **Security**
  * Restricted OAuth credential injection to trusted xAI HTTPS endpoints.
* **Tests**
  * Added coverage for authentication, token refresh, secure storage, endpoint validation, and settings UI behavior.

### DIFF
--- a/src/main/presenter/llmProviderPresenter/providers/aiSdkProvider.ts
+++ b/src/main/presenter/llmProviderPresenter/providers/aiSdkProvider.ts
@@ -48,6 +48,9 @@ import {
 } from '../aiSdk'
 import type { AiSdkProviderKind } from '../aiSdk/providerFactory'
 import { normalizeAzureBaseUrl, normalizeGeminiBaseUrl } from '../aiSdk/providerFactory'
+import { shouldUseXaiGrokOAuthFetch } from '../xaiGrokAuthAdapter'
+import { getGlobalXaiGrokAuth } from '../../xaiGrokAuth'
+import { isTrustedXaiApiEndpoint } from '../../xaiGrokAuth/constants'
 import { proxyConfig } from '../../proxyConfig'
 import type { ProviderMcpRuntimePort } from '../runtimePorts'
 import {
@@ -585,10 +588,22 @@ export class AiSdkProvider extends BaseLLMProvider {
   }
 
   private getRuntimeProvider(decision: RouteDecision): LLM_PROVIDER {
-    return {
+    const base: LLM_PROVIDER = {
       ...this.provider,
       ...decision.providerPatch
     }
+
+    if (shouldUseXaiGrokOAuthFetch(base)) {
+      const oauthToken = getGlobalXaiGrokAuth().peekAccessToken()
+      if (oauthToken) {
+        return {
+          ...base,
+          oauthToken
+        }
+      }
+    }
+
+    return base
   }
 
   private getResolvedModelConfig(modelId: string, modelConfig?: ModelConfig): ModelConfig {
@@ -834,7 +849,6 @@ export class AiSdkProvider extends BaseLLMProvider {
     decision?: RouteDecision
   ): Promise<T> {
     const resolvedDecision = decision ?? { providerKind: this.definition.runtimeKind }
-    const runtimeProvider = this.getRuntimeProvider(resolvedDecision)
     const defaultHeaders = {
       ...this.defaultHeaders,
       ...this.definition.defaultHeadersPatch
@@ -862,13 +876,26 @@ export class AiSdkProvider extends BaseLLMProvider {
     }
 
     try {
+      // Refresh Grok OAuth tokens before model/list requests that use sync headers.
+      const runtimeProvider = this.getRuntimeProvider(resolvedDecision)
+      const useGrokOAuth =
+        shouldUseXaiGrokOAuthFetch(runtimeProvider) && isTrustedXaiApiEndpoint(url)
+      if (useGrokOAuth) {
+        await getGlobalXaiGrokAuth()
+          .ensureAccessToken()
+          .catch(() => null)
+      }
+      const refreshedRuntimeProvider = this.getRuntimeProvider(resolvedDecision)
+      const requestRuntimeProvider = useGrokOAuth
+        ? refreshedRuntimeProvider
+        : { ...refreshedRuntimeProvider, oauthToken: undefined }
       const dispatcher = this.getFetchDispatcher()
       const response = await fetch(url, {
         ...init,
         headers: {
           ...this.getRequestHeaders(
             resolvedDecision,
-            runtimeProvider,
+            requestRuntimeProvider,
             defaultHeaders,
             init.body && !(init.body instanceof FormData) ? 'application/json' : undefined
           ),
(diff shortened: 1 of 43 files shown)


src/main/presenter/xaiGrokAuth/index.ts
```diff
@@ -0,0 +1,744 @@
...
+      this.pendingDeviceFlow = null
+    }
+    this.publishStatusChanged()
+    return this.getStatus()
+  }
+
+  async logout(): Promise<XaiGrokAuthStatus> {
+    const tokens = this.store.load()
+    this.cancelLogin()
+    this.store.clear()
```
zerob13, line 258: [P1] Prevent an in-flight refresh from restoring credentials after logout

`logout()` clears the store but neither invalidates nor waits for `refreshPromise`. If `ensureAccessToken()` is already refreshing, `refreshAccessTokenOnce()` can complete after this line and call `store.save(refreshed)` at line 594, recreating the credential file. The final `getStatus()` may then return `authenticated`; with refresh-token rotation, revoking the old token may not revoke the newly persisted token either. Serialize logout with refresh or add an auth-generation guard that prevents stale refreshes from saving, and cover this interleaving in a test.

src/main/presenter/llmProviderPresenter/xaiGrokAuthAdapter.ts
```diff
@@ -0,0 +1,110 @@
...
+    try {
+      const refreshed = await auth.forceRefreshAccessToken()
+      if (!refreshed || refreshed === accessToken) {
+        return response
+      }
+      return underlyingFetch(url, {
+        ...nextInit,
+        headers: applyBearerHeaders(init?.headers, defaultHeaders, refreshed)
+      })
+    } catch {
```
zerob13, line 99: [P1] Fall back to the configured API key when OAuth recovery fails

If a stored access token is revoked while still inside its local expiry window and its refresh token is also invalid, this catch returns the OAuth 401. Every later request selects the same stored OAuth token again, so a valid `provider.apiKey` is never tried until token expiry or manual sign-out. That contradicts the documented API-key fallback when OAuth is unavailable. When an API key exists, dispose of the rejected response and retry once with that key; add a revoked-OAuth + valid-API-key regression test.

src/main/presenter/llmProviderPresenter/providers/aiSdkProvider.ts
```diff
@@ -862,13 +876,26 @@ export class AiSdkProvider extends BaseLLMProvider {
...
+      const runtimeProvider = this.getRuntimeProvider(resolvedDecision)
+      const useGrokOAuth =
+        shouldUseXaiGrokOAuthFetch(runtimeProvider) && isTrustedXaiApiEndpoint(url)
+      if (useGrokOAuth) {
+        await getGlobalXaiGrokAuth()
+          .ensureAccessToken()
+          .catch(() => null)
+      }
+      const refreshedRuntimeProvider = this.getRuntimeProvider(resolvedDecision)
+      const requestRuntimeProvider = useGrokOAuth
```
zerob13, line 889: [P2] Apply OAuth failure fallback to model/list requests too

This path always keeps `refreshedRuntimeProvider` whenever the provider and URL are OAuth-eligible, then converts any non-OK response directly into `ProviderHttpError`. A server-revoked but locally unexpired OAuth token therefore blocks model refreshes and connection checks even when `provider.apiKey` is valid; this path never reaches `createXaiGrokFetch`'s recovery. Retry an auth rejection with a refreshed token and then the configured API key, or centralize the auth fallback so both request paths share it.

src/main/presenter/llmProviderPresenter/providers/aiSdkProvider.ts
```diff
@@ -2297,6 +2324,12 @@ export class AiSdkProvider extends BaseLLMProvider {
   private validateCredentials(strategy: AiSdkCredentialStrategy): string | null {
     switch (strategy) {
       case 'api-key':
+        if (shouldUseXaiGrokOAuthFetch(this.provider)) {
+          if (this.provider.apiKey || getGlobalXaiGrokAuth().isAuthenticated()) {
```
zerob13, line 2328: [P2] Count the supported environment access token as a credential

`peekAccessToken()` and `ensureAccessToken()` honor `XAI_GROK_ACCESS_TOKEN` / `XAI_ACCESS_TOKEN`, but `isAuthenticated()` only reflects credentials loaded from the store. With only an environment token configured, `check()` exits here with “Missing API key or xAI Grok OAuth sign-in” before `runText()`, even though runtime requests can authenticate. Use a credential predicate that includes the environment token (and keep status semantics explicit), then cover the env-only check path.

src/main/presenter/xaiGrokAuth/index.ts
```diff
@@ -0,0 +1,744 @@
...
+
+  private parseTokenResponse(
+    payload: TokenResponse,
+    previous?: XaiGrokTokenSet,
+    endpoints?: {
+      tokenEndpoint?: string
+      deviceAuthorizationEndpoint?: string
+    }
+  ): XaiGrokTokenSet {
+    const accessToken = toStringValue(payload.access_token) || previous?.accessToken
```
zerob13, line 608: [P2] Do not extend the previous access token on a malformed refresh response

For a refresh, a 2xx payload without `access_token` falls back to `previous.accessToken`, and line 628 then assigns it a new expiry from `expires_in` (or the one-hour default). This can silently make a stale token look valid again and suppress both the error state and API-key fallback. A successful OAuth token response must provide an access token; only fields such as `refresh_token` or `id_token` should inherit from the previous set. Reject this response and add a regression test.

src/main/presenter/xaiGrokAuth/index.ts
```diff
@@ -0,0 +1,744 @@
...
+      )
+      if (flow.cancelled || this.pendingDeviceFlow !== flow) {
+        return
+      }
+
+      const tokenSet = this.parseTokenResponse(tokens, undefined, {
+        tokenEndpoint: discovery.tokenEndpoint,
+        deviceAuthorizationEndpoint: discovery.deviceAuthorizationEndpoint
+      })
+      this.store.save(tokenSet)
```
zerob13, line 370: [P2] Reject a device login that cannot be refreshed

The initial device-token response is persisted even when `refresh_token` is absent. The UI then reports a successful authenticated session, but the first access-token expiry always fails with “refresh token is unavailable,” violating this feature's automatic-refresh acceptance criterion. Since the flow requests `offline_access`, require a non-empty refresh token before saving the initial device login (while still allowing refresh responses to reuse the previous refresh token), and test the missing-token response.

zerob13 (changes requested):
Found six correctness issues not covered by the existing review: a logout/refresh race, incomplete API-key fallback in both request paths, environment-token validation drift, and two token-response validation gaps. Please add focused regression tests for these paths.

PR #1966, opened 2026-07-14
fix(memory): harden embedding reindex recovery

yyhhyyyyyy:
- enforce provider embedding batch limits with bounded fallback
- propagate abort signals through embedding and dimension requests
- eliminate reindex drain and warmup result races
- redact sensitive errors and localize rebuild failure feedback

## Summary by CodeRabbit

* **New Features**
  * Added enhanced memory reindexing status tracking (completed/blocked/aborted) with user-visible failure details.
  * Added a Memory Diagnostics banner for failed reindex runs, including one-click retry when the failure is retryable.
  * Added localized messaging for incomplete vector rebuilding and internal reindex failure reasons.
* **Bug Fixes**
  * Improved embedding batching for large embedding requests by adapting to provider limits.
  * Added request cancellation support for embedding and dimension operations across the embedding pipeline.
  * Preserve memory content and fall back to keyword search when vector rebuilding is incomplete.
  * Clear stale reindex status when memories are removed.


src/renderer/settings/components/MemoryDiagnosticsPanel.vue
```diff
@@ -25,6 +25,34 @@
...
+        <Icon icon="lucide:triangle-alert" class="mt-0.5 h-4 w-4 shrink-0 text-destructive" />
+        <p class="text-xs leading-5 text-foreground">
+          {{
+            t('settings.memory.redesign.reindexIncomplete', {
+              reason: reindexFailureReason
+            })
+          }}
+        </p>
+      </div>
+      <Button
```
zhangmo8, line 44: [P2] Only offer this action for retryable failures. `lastError.retryable` is part of the new status contract but is never consumed in the renderer, so deterministic failures still render a retry button. In particular, `pending-restart` is explicitly classified as non-retryable; clicking this button just starts the same rebuild against quarantined files and returns the same blocked result. Please gate the retry action on `props.status?.lastReindex?.lastError?.retryable === true` (and show restart-only guidance for `pending-restart`), and add coverage for the non-retryable case.

PR #1973, opened 2026-07-14
refactor(agent): tighten message, permission, and stop boundaries

zerob13:
## Summary

- canonicalize send-message input once at the session application boundary
- assign explicit owners for permission-mode defaults and persisted decoding
- preserve provider stop reasons through AI SDK, ACP, accumulator, and settlement paths
- validate pending input payloads at the storage boundary and degrade corrupt payloads to raw user text
- retain completed architecture contracts while removing execution-only SDD artifacts and redundant tests

## Why

Repeated normalization and lossy stop-reason mapping made downstream modules distrust typed callers and allowed invalid or bounded outcomes to be hidden. This change narrows internal contracts and keeps compatibility handling at owned boundaries. Pending payload corruption is logged but fails open so conversation continuity is not blocked.

## Validation

- pnpm run format
- pnpm run i18n
- pnpm run lint
- pnpm run typecheck
- pnpm run test:memory (46 files passed, 745 tests passed)
- pnpm run test:main -- --run (365 files passed, 4189 tests passed)

## Summary by CodeRabbit

* **New Features**
  * Added structured message input handling for sending, steering, and queueing (text/files/skills/inline items), with schema-based validation and graceful decoding of legacy or malformed saved queued payloads.
* **Bug Fixes**
  * Improved provider/ACP terminal outcome handling: missing terminal events now error; token/turn-limit outcomes (including `max_turn_requests`) are preserved; tool-use cancellations/refusals produce correct error+stop sequences without “successful completion” fallbacks.
  * Permission modes now retain the explicitly provided mode and persist as a required value, with safer handling of unknown legacy values.
* **Documentation**
  * Documented message/application, permission-policy ownership, and provider stop-reason mapping contracts; removed an outdated terminal-outcome inconsistency spec.

PR #1974, opened 2026-07-15
fix(chat): stabilize scroll ownership

zhangmo8:
## Summary

- centralize every ChatPage viewport write behind one typed scroll controller, request arbiter, and one-write-per-frame queue
- preserve durable user ownership across restore, streaming follow, history prepend, measurement anchoring, Cmd/Ctrl+F, and Spotlight navigation
- commit target-session message views atomically and reuse a bounded recent-session cache for faster initial paint and switching
- isolate header, viewport, and composer geometry; keep loading indicators out of message flow and use a chat-shaped skeleton for uncached switches
- require a real loaded-history window, upward intent, top proximity, and scroll idle before pagination

### Before

```text
ChatPage shared flow
+-- top bar
+-- message viewport
|   +-- in-flow loading state
|   +-- message rows
|   +-- restore / follow / history / search writers
+-- plan and pending UI
+-- composer
```

### After

```text
ChatPage grid
+-- top bar
+-- viewport region
|   +-- message scroll owner
|   |   +-- zero-height history status
|   |   +-- bounded message window
|   +-- uncached-session skeleton overlay
+-- composer region

all scroll intents -> controller -> one active operation -> one write per frame
```

## Behavior and Compatibility

- user wheel, touch, scrollbar, or keyboard intent cancels active and queued programmatic work
- cached session switches render immediately; uncached switches hide stale content until one atomic commit
- short conversations do not trigger history loading or measurement-driven scroll writes
- search and Spotlight retain explicit navigation priority without triggering pagination
- recent session views are capped at 5 entries and approximately 24 MB
- existing messages and persisted session data are not migrated or rewritten

## Validation

- focused renderer regression: 9 files, 141 tests passed
- stabilized full renderer run: 167 files, 1252 tests passed
- production build passed
- pnpm run typecheck:web
- pnpm run format
- pnpm run i18n
- pnpm run lint
- git diff --check

## Remaining Manual Validation

- run the provider-backed Electron scroll ownership smoke scenario
- verify physical macOS trackpad and mouse-wheel behavior in short and multi-page conversations
- record cold-load, cached-switch, uncached-switch, and anchor-drift metrics on the reference machine

## Summary by CodeRabbit

* **New Features**
  * Improved chat scrolling with exclusive scrollbar ownership across session restore, streaming, search, history loading, and message navigation.
  * Added bounded message rendering to keep performance stable in long chats.
  * Preserved recent session views and message measurements across session switches and remounts.
  * Added a chat session loading skeleton while content prepares.
* **Bug Fixes**
  * Reduced unwanted scroll jumps, incorrect history pagination triggers, and stale UI updates during session switching and message resizing.
* **Tests**
  * Added Playwright smoke coverage and expanded renderer unit tests for scroll ownership and caching behavior.


src/renderer/src/pages/ChatPage.vue
```diff
@@ -737,54 +623,96 @@ function isAtBottom(): boolean {
...
+
+  const viewport = scrollContainer.value
+  const messageRoot = messageSearchRoot.value
+  if (!viewport || !messageRoot) return
+
+  viewportResizeObserver = new ResizeObserver(() => {
+    const currentViewport = scrollContainer.value
+    if (!currentViewport) return
+    syncMessageViewportMetrics(currentViewport)
+    chatScrollController.notifyViewportResize()
```
zhangmo8, line 675: The persistent geometry observer bypasses the auto-scroll setting. Both the message root and viewport feed this callback, and notifyViewportResize() unconditionally enqueues an auto-follow request whenever the controller is in following mode. After session restore, that mode is active even when uiSettingsStore.autoScrollEnabled is false, so streamed row growth or a composer/inline-chip height change that resizes the viewport can still write scrollTop to the bottom. Please gate only the auto-follow branch on the setting while retaining session-restore behavior, and add an autoScrollEnabled=false resize regression.

src/renderer/src/composables/chat/chatScrollState.ts
```diff
@@ -0,0 +1,175 @@
...
+        mode: 'following',
+        userOwned: false,
+        nearBottom: true,
+        resumeUserOwnedAfterNavigation: false
+      }
+    case 'explicit-navigation-start':
+      return {
+        ...state,
+        mode: 'navigating',
+        resumeUserOwnedAfterNavigation: state.userOwned,
```
zhangmo8: Repeated explicit navigation can erase durable reading ownership. acceptRequest() runs this transition for every accepted request, including a same-owner request coalesced while the prior search/Spotlight operation is still active. The first start from reading stores true and clears userOwned; the second start then stores the now-false userOwned value, so completion enters following mode and later stream/resize work can pull away from the searched result. Preserve resumeUserOwnedAfterNavigation while already navigating (or OR it with state.userOwned) and cover two starts before one completion.

src/renderer/src/pages/ChatPage.vue
```diff
@@ -1158,37 +993,69 @@ async function focusPendingSpotlightMessageJump(attempt = 0): Promise<void> {
...
-    uiSettingsStore.autoScrollEnabled && isNearBottom.value ? 'auto-follow' : 'anchored-reading'
 }
 
 // Load messages when sessionId changes, then scroll to bottom
 watch(
   () => props.sessionId,
   async (id) => {
+    if (measurementSessionId && measurementSessionId !== id) {
+      recentMessageMeasurements.delete(measurementSessionId)
+      recentMessageMeasurements.set(measurementSessionId, captureMessageWindowMeasurements())
```
zhangmo8: This measurement cache cannot survive the real session-switch path. ChatTabView keys the chat route wrapper by pageRouter.chatSessionId, so every switch unmounts the old ChatPage and creates a new setup instance. The watcher therefore never observes an old-to-new id in production, and this Map is discarded before cachedMeasurements can read it; the component test uses wrapper.setProps(), which hides that lifecycle mismatch. Move the snapshots to a longer-lived store/composable or preserve the ChatPage instance, and add a switch test through ChatTabView/the keyed parent.

zhangmo8 (commented):
I found three additional production-path issues that should be addressed before merge. The combined #1973 + #1974 tree merges cleanly and passes typecheck plus the focused skill/message/scroll suites, so these findings are isolated to scroll eligibility, coalesced navigation state, and cache lifetime.

PR #2082, opened 2026-08-03
feat(orchestration): add proactive collaboration

yyhhyyyyyy:
## Summary

Add proactive multi-Agent collaboration to DeepChat while keeping orchestration policy, reasoning effort, and execution strategy independent.

This PR introduces:

- a session-level `explicit | proactive` collaboration policy;
- persistent live delegation with first-class child Sessions;
- a durable JavaScript Workflow runtime backed by QuickJS;
- native approval, progress, recovery, result handoff, and child-session UI;
- shared admission, permission, effect-safety, and lifecycle infrastructure.

The parent Agent can now choose the appropriate execution path per task:

1. execute directly;
2. delegate adaptive tasks to live Subagents;
3. prepare a durable Workflow for large, repeatable, programmatic, or recoverable work.

## Product behavior

- `explicit`: delegation requires an explicit user, project, or Skill instruction.
- `proactive`: the parent may delegate when parallel or isolated work materially improves quality or latency.
- Proactive collaboration does not force delegation for every request.
- Reasoning effort, model settings, permissions, and collaboration policy remain independent.
- Natural language is the user interface; Workflow JavaScript remains internal execution IR.

### Before

```text
Session mode
├── adaptive  → legacy Subagent executor
└── workflow  → durable Workflow executor

Selecting one executor removed the other.
```

### After

```text
Collaboration policy
├── explicit
└── proactive

Parent Agent
├── direct execution
├── live delegation
└── durable Workflow

Reasoning effort remains independent.
```

## Live delegation

- Add durable delegation, turn, and event storage.
- Support spawn, follow-up, list, inspect, wait, interrupt, and paginated result reads.
- Preserve every Subagent as a navigable first-class Session with its own context and Tape.
- Add concise task-first titles and inline child activity cards.
- Allow users to open child Sessions and respond to pending permissions or questions.
- Store bounded semantic handoffs and stable references to immutable child results.
- Reconcile active work after application restart.
- Fence Session deletion against in-flight child creation and drain related work before removal.
- Close child creation, interruption, terminal-state, and result-delivery races.

## Durable Workflow runtime

- Run one synchronous QuickJS WASM sandbox inside an Electron utility process per Workflow run.
- Bridge concurrent guest promises through deferred promises and serialized pending-job draining.
- Expose versioned orchestration helpers:
  - `agent()`
  - `parallel()`
  - `pipeline()`
  - `mapLimit()`
  - `phase()`
  - `log()`
- Persist Workflow runs and invocations in SQLite.
- Use stable logical call paths and immutable input hashes for safe replay.
- Recover queued, interrupted, and partially completed runs after restart.
- Enforce invocation, concurrency, token, duration, source-size, and result-size budgets.
- Freeze generation settings and launch capabilities when a run starts.
- Support structured child output and deterministic validation.
- Add saved Workflows, bounded source outlines, native launch approval, progress projection, retry, resume, and cancellation.
- Deliver completed results to the parent without automatically starting another model turn.

## Reliability and safety

- Add one cancellation-aware global admission layer shared by live delegation, Workflow children, and the legacy orchestrator.
- Record write-ahead tool-effect evidence only after permission authorization and before actual dispatch.
- Treat missing or untrusted effect metadata as `unknown` and fail closed during automatic recovery.
- Keep child Tape lineage frozen and avoid copying full child transcripts into the parent.
- Treat child and Workflow results as untrusted input when the parent synthesizes them.
- Prevent guest access to Node.js, Electron, filesystem, network, shell, credentials, and database APIs.
- Bound utility-process startup, execution, cancellation, and forced termination.
- Preserve MCP App result metadata and the MCP permission contract from `dev`.

## UX

- Add a compact collaboration indicator without replacing the reasoning label.
- Add a native Workflow approval card showing:
  - source outline and hash;
  - target Agents;
  - invocation and budget limits;
  - capability scope;
  - effect-risk warnings.
- Add Workflow and saved-Workflow panels to the workspace side panel.
- Add live Subagent cards with status, result preview, and child-Session navigation.
- Keep advanced model sampling, output, context, timeout, and reasoning defaults available in model settings.
- Preserve unsaved Workflow authoring state across panel and Session navigation.

## Summary by CodeRabbit

* **New Features**
  * Added durable Workflows with sandboxed execution, approvals, saved workflows, retries, cancellation, progress tracking, and result delivery.
  * Added live child-agent delegation with messaging, follow-ups, waiting, interruption, recovery, and activity panels.
  * Added explicit or proactive collaboration controls for supported sessions.
  * Added workflow slash commands, launch approvals, workflow panels, and localized interface text.
* **Bug Fixes**
  * Improved session deletion coordination, delegation lifecycle handling, result safety, usage reporting, and memory projections.
* **Chores**
  * Updated model and agent catalogs and increased renderer test memory limits.


zerob13 (changes requested):
## Blocking: recheck the token budget when admission is granted

The durable token budget is checked only before the invocation is created in [`WorkflowService.resolveInvocation()`](https://github.com/ThinkInAIXYZ/deepchat/blob/85b2263c445bbe6c61bf550e8f2bff57a7a587c1/src/main/workflow/service.ts#L840-L868). A burst from `parallel()` or a high-concurrency `mapLimit()` can therefore create more invocations than the per-owner active limit while they all observe the same pre-settlement usage.

After a queued invocation obtains a global permit, [`executeAdmitted()`](https://github.com/ThinkInAIXYZ/deepchat/blob/85b2263c445bbe6c61bf550e8f2bff57a7a587c1/src/main/workflow/childExecutor.ts#L209-L231) revalidates the launch scope and immediately calls `markInvocationAdmitted()` without reading the current durable usage again. Usage is not aggregated until a child settles ([`childExecutor.ts`](https://github.com/ThinkInAIXYZ/deepchat/blob/85b2263c445bbe6c61bf550e8f2bff57a7a587c1/src/main/workflow/childExecutor.ts#L314-L341)).

A deterministic example is:

1. Configure `maxTotalTokens: 5`.
2. Emit more than four concurrent invocations.
3. The first four pass the initial budget check with usage `0`; the fifth remains queued.
4. One active child settles with `5` tokens and releases its permit.
5. The queued invocation receives the permit and starts another provider call without a budget recheck.

This violates the documented soft-stop contract: once the durable usage reaches the threshold, no new child should be admitted. It also produces a user-visible provider request and additional token usage after scheduling should have stopped.

Please keep the early dispatch-time check, and add a second check after obtaining the admission permit but before `markInvocationAdmitted()`. When exhausted, terminalize the queued invocation with `WORKFLOW_TOKEN_BUDGET_EXCEEDED` without creating a child session or starting a provider call. If admission may later have multiple writers/processes, the durable usage check and `queued -> admitted` transition should be one repository transaction.

Please add a regression test for this exact ordering:

```text
invocation A running
invocation B queued
A settles and exhausts maxTotalTokens
B receives the released permit
B is never admitted and no provider/child execution starts for B
```

The existing test in [`workflowService.test.ts`](https://github.com/ThinkInAIXYZ/deepchat/blob/85b2263c445bbe6c61bf550e8f2bff57a7a587c1/test/main/workflow/workflowService.test.ts#L1210-L1250) covers only the serial case where usage is already durable before the next request is emitted.

## Additional recommendations (non-blocking)

1. **Make the soft-budget semantics explicit in user-facing copy.** `maxTotalTokens` cannot guarantee a hard final total because already admitted children are allowed to settle. Describe it as a scheduling threshold based on settled usage rather than a strict consumption ceiling. A true hard ceiling would require provider-enforced per-invocation limits and transactional reservation, which is not necessary for this V1 fix.

2. **Collapse unreleased intermediate migrations if they were never distributed.** This branch retains multiple intermediate Workflow/session schema transitions. If none of those versions reached users or internal dogfood builds, migrate directly from the last released schema to the final schema and keep only fresh-DB plus released-upgrade coverage. If they were distributed, document that compatibility requirement so the retained migration chain is intentional.

3. **Narrow the authoring-contract claim or enforce parity.** The architecture document says one canonical definition generates or verifies runtime schemas, model schema, examples, validation, and runtime bindings, while the current shared contract mainly centralizes helper signatures/arity and the runtime bindings and wire schemas remain handwritten. Either narrow the documented guarantee to match the implementation or add focused parity tests for the independently maintained surfaces.

PR #2094, opened 2026-08-06
fix(mcp): persist generated images

zerob13:
## Summary

- Cache signed image URLs embedded in MCP text and expose stable imgcache:// references to provider history.
- Resolve exact cached references to data URLs only in cloned MCP execution arguments, preserving persisted tool-call arguments and KV-cache prefixes.
- Suppress duplicate Markdown image nodes when the same durable image block is already rendered, without rewriting stored Markdown.

## Root cause

MCP image generators can return temporary signed image URLs inside ordinary text. Those URLs were neither detected nor cached, so the provider and persisted transcript retained an expiring remote URL. After expiry or restart, the renderer correctly reloaded that URL and received an authorization failure.

## UI behavior

BEFORE

    Assistant response
    ├── promoted local image
    └── matching Markdown image

AFTER

    Assistant response
    └── promoted local image

Unrelated Markdown images and links remain visible.

## Validation

- Focused Vitest suite: 9 files, 280 tests passed
- pnpm run format
- pnpm run i18n
- pnpm run lint
- pnpm run typecheck

The SDD records the persistence contract, execution-only image resolution, security boundaries, and release smoke-test procedure.

## Summary by CodeRabbit

* **New Features**
  * Images returned through tool results can be cached and reused during follow-up actions.
  * Image URLs in text are detected, deduplicated, normalized, and displayed as previews.
  * Cached images are securely resolved when used in tool actions.
  * Duplicate images are hidden when the same image appears in previews and Markdown.
* **Bug Fixes**
  * Improved handling of URL punctuation, supported file types, invalid references, oversized files, and canceled operations.
  * Added safeguards against unsafe cached-image access and restricted private-network caching where appropriate.
  * Prevented failed image resolution from triggering tool actions.


zhangmo8:
Addressed the cache URI compatibility and safety follow-up in 565380d75:

- accept mixed-case `imgcache://` schemes consistently in both argument dispatch and disk resolution
- add resolver and MCP-dispatch coverage for uppercase cache references
- add a regression test rejecting an image-cache symlink that targets a file outside the cache root

Validated locally:
- focused MCP image suite: 9 files, 290 tests passed
- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`

yyhhyyyyyy (changes requested):
The overall approach is coherent: normalize temporary URLs before they first enter provider
history, resolve `imgcache://` only in cloned execution arguments, and deduplicate the
presentation without rewriting persisted Markdown.

I did not find a P0 issue, but I think the following two P1 issues should be addressed before
merge. There are also three P2 correctness/resource issues closely related to the new behavior.

1. **[P1] The automatic download path needs a trust-aware SSRF policy.**

   [`extractStringImagePreviews`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/lib/toolCallImagePreviews.ts#L214)
   now turns any matching URL in MCP-controlled text into a main-process download.
   [`cacheImageFromUrl`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/platform/imageCache.ts#L39)
   accepts HTTP, follows redirects by default, and does not reject loopback, link-local, or
   private-network destinations.

   This gives an untrusted or compromised remote MCP server a client-side SSRF primitive.
   The extension check does not close the boundary because a public `.png` URL can redirect to
   an arbitrary private endpoint. The resulting cache URI can subsequently be resolved to a
   data URL and sent to another MCP call, creating a possible read-and-return path for the
   fetched bytes.

   Please make the policy depend on the MCP server's trust/source, validate every redirect
   target, and reject private destinations for remote servers unless explicitly allowed.
   A public-to-private redirect regression test would be important here.

2. **[P1] Automatic caching is unbounded in both count and response size.**

   The loop in
   [`prepareToolCallImageContent`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/lib/toolCallImagePreviews.ts#L312)
   downloads every distinct matching URL sequentially. Each failed URL may consume the full
   10-second timeout, while axios buffers the complete response as an `arraybuffer` without a
   content-length limit. The abort signal only stops waiting at the caller; it is not passed to
   axios, so the active request may continue and write a file after cancellation.

   A normal search/browser MCP result containing many image links can therefore block a tool
   round for a long time, while a hostile result can cause excessive memory or disk use.
   Please add a maximum image count, per-image and aggregate byte limits, a total time budget,
   and propagate cancellation to the underlying request.

3. **[P2] The cache producer does not preserve the actual MIME type.**

   [`getImageExtensionFromMimeType`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/platform/imageCache.ts#L30)
   stores AVIF, BMP, ICO, `application/octet-stream`, and every unknown content type as `.jpg`.
   The resolver later trusts that extension and emits `data:image/jpeg`, even when the bytes are
   another format. A `200` HTML or JSON response is also treated as a durable image.

   This breaks the stated MIME-correct data URL contract and may make follow-up image tools
   reject otherwise valid cached images. Please validate that the response is a supported image
   and keep the producer/resolver MIME mapping consistent.

4. **[P2] Raw `replaceAll` can rewrite unrelated URLs.**

   [`replaceImageReferences`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/lib/toolCallImagePreviews.ts#L267)
   replaces arbitrary string occurrences rather than the URL tokens identified during
   extraction. For example, after caching `https://host/a.png`, an unrelated
   `https://host/a.png.json` in the same result is changed to
   `imgcache://cached.png.json`.

   Please replace only the exact token spans collected by the extractor and add a regression
   test for URLs where one value is a prefix of another.

5. **[P2] Expanded MCP arguments have no aggregate size limit.**

   [`resolveCachedImageArguments`](https://github.com/ThinkInAIXYZ/deepchat/blob/565380d75299ceb109ee29d307ec744a69807fe5/src/main/mcp/toolManager.ts#L1188)
   enforces an 8 MiB limit per file, but it does not cap the number of references, deduplicate
   repeated reads, or validate the final serialized request size. Repeating a small
   `imgcache://` value several times can expand a small model-generated argument into a
   request of tens or hundreds of MiB.

   Please add per-call memoization and an aggregate encoded-size/reference-count limit before
   invoking the MCP client.

The renderer-side deduplication and the execution-only argument cloning otherwise look sound.
The two P1 items above are merge blockers from my perspective.

yyhhyyyyyy (approved):
LGTM

PR #2097, opened 2026-08-07
fix(cron): stabilize scheduler restart

zhangmo8:
## Summary
- retain the new scheduler utility host when the old host exits after a restart
- persist scheduler spawn failures in the status for diagnostics
- treat a successfully started manual run as started rather than failed

## Validation
- `pnpm exec vitest run --config vitest.config.ts test/main/scheduler/schedulerService.test.ts test/renderer/components/CronJobsSettings.test.ts test/main/scheduler/runSessionStarter.test.ts test/main/session/assignmentPolicy.test.ts`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run typecheck:node` *(timed out after 120s without diagnostics in this environment)*
- `pnpm run typecheck:web` *(timed out after 120s without diagnostics in this environment)*

## UI
Before:
```
[Run now] -> task starts -> Operation failed
```
After:
```
[Run now] -> task starts -> Run now
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated scheduler processes from overwriting the status of active replacements.
  * Improved handling and reporting of scheduler startup failures.
  * Corrected manual job-run notifications for jobs that are starting or still running.
  * Failed or cancelled runs continue to display error notifications, while successful and in-progress runs show appropriate status messages.
  * Improved settings menu reliability.

### DIFF
--- a/src/main/scheduler/schedulerProcessManager.ts
+++ b/src/main/scheduler/schedulerProcessManager.ts
@@ -151,6 +151,10 @@ export class SchedulerProcessManager {
     this.hostReady = this.spawnHost()
     try {
       return await this.hostReady
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error)
+      this.updateStatus({ state: 'error', pid: null, lastError: message })
+      throw error
     } finally {
       this.hostReady = null
     }
@@ -158,9 +162,12 @@ export class SchedulerProcessManager {
 
   private async spawnHost(): Promise<SchedulerUtilityProcess> {
     const host = this.deps.spawnHost ? await this.deps.spawnHost() : await this.spawnDefaultHost()
-    host.on('message', (message) => this.handleHostMessage(message))
-    host.on('exit', (code) => this.handleHostExit(code))
+    host.on('message', (message) => this.handleHostMessage(host, message))
+    host.on('exit', (code) => this.handleHostExit(host, code))
     host.on('error', (type, location) => {
+      if (this.host !== host) {
+        return
+      }
       const message = `Scheduler utility process error: ${type} at ${location}`
       console.error('[CronJobs] Scheduler utility process error:', { type, location })
       this.updateStatus({ state: 'error', lastError: message })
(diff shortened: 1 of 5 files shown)


src/main/scheduler/schedulerProcessManager.ts
```diff
@@ -235,7 +242,10 @@ export class SchedulerProcessManager {
     return candidates.find((candidate) => fs.existsSync(candidate)) ?? candidates[0]
   }
 
-  private handleHostMessage(message: unknown): void {
+  private handleHostMessage(host: SchedulerUtilityProcess, message: unknown): void {
+    if (this.host !== host) {
```
zerob13, line 246: [P1] Preserve persisted `RUN_DUE` events from a replaced host

This guard drops every message from the old host, including `RUN_DUE`. The utility host inserts the run as `queued` and advances `next_run_at` before posting that event. If delivery happens after restart installs the replacement host, the message is ignored; no code scans existing queued rows, so the run can remain queued indefinitely. I reproduced this by emitting `RUN_DUE` from the first host after `restart()`: `onRunDue` was called 0 times. Parse the event first and suppress only stale lifecycle/status events, or explicitly recover queued runs. The existing atomic `claimRun` path already makes duplicate delivery safe.

test/main/scheduler/schedulerService.test.ts
```diff
@@ -1568,6 +1568,78 @@ describeIfSqlite('Cron Jobs persistence and service', () => {
     }
   })
 
+  it('ignores a stale utility exit after restarting the scheduler', async () => {
```
zerob13, line 1571: [P2] Move these process-manager regressions out of the SQLite-gated suite

These tests are inside `describeIfSqlite`. The required CI run reports `schedulerService.test.ts (25 tests | 17 skipped)`; both newly added process-manager tests are in that skipped block, so CI is green without executing them: https://github.com/ThinkInAIXYZ/deepchat/actions/runs/31154698166/job/92791648285. `SchedulerProcessManager` has no runtime SQLite dependency, so place these cases in an ungated test file and add coverage for a stale host delivering an already-persisted `RUN_DUE`.

zerob13 (commented):
Two findings from the restart-lifecycle review. The identity-based approach is otherwise lean, but the message guard is too broad and the new process-manager regressions are skipped in required CI.

PR #2103, opened 2026-08-08
feat(agent): configure output limits

### LINKED ISSUE #2102: [Feature] Make tool-output / file-read truncation thresholds configurable in Settings

**Repo:** ThinkInAIXYZ/deepchat
**Type:** Feature request (enhancement)
**Status:** Draft — not yet submitted

---

## Title

[Feature] Make tool output / file-read truncation & offload thresholds configurable in Settings / 将工具输出与文件读取截断阈值开放为可配置项

### Problem / 问题

Several hardcoded constants in the packaged app control when tool output and file reads get truncated or offloaded to a file. They are tuned for small-context models:

| Constant | Default | Effect |
| --- | --- | --- |
| `READ_FILE_AUTO_TRUNCATE_THRESHOLD` | 4.5 KB | Files larger than this are truncated when read |
| `TOOL_OUTPUT_OFFLOAD_THRESHOLD` | 5 KB | Tool results larger than this are offloaded to a file |
| `COMMAND_OFFLOAD_THRESHOLD` | 10 KB | Background command output offloaded beyond this |
| `FOREGROUND_OFFLOAD_THRESHOLD` | 10 KB | Foreground command output offloaded beyond this |
| `webContentLengthLimit` | 3 KB | Browser/web content truncation |

With 1M-token models (DeepSeek V4 Flash, GPT-5.6-sol, etc.) these defaults waste the context window: a 50 KB source file gets truncated to 4.5 KB, and a 30 KB `ls`/`grep` result gets offloaded to a file that the model can't fully see. Power users on large-context models want more inline content.

 Efficiency  ≠  Effectiveness

  Old defaults (4.5 KB) optimized for low tokens per request
  (= efficient). But truncated output meant the model worked
  with incomplete info — wrong fixes, missed problems, and
  multiple re-prompts.

  Insight: A cheap wrong answer costs more than an expensive
  right one — re-prompt, re-debug, re-fix.

### Current workaround / 现状

To change these, users must manually:

1. Extract `app.asar`
2. Patch the compiled JS constants in `out/main/index.js`
3. Re-pack

…and repeat on **every release** (v1.0.7 → v1.0.8 → v1.0.9 → v1.1.0…). That is an indefinite, fragile patching burden — the exact issue this request wants to eliminate.

### Request / 建议

1. Expose these thresholds as user-configurable values in **Settings → Agent (or Tools)**:
  - Max inline file-read size
  - Max inline tool-output size
  - Max inline command-output size
2. Keep current defaults as defaults (backwards compatible).
3. Alternatively (minimal change): read overrides from `app-settings.json` or an env var, so power users can tune without re-packing the binary.

### Why now / 为什么现在

- **beta.9** already fixed "model context and output limits being overwritten on refresh/reset/restart/backup-restore" — limits clearly matter and are being actively maintained.
- The redesigned usage dashboard now shows input/output/cache-hit tokens — users who manage context budgets will want control over these thresholds.
- Related closed issue **#1327** (terminal output truncation) shows the team is already invested in tool-output handling; this extends that work from "bug fix" to "user control".

### Environment / 环境

- Linux Mint 22.3 (x64)
- DeepChat v1.0.9 (issue also applies to v1.1.0-beta.x line)

zerob13:
## Why

Issue #2102 identifies that hardcoded file-read, tool-output, and command-output limits are too conservative for Agents using large-context models. This change exposes the effective user-visible limits per Agent while keeping internal disk-spooling details and the removed legacy web limit out of Settings.

## Implementation

- Adds three optional fields to `DeepChatAgentConfig`:
  - `readFileAutoTruncateChars` (default: 4,500)
  - `toolOutputInlineChars` (default: 5,000)
  - `commandOutputInlineChars` (default: 12,000)
- Validates integer values in the 1,000-200,000 character range at the config boundary and normalizes missing or defensive in-process values through a shared helper.
- Resolves the current Agent settings at tool-call time, so changes apply to subsequent calls in existing sessions.
- Applies the file-read limit to raw text and prepared-document reads while preserving an explicit `read.limit`.
- Applies the generic inline limit in `ToolOutputGuard`, while `exec` and `skill_run` use the dedicated command preview path.
- Keeps the existing 10,000-character disk-spooling ceiling as an internal upper bound and starts spooling earlier when a configured command preview is smaller.
- Reuses an existing command log during context-pressure fallback instead of creating nested offload files.
- Adds a collapsed Advanced output limits section with reset behavior, localized copy, and visible character units across all supported locales.

The configured limits remain upper bounds. The existing full-request context preflight can still offload or reject results that do not fit the model context.

### Before

```text
Tools
└─ Built-in tool groups and toggles
```

### After

```text
Tools
├─ Built-in tool groups and toggles
└─ Advanced output limits
   ├─ File read      [ 4500 | characters ]
   ├─ Tool output    [ 5000 | characters ]
   └─ Command output [12000 | characters ]
```

Advanced output limits settings

## Compatibility

- Existing Agent configurations remain valid; no migration is required.
- Existing defaults remain effective when the new fields are absent.
- ACP Agents are unaffected.
- Internal command spooling thresholds and the removed legacy `webContentLengthLimit` are intentionally not exposed.

## Validation

- `pnpm format`
- `pnpm i18n`
- `pnpm lint`
- `pnpm typecheck`
- Focused main-process regression suite: 10 files, 479 tests passed
- Agent settings component suite: 1 file, 27 tests passed

Closes #2102

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-agent limits for file reads, tool output, and command output.
  * Added controls to view, edit, reset, and validate these limits.
  * Large outputs can be shortened or stored for later access while preserving existing output files.
  * Applied limits across foreground and background commands, skills, tools, and file reads.
  * Improved cancellation handling and cleanup for deferred output processing.

* **Localization**
  * Added translations across supported languages.

* **Tests**
  * Added coverage for validation, persistence, truncation, output storage, cancellation, and settings behavior.

* **Documentation**
  * Added specifications, implementation plans, and completion checklists.


yyhhyyyyyy (changes requested):
I would mark this as **Request changes**. I found one end-to-end blocker and two correctness issues in the limit-enforcement path.

1. **Blocking: the persisted values are discarded by the runtime resolver.**

   Both `AgentToolManager` and `ToolOutputGuard` consume `resolveDeepChatAgentConfig()`. That path eventually reaches `DeepChatAgentRepository.resolveConfig()`, where `mergeDeepChatConfig()` reconstructs the config from an explicit field list. The three new fields are missing from that list, so `resolveAgentOutputLimits()` always receives `undefined` and falls back to the defaults.

   This also explains why the current tests stay green: the renderer test verifies persistence, while the runtime tests mock `resolveDeepChatAgentConfig()` and therefore bypass the real repository path. Please preserve the three fields in `mergeDeepChatConfig()` and add a repository-level regression that writes non-default values and resolves them through the production config path.

2. **The persisted-output readers do not enforce the configured character cap.**

   The `readLastCharsFromFile()` implementations in `backgroundExecSessionManager.ts`, `skillExecutionService.ts`, and `agentBashHandler.ts` read up to `maxChars * 4` bytes for UTF-8, but return the decoded text without a final character cap. A long single-line ASCII output therefore returns almost four times the configured limit. Since `exec` is no longer covered by the generic 5,000-character guard, this can now flow inline despite the command limit.

   Please cap the decoded tail after partial-line handling and add a regression using a single line longer than four times the configured limit.

3. **The resume path can persist stale output after an awaited fallback.**

   `fitExistingToolOutput()` may perform file I/O, but the `kind === 'ok'` branch updates the message store before re-checking the abort signal or current run scope. A cancellation or superseding run during that await can therefore mutate the transcript from a stale run. Please check both the signal and scope before the first mutation, and clean up any newly created guard-owned offload file if the result is abandoned.

Non-blocking follow-ups:

- The new strings remain English in 14 non-English locale files.
- The deletion ownership difference between `offloadPath` and `existingOffloadPath` is safety-critical but currently encoded only by naming convention.
- The context-pressure loop adds repeated full-context preflights; this is worth profiling for large contexts and large parallel tool batches.

yyhhyyyyyy (changes requested):
I re-reviewed the latest head (`7fb0891`). The persisted-output cap issue and the live input assertions are fixed, and the new scope check prevents the previously reported stale mutation in the non-offloaded success case. I still see two issues that should be addressed before merging:

1. **Blocking: the production config resolver still drops all three settings.**

   `mergeDeepChatConfig()` still does not carry `readFileAutoTruncateChars`, `toolOutputInlineChars`, or `commandOutputInlineChars`. Every runtime consumer obtains these values through `resolveDeepChatAgentConfig()`, so custom values are silently replaced by the defaults. The UI can save and reload the values, which makes the failure especially difficult to notice.

   Please preserve these fields in `mergeDeepChatConfig()` and add a repository-level regression test that persists non-default values and resolves them through the actual production path. The current renderer and runtime tests do not cover this because they either verify raw persistence or mock `resolveDeepChatAgentConfig()`.

2. **The resume fallback lifecycle is still only partially protected.**

   The awaited fallback is not tied to `preStreamAbortSignal`, and not all file side effects are guarded by cancellation and scope freshness. In the `tool_error` and `terminal_error` branches, `budgetToolCall.offloadPath` is deleted before `scope.assertCurrent()`. If the session instance is replaced during the await, the stale run can therefore delete a file still referenced by the already-persisted tool result.

   Conversely, if a successful fallback creates a new `resumeBudget.offloadPath` and the subsequent scope check fails, that newly created file is abandoned without cleanup. The new test only covers `kind: 'ok'` with `offloaded: false`, so neither ownership case is exercised.

   Please make the fitting step abort-aware, validate cancellation and scope freshness before cleanup or persistence, clean up newly created guard-owned fallback files when their result is abandoned, and never delete the tool-owned `existingOffloadPath`. Focused tests should cover cancellation, stale replacement with an offloaded success result, and stale replacement in both error branches.

Non-blocking: the ten new strings are still left in English in 16 non-English locale files.

Because the first issue means the configurable values still have no runtime effect, I would keep this as **Request changes**.

yyhhyyyyyy (approved):
LGTM

PR #2104, opened 2026-08-08
feat(tape): add durable execution journal

yyhhyyyyyy:
## Summary

Add a durable Execution Journal to the existing Tape store so DeepChat can reliably classify tool execution and Run state after process crashes.

## Changes

- Add immutable journal facts for Run start, dispatch, tool outcome, and Run terminal state.
- Give each physical Run a UUID and each tool operation a structured identity.
- Enforce strict idempotency: identical facts are idempotent; conflicting facts fail as corruption.
- Commit dispatch facts after all local refusal gates and immediately before side effects.
- Commit tool outcomes and Run terminal facts before transcript, context, status, hook, or UI projection.
- Apply the same lifecycle contract to normal, deferred, MCP, and built-in Agent tool execution.
- Classify startup recovery as `not_dispatched`, `completed`, `indeterminate`, or `corruption`.
- Park indeterminate and corrupt executions without automatically retrying tools.
- Keep journal facts out of Context Tape views, search, recall, and fork merge paths.
- Preserve Context Tape compatibility while making message revision and derived tool ordering explicit.
- Add bounded recovery scanning, fail-closed cleanup, and connection-reopen handling.

## Failure Semantics

This does not claim exactly-once execution for arbitrary external systems. A crash after remote acceptance but before outcome persistence can still be indeterminate.

The journal instead guarantees that uncertainty has a durable, classifiable boundary and is never silently converted into success or automatic replay.

## Summary by CodeRabbit

- **新功能**
  - 增强长任务执行记录与恢复能力，异常中断后可安全识别并暂停处理。
  - 工具调用支持更严格的执行顺序与结果确认，降低重复操作风险。
  - 浏览器、文件、记忆、技能及自动化任务等操作增加执行前校验。
- **错误修复**
  - 修复暂停任务、取消操作和终止状态处理不一致的问题。
  - 改进会话清理、消息恢复及工具结果展示的一致性。
- **文档**
  - 新增持久化执行记录、恢复流程与故障处理规范。

### DIFF
--- a/docs/architecture/agent-system.md
+++ b/docs/architecture/agent-system.md
@@ -1,6 +1,6 @@
 # Agent 系统
 
-本文描述 `2026-07-24` 的当前合同。历史迁移方案、Presenter 拆分过程和已完成的分阶段
+本文描述 `2026-08-07` 的当前合同。历史迁移方案、Presenter 拆分过程和已完成的分阶段
 SDD 已删除；需要时从 Git 历史查询。
 
 ## 两种 Agent backend
@@ -53,7 +53,8 @@ flowchart TD
 - Session 拥有长期身份、settings、transcript、Tape 和 pending input。
 - DeepChat Agent 拥有自己的配置和物理 Skill root；Session 只保存 Agent assignment 和所选 Skill 名称。
 - 已载入 Session 在一个 backend 中最多有一个 instance。
-- 每次执行创建独立 Run，Run 拥有取消信号、logical round、request sequence、physical attempt 和临时输出。
+- 每次执行创建独立 UUID Run，Run 拥有取消信号、logical round、request sequence、physical attempt 和
+  临时输出；loop、resume 和 deferred tool execution 都不得复用旧 Run identity。
 - Window、Remote endpoint 和 Cron run 只保存各自 binding，不拥有 Agent instance。
 - 只读 list/query 不 hydrate backend；执行、完整 restore 或明确 backend 设置操作才允许 hydrate。
 
@@ -80,6 +81,10 @@ src/main/agent/deepchat/
 - `requestSeq` 标识一份确定的 provider payload 和 ViewManifest；`physicalAttempt` 标识该 request 的
   实际发送次数。context recovery 改变 payload，因此推进 requestSeq 并把 physicalAttempt 重置为 1；
   同 payload 的 transient retry 保持 requestSeq，只推进 physicalAttempt。
+- 工具 operation identity 为 `(runId, requestSeq, providerToolCallId)`。`run_started` 必须先于 Run
+  registration；dispatch fact 必须位于所有本地拒绝 gate 之后、真实副作用调用之前；known outcome
+  必须先于 result projection；terminal fact 必须先于 transcript、status、hook 和 renderer terminal
+  projection。任一 strict Journal commit 失败都阻止下游边界继续。
 - `DeepChatContextCoordinator.streamProviderAttempts` 是 transient retry 的唯一 owner。每个 logical
   round 最多重试两次，使用可取消的指数退避并服从有上限的 `Retry-After`；context recovery 不消耗
   logical-round 或 transient-retry budget，每次实际 attempt 都重新进入 provider rate gate。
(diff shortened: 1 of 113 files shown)


zerob13 (commented):
## Complexity and test-value review

I would not merge this PR in its current shape. The crash-boundary direction is useful and the
boundary/real-process tests are valuable, but the implementation is not yet the smallest coherent
change: the PR is 111 files and `+11,382/-827`, mixes a durable journal with unrelated Context Tape
repairs, and puts an unbounded amount of historical work on application startup.

### Findings

1. **[P1] Recovery performs an unbounded full-history scan on the synchronous startup path.**

   `ExecutionJournalService.classifyRecoveryCandidates()` reads every journal event through
   `listEventsByNames()` (`src/main/tape/application/executionJournalService.ts:163`), whose SQL has
   no run-state predicate, checkpoint, or limit
   (`src/main/tape/infrastructure/sqlite/tapeEntryStore.ts:642`). The classifier then materializes
   every Run and reparses every payload; outcome parsing also hashes the persisted response again.
   `MAX_STARTUP_RECOVERY_DETAILS` only bounds logging, not scanning. With response payloads allowed
   up to 256,000 characters, launch cost grows with total lifetime journal bytes rather than unresolved
   work. This also contradicts the PR description's “bounded recovery scanning” claim.

   The minimal V1 is to query only Runs without a terminal fact and load their related facts. If a
   permanent global corruption audit is required, run it outside synchronous startup or add a durable
   acknowledgement/retention design before landing this path.

2. **[P2] V1 persists outcome data that no production recovery path consumes.**

   `ExecutionToolOutcomeFact` stores `responseText` and `offloadPath`
   (`src/main/tape/domain/executionJournal.ts:112`), but the classifier reduces an outcome to
   `{ entryId }` (`:708`, `:762`) and startup recovery only marks the pending message interrupted.
   There is no production reader that replays or projects either field. Normal cleanup can also delete
   the referenced offload file, leaving a durable stale path. This duplicates sensitive tool output,
   increases database size and startup hashing cost, and implements a future recovery format without
   implementing that recovery.

   For V1, persist only the operation identity, `isError`, and a response hash if idempotency needs it.
   Add durable response recovery in a separate change when there is an actual consumer and retention
   contract.

3. **[P2] Split the Context Tape ordering fixes out of the journal PR.**

   Acceptance criteria 18-20 in
   `docs/architecture/durable-execution-journal/spec.md:254` cover compaction provenance, message
   replacement semantics, tool ordering, and backfill revisions. Commits `f0e045ed3`, `58c0189a1`,
   and `e718512bd` account for `+989/-149` lines and are independently reviewable Context Tape fixes;
   they are not required to establish T0/T1/T2/T3 journal boundaries. Combining them makes failures
   harder to attribute and forces journal reviewers to validate a second storage model. Move those
   commits and their tests to a separate PR.

4. **[P3] The missing-dispatch gate should be a type contract, not a runtime branch and test.**

   `ToolExecutionPort.execute()` still accepts optional options, so
   `createToolExecutionPort()` checks `options?.commitDispatch` at runtime
   (`src/main/agent/deepchat/runtime/toolAdapters.ts:61`). The only production callers of this port are
   normal dispatch and deferred dispatch, and both always provide the callback. Define the internal
   execution options as `ToolCallOptions & { commitDispatch: ToolDispatchCommit }`, require them on
   `ToolExecutionPort.execute`, and delete this guard plus its dedicated test. Lower-level/direct tool
   APIs may keep optional callbacks where non-journal execution is a real use case.

### Gate assessment

The T1/T2 prerequisite checks, duplicate-operation checks, local preflight-before-T1 checks, and the
host-transaction prohibition protect different failure windows; they are not mutually redundant.
The redundant gate is the runtime capability-presence check above because the internal port can make
the invalid state unrepresentable. The broader callback plumbing is costly (new `beforeMutation` /
`beforeInvoke` variants span 23 source files), but most of it exists because side-effect preflight is
owned deep in each subsystem. Replacing it safely would require one uniform prepare/commit/invoke
contract, not another layer of booleans.

### Test assessment

The test total (`+6,065/-138`) exceeds the production-source addition (`+4,455/-676`), but the ratio
alone is not the problem. Keep the native crash tests and the dispatch/outcome/terminal ordering tests;
they protect observable durability contracts. The avoidable bulk is mainly implementation proof:

- `deepChatAgentHarness.test.ts:2317-2460` spends about 140 lines on log caps, priority, and string
  sanitization through full harness construction. Build classifier reports directly and collapse this
  to one table-driven diagnostic test.
- `deepChatAgentHarness.test.ts:2928-3006` repeats nearly the same fallback-error scenario for typed,
  corruption, and untyped errors. One table-driven contract test can preserve subtype/cause assertions.
- The dedicated missing-`commitDispatch` adapter test disappears when the port type is corrected.
- Keep the background-process ownership and Context Tape regression tests with the separate PRs that
  own those behaviors; they are useful tests, but they inflate and couple this journal review.

### Minimal-cut summary

`src/main/tape/application/executionJournalService.ts:L163: scale: replace the lifetime scan with an unresolved-Run recovery query.`

`src/main/tape/domain/executionJournal.ts:L112: yagni: remove raw response/offload persistence until recovery consumes it.`

`src/main/agent/deepchat/runtime/toolAdapters.ts:L61: type: require commitDispatch on the internal execution port and delete the runtime gate.`

`docs/architecture/durable-execution-journal/spec.md:L254: scope: move Context Tape compaction/order repair to its own PR.`

`docs/architecture/durable-execution-journal/plan.md:L1: artifact: keep the final spec; drop the 375-line plan and 111-line task ledger, moving open follow-ups to an issue.`

`net(PR scope): at least -1,300 lines possible before parser and test deduplication.`

Validation at `f9098f6513a317dcd265709d0959935f0c1ec0ff`: required GitHub checks are green; focused
Vitest coverage passed 36 tests, with 11 environment-gated native crash cases skipped locally.

PR #2105, opened 2026-08-08
fix(browser): prevent macOS PiP black window

zerob13:
## Summary

Fix the macOS black fullscreen window caused by the Browser PiP off-screen host, and reduce the host's background resource cost.

## Root cause

Browser PiP keeps the existing `WebContentsView` renderable by attaching it to a separate transparent, off-screen `BaseWindow`. The host was created with `show: true`. On macOS, that independent native window could participate in fullscreen window handling despite being transparent and positioned off-screen, surfacing as a black fullscreen window.

## Changes

- Keep the preview `BaseWindow` hidden and non-fullscreenable on macOS.
- Release the hidden preview host when an Agent run becomes `idle` or `error`:
  - stop the capture loop
  - destroy the host window
  - restore Electron background throttling
  - release the preview ownership claim
- Guard terminal cleanup by Agent run ID so cleanup from an old run cannot destroy a newly started run's host.
- Destroy the session browser during session deletion, including best-effort cleanup when another deletion stage fails.
- Reduce preview capture cadence:
  - active: `250 ms -> 500 ms` (`4 FPS -> 2 FPS`)
  - idle: `1000 ms -> 2000 ms` (`1 FPS -> 0.5 FPS`)
- Reduce the encoded PiP frame from `480x300` to `400x250` (about 31% fewer output pixels).
- Keep the browser viewport at `1280x800` so page layout, CDP coordinates, and automation behavior do not change.

The smaller output frame reduces resize, JPEG encoding, IPC, and renderer drawing work. It does not reduce the source viewport captured by Electron.

## Lifecycle boundary

The hidden host still exists while an Agent is actively using a background browser because PiP needs a renderable surface. It is now destroyed at the terminal session state. Expanding the Browser side panel continues to detach the view from the hidden host through the existing attach path.

## Behavior

```text
BEFORE
macOS fullscreen -> transparent BaseWindow joins fullscreen handling -> black window

AFTER
macOS fullscreen -> hidden, non-fullscreenable preview host -> normal app fullscreen

Agent running -> hidden preview host -> reduced-rate capture
Agent idle/error -> stop capture -> destroy host -> background throttling restored
```

## Testing

- `pnpm exec vitest run --config vitest.config.ts test/main/desktop/browser/YoBrowserPresenter.test.ts test/main/session/deletion.test.ts` (24 passed)
- `pnpm exec vitest run --config vitest.config.renderer.ts test/renderer/components/AgentBrowserPiP.test.ts` (11 passed)
- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`

The application demo was not run.

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced browser preview size and adjusted capture intervals for smoother, more efficient operation.
  * Inactive previews are now released when sessions become idle or encounter errors.

* **Bug Fixes**
  * Browser resources are cleaned up when sessions are deleted, even if earlier cleanup steps fail.
  * macOS previews remain hidden and cannot be switched to fullscreen unnecessarily.

* **UI Improvements**
  * Updated the message copy control for consistent button sizing and behavior.

### DIFF
--- a/src/main/app/composition.ts
+++ b/src/main/app/composition.ts
@@ -1427,7 +1427,17 @@ export async function createMainProcessControl(dependencies: {
     toolService,
     hookObserver: hookService,
     publishEvent: publishDeepchatEvent,
-    publishSessionUpdate: (update) => sessionRuntimeEvents.publish(update),
+    publishSessionUpdate: (update) => {
+      sessionRuntimeEvents.publish(update)
+      if (update.kind === 'status' && (update.status === 'idle' || update.status === 'error')) {
+        void yoBrowserPresenter.releaseInactivePreview(update.sessionId).catch((error) => {
+          logger.warn('[YoBrowser] Failed to release inactive preview', {
+            sessionId: update.sessionId,
+            error
+          })
+        })
+      }
+    },
     providerCatalogPort,
     sessionPermissionPort,
     acpAsLlmProviderPermission: acpAsLlmProviderPermission,
@@ -1575,7 +1585,9 @@ export async function createMainProcessControl(dependencies: {
     },
     runtime: {
       cleanupSessionBackends: async (sessionId) =>
-        await agentManager.cleanupSessionBackends(sessionId)
+        await agentManager.cleanupSessionBackends(sessionId),
+      destroySessionBrowser: async (sessionId) =>
+        await yoBrowserPresenter.destroySessionBrowser(sessionId)
     },
     state: deepChatAgentHarness,
     permissions: sessionPermissionPort,
--- a/src/main/desktop/browser/YoBrowserPresenter.ts
+++ b/src/main/desktop/browser/YoBrowserPresenter.ts
@@ -1355,7 +1355,7 @@ export class YoBrowserPresenter implements IYoBrowserPresenter {
         y: -10000,
         width: PREVIEW_VIEWPORT.width,
         height: PREVIEW_VIEWPORT.height,
-        show: true,
+        show: !isMac,
         fullscreenable: !isMac,
         opacity: 0,
         focusable: false,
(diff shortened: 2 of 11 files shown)


yyhhyyyyyy (changes requested):
I found two issues that should be addressed before this is merged:

1. **The hidden macOS host is not a valid capture target.**

   With the repository's Electron 41.10.4 on macOS 27.0 arm64, a minimal
   `BaseWindow`/`WebContentsView` using the same `show: false` configuration
   loads successfully, but `capturePage(rect, { stayHidden: true })` rejects
   with `Current display surface not available for capture`.

   Electron checks `IsSurfaceAvailableForCopy()` before incrementing the
   capturer count, so `stayHidden` cannot create a compositor surface for a
   window that never had one:
   https://github.com/electron/electron/blob/v41.10.4/shell/browser/api/electron_api_web_contents.cc#L3793-L3806

   In the current loop this means no PiP frames on macOS, plus repeated capture
   warnings while the Agent remains active. The unit test cannot catch this
   because `capturePage` is mocked. Please preserve a renderable surface, for
   example by keeping the off-screen host shown while setting
   `fullscreenable: false`, or validate another approach with a real Electron
   smoke test on supported macOS versions.

2. **A timed-out capture can permanently block preview capture for the next run.**

   `stopPreviewCapture()` returns after 500 ms but leaves the old
   `state.previewCapture` promise installed when the timeout wins.
   `resumeClaimedPreview()` will not schedule another frame while that field is
   non-null. I reproduced this with a first capture that never settles:
   terminal cleanup returns, run 2 claims and recreates the host, and
   `setPreviewMode('capturing')` succeeds, but no second `capturePage` call is
   ever made.

   Please detach the stale capture slot when the timeout wins and add a
   regression test covering a new run after a non-settling capture. The
   existing epoch checks can still fence any late completion from the old run.

yyhhyyyyyy (approved):
LGTM

PR #2129, opened 2026-08-10
fix(chat): recover pending inputs after restart

### LINKED ISSUE #2111: [bug] Wake durable pending inputs after restart and session hydration
## Summary

Durable pending inputs can remain in the database indefinitely after an application restart because startup recovery repairs claimed rows but never schedules the pending-input pump after the owning Session is hydrated.

This is a reliability defect, not a permission-waiting problem. The affected rows are ordinary durable Queue/Steer inputs that have no active blocker and no runtime owner.

**Priority:** P1 (potentially P0 for input-loss perception)

## Observed behavior

Sanitized development data contained multiple pending-input records that:

- remained in `state = 'pending'` across clean restarts;
- had no linked user-message IDs or assistant-message reservation;
- had no attachment or interaction blocker;
- were not consumed when the corresponding application version, and later the current `dev` source, started.

The user-visible result is a message that appears to have been accepted but never starts and has no recovery indication.

## Root cause

The composition sequence in `src/main/agent/deepchat/harness/createDeepChatAgentHarness.ts` calls:

1. `recoverClaimedInputsAfterRestart()`;
2. `recoverPendingMessages()`;
3. returns the completed runtime graph.

It does not wake `PendingInputPump` for Sessions that already contain durable pending rows.

There is a second ordering constraint: `PendingInputPump.drain()` in
`src/main/agent/deepchat/runtime/pendingInputPump.ts` requires an existing hydrated scope and returns `false` when no scope exists. Therefore, blindly scheduling every Session during harness construction would also lose the wakeup before hydration.

Current wakeups are driven by live enqueue/edit/delete/retry/completion paths. A row that predates process startup may never cross any of those paths.

- `docs/architecture/session-management.md` states that active Session restore must wake a backend for a pending Steer and identifies Session data as the durable source of truth.
- `docs/features/im-style-steer-messages/spec.md` requires pending state and receipt state to remain correct across process restart.
- `docs/architecture/deepchat-agent-harness-boundaries/spec.md` assigns pending-input recovery, single-flight drain, claim/release/consume, and turn start ownership to `PendingInputPump`.

The implementation currently has durable persistence and pump ownership, but it is missing the hydration-time bridge between them.

## Impact

- Users can believe an input was sent while no runtime will ever process it.
- Restart does not repair the condition.
- The stale row can remain invisible until the Session is inspected manually.
- Support and diagnostics may misclassify this as a provider stall because no explicit failure is produced.

## Proposed direction

- Add a hydration/open-time check that schedules the pump when the Session is idle, has a pending turn input, and has no interaction blocker.
- Make the wakeup idempotent and preserve the existing instance-scoped single-flight lease.
- Do not automatically execute arbitrarily old inputs on cold startup. Mark stale rows as requiring explicit user recovery, or apply a documented age policy.
- Surface a recovery state in the Session/sidebar UI with explicit **Retry** and **Delete** actions.
- Emit a content-free diagnostic when a wakeup cannot proceed because hydration or another owner is missing.

## Acceptance criteria

- [ ] A pending Queue input persisted before shutdown is admitted exactly once after the Session is restored and hydrated.
- [ ] A pending Steer follows its documented restart recovery contract.
- [ ] Scheduling before hydration is not silently lost; the hydration boundary performs or replays the wakeup.
- [ ] Interaction blockers and active runs continue to prevent unsafe admission.
- [ ] Stale pending inputs are visible and require an explicit decision instead of executing unexpectedly.
- [ ] Restart, hydration ordering, duplicate wakeup, and exact-once regression tests are added.

## User benefit

Messages no longer disappear into an invisible durable queue. After a restart, users either see the input resume safely or receive a clear, actionable recovery choice.

zerob13:
## Summary

- reconcile durable Queue and Steer inputs deterministically during cold startup
- keep restored Queue items in the existing lane until the user explicitly resumes them
- terminalize interrupted Steer attempts without recovery-specific failure or retry chrome
- make pending-input listing a pure read so opening a Session cannot execute historical input
- prevent a manually resumed Queue head from returning after a provider error

## Root cause

Durable rows in `deepchat_pending_inputs` survived a process restart, but their in-memory scheduling edge and user-visible disposition did not. Queue and Steer were also treated as equivalent pending work even though they represent different user commitments. A later read or lifecycle wake could therefore execute an old input, while a failed manually resumed Queue claim could be released back to `pending` and reappear on the next Session entry.

### Before

```text
Restored Session
+--------------------------------------------------+
| You  10:21  Unread / Failed                      |
| Please change the output format   [Retry]        |
|                                                  |
| Queue (2)                                        |
|  1. Add a short example          [Edit] [Delete] |
|  2. Translate it to Chinese      [Edit] [Delete] |
+--------------------------------------------------+

Opening or listing the Session could wake historical input.
Resume queue -> provider error -> row returns to pending.
```

### After

```text
Restored Session
+--------------------------------------------------+
| You  10:21                                       |
| Please change the output format                  |
|                                                  |
| Queue (2)                         [Resume queue]  |
|  1. Add a short example          [Edit] [Delete] |
|  2. Translate it to Chinese      [Edit] [Delete] |
+--------------------------------------------------+

Opening the Session is read-only.
Resume queue -> provider boundary -> consume durable row.
Provider errors remain in the transcript and do not restore the Queue item.
```

The existing message toolbar remains available for normal retry. No Steer-specific failure receipt or recovery action is added.

## Implementation

- Reconcile active pending inputs before any Session runtime is restored.
- Keep pending Queue rows durable and add their IDs to a process-local restart hold.
- Release that hold only through the typed `sessions.resumePendingQueue` route.
- Preserve the existing Steer-before-Queue priority, Queue FIFO ordering, instance lease, and durable claim fences.
- Consume a manually resumed Queue claim immediately before the provider stream starts.
  - Failures before that boundary release the claim, preserving the accepted draft.
  - Returned provider errors after that boundary remain transcript failures and cannot requeue the same item.
- Convert an unclaimed restart Steer into an internal terminal `error` transcript message, consume its active row, and hide the receipt in the renderer.
- Remove the scheduling side effect from pending-input listing.
- Add focused main-process and renderer regressions for restart reconciliation, explicit resume, duplicate actions, failure boundaries, and Steer presentation.

## Impact

- Historical user intent never executes merely because the app starts or a Session hydrates.
- Restored Queue items remain visible, editable, reorderable, and deletable until explicit resume.
- A successful manual resume cannot leave the same Queue item behind after provider failure.
- Interrupted Steer content stays in transcript history but is excluded from later model context until normal retry.
- No database schema or migration is introduced.
- Live Queue FIFO, Steer priority, attachment handling, provider execution, permission behavior, Session summaries, and sidebar behavior are unchanged.

## Validation

- `pnpm format`
- `pnpm i18n`
- `pnpm lint`
- `pnpm typecheck`
- focused main-process suites: 515/515 passed
- focused renderer suites: 34/34 passed
- full renderer suite: 2056/2056 passed

The full main-process run has 60 pre-existing Windows-environment failures across 27 unchanged files, dominated by POSIX path expectations, symlink permissions, executable-bit checks, and packaging fixtures. All affected main-process suites pass.

Closes #2111

## Summary by CodeRabbit

- **New Features**
  - Added a **Resume queue** action for queued messages retained after restarting the app.
  - Queued messages now wait for explicit confirmation instead of running automatically.
  - Added localized labels and failure notifications for queue resumption.

- **Bug Fixes**
  - Improved recovery of interrupted Steer messages after restart.
  - Failed resumed messages are no longer duplicated or returned to the queue.
  - Pending-input views no longer trigger unexpected execution during session restoration.
  - Restart-failed Steer messages now support the standard retry flow without showing misleading receipt controls.


yyhhyyyyyy (changes requested):
I agree with the overall direction: pending-input listing should be a pure read, historical Queue items should require an explicit resume after restart, and an unclaimed Steer should become a terminal transcript fact rather than being executed by a later hydration/listing edge.

I found three restart-safety issues that I think should block the merge:

1. **A claimed Queue does not durably record that its user message was materialized.**

   Startup recovery decides whether to consume or release a claimed Queue from `input.messageIds.length`, but the normal `appendUserFact` path in `TurnCoordinator` creates/adopts the user message without linking it back to the claimed pending-input row. `linkClaimedQueueMessage()` appears to be used only by the pre-stream Steer handoff path.

   If the process exits after the user message is persisted (or even after provider dispatch) but before the Queue claim settles, the next startup sees `messageIds: []`, releases and holds the row, and also recovers the existing transcript attempt as an error. Resuming or retrying can then execute the same input again.

   This may predate the PR, but it directly invalidates the new claimed-Queue recovery contract. Please make user-message materialization/adoption and the claimed-row association durable in the same transaction, or establish another authoritative recovery marker. A follow-up link in a second write would retain the same crash window.

2. **Pending Steer terminalization is not crash-atomic.**

   `recoverInputsAfterRestart()` consumes the Steer row in one transaction and returns its linked message IDs through an in-memory `forceRecoverMessagesBySession` map. The harness calls `recoverPendingMessages()` only afterwards.

   If the process exits or recovery throws after the Steer row is consumed but before every linked user message is changed to `error`, the next startup cannot repair the state: the row is no longer active, while an unforced pending user message with Steer receipt metadata is intentionally kept pending. This leaves a permanent `Unread` message with no pending row behind it.

   Please terminalize all linked Steer user messages and consume the Steer row in the same database transaction. A fault-injection test should restart again after the first commit / after the first message of a merged Steer and verify that every message is `error` and no active row remains.

3. **Manual Resume semantics are lost when attachment preparation blocks.**

   `consumeClaimBeforeProviderStream` is derived only from the current wake reason. A manually resumed item that returns `needs_user_action` becomes `blocked`; resolving it schedules an ordinary `enqueue` wake, so the next launch no longer has consume-before-provider enabled. If that provider attempt returns an error, the item is rolled back and released to pending, even though it is still the same explicitly resumed Queue head.

   Please preserve the manual-resume disposition by item ID across `blocked -> pending -> claimed`, and clear it only when the item is consumed, deleted, or promoted. Please add coverage for both attachment Retry and “send without image content,” followed by a provider error.

One non-blocking scope issue: the renderer shows Resume for any idle pending Queue, and the backend ignores whether `releaseRestartHoldForSession()` actually released a hold. This applies the new manual semantics to ordinary live Queue rows as well. At minimum, the route should return `started: false` when no restart hold was released; ideally the renderer should receive an authoritative `resumeAvailable` signal.

The existing CI is green, but the current tests do not cover these crash boundaries or the blocked-attachment resume sequence.

zerob13:
Addressed the two applicable CodeRabbit nitpicks in `10e8e0ce9`:

- Moved the provider-start pending-row assertion out of the `coreStream` mock so a failure cannot be absorbed by the stream error path.
- Added focused dispatcher coverage for `sessions.resumePendingQueue`.

No change was made to the direct fake-store reset in `pendingInputPump.test.ts`. `PendingInputPumpStorePort` intentionally does not expose a blocked-input reset/resolve operation; that transition is owned by `PendingInputAdmissionCoordinator.resolveBlockedPendingInput`. Adding a reset method to the pump solely for this unit test would widen the production port and couple the pump to an admission-layer responsibility. The test therefore retains the local state transition, matching the review's condition to keep it when no suitable pump reset entry point exists.

yyhhyyyyyy (approved):
LGTM

PR #2131, opened 2026-08-10
refactor(dc-ui): extract DcCopyButton and wire copy-text

zhangmo8:
## 概述

1. **DcCopyButton 归位到 button 模块**：由 `components/copy-button` 并入 `components/button`（旧目录删除），`props.ts` 承载 DcButtonProps，dc-ui 顶层统一走 `components` barrel 导出。
2. **复制流程修复**：组件内部已持有 `copied` 状态，但此前 `copy-text` 未传入，复制流程根本没启动，icon 永远停在 copy。现在所有消费方均经 `copy-text` 传入复制内容，成功态（✅ + emerald，1200ms 回落）恢复生效。
3. **MessageToolbar**：新增 `copy-text` prop；用户/助手消息复制内容由父级（MessageItemUser / MessageItemAssistant）计算后传入（原 `handleCopy` 逻辑抽成 `copyText` computed，行为不变）。
4. **DcToggleRow**：行内开关补齐 `w-full`，保证开关靠右对齐。
5. **Composer 草稿持久化**：`composerDraftPersistence` + `useComposerSubmit` 接入 + 对应测试。
6. **docs/design-system.md**：组件表补回 `DcCopyButton` 条目（新导入路径 `@dc-ui/components/button`），迁移状态同步登记。

## 涉及文件

- dc-ui：button（DcCopyButton / props / index）、index barrel、toggle-row
- renderer：MessageToolbar、MessageItemUser/Assistant、ArtifactBlock、CodeArtifact、McpJsonViewer、McpServers、TraceDialog、MessageBlockToolCall、useComposerSubmit、composerDraftPersistence
- 测试：ChatPage、useComposerSubmit、composerDraftPersistence
- 文档：docs/design-system.md

## 说明

- 未跑测试（用户明确要求不需要）

## Summary by CodeRabbit

- **New Features**
  - Added a reusable copy button with accessible labels, visual success feedback, and copy/error events.
  - Composer drafts are automatically saved per conversation and restored when switching sessions or returning to the page.
  - Expanded top-level access to UI components.

- **Bug Fixes**
  - Improved copy-text handling for messages, including reasoning content.
  - Updated toggle-row layout to use the full available width.

- **Tests**
  - Added coverage for copy-button behavior and draft persistence, restoration, validation, cleanup, and storage failures.


zerob13 (commented):
Conclusion: this is not the smallest correct implementation, and I would not merge it as-is.

The copy defect is narrow: relocate the existing component mechanically, pass `copyText` from `MessageToolbar`, preserve the public contract, and add focused regression coverage. Instead, this PR also adds an unrelated composer-persistence feature and a toggle-row layout change, creates a broad `@dc-ui/components` barrel, extracts button props solely to support a render-function proxy, and rewrites `DcCopyButton` while dropping `showToast` / `successDuration` and changing clipboard error semantics. That is unnecessary scope and abstraction for the stated PR.

Recommended split:

1. Copy-button PR: keep the template implementation and existing API, move/re-export it from `components/button`, update direct imports, wire `copy-text`, and test the observable copy state.
2. Composer-draft PR: implement and validate persistence independently, including the real keyed `ChatPage` mount path, bounded attachment storage, session deletion/cleanup, and corrupted/quota cases.
3. Keep the `DcToggleRow` layout fix separate or justify it in the PR scope.

The new broad component barrel is not needed for this move; `@dc-ui/components/button` already provides the correct ownership boundary and avoids making every consumer depend on a package-wide export surface.

UI regressions introduced by the rewrite:

```text
Expected focus: [Copy*] --Enter--> [Check*]
Current focus:  [Copy*] --Enter--> [Check]  -> focus moves to <body>

Expected toolbar: [12 px icon inside 16 px control]
Current toolbar:  [16 px icon fills 16 px control]
```

The seven existing unresolved CodeRabbit threads are still actionable; I did not duplicate them below. The inline comments in this review are additional findings.

Validation performed at this head: 143 relevant Vitest tests passed; format, lint, web typecheck, and i18n passed; GitHub PR Check and Package Check are green. The local runtime was Node 24.15.0 while the repository requires >=24.18.0, so local commands emitted an engine warning. The green tests do not cover the actual keyed remount restore path described below.

src/renderer/src/features/chat-page/composables/useComposerSubmit.ts
```diff
@@ -580,18 +647,34 @@ export function useComposerSubmit(options: UseComposerSubmitOptions) {
...
       )
       sessionDrafts.set(outgoingSessionId, outgoingDraft)
       observedDraftFingerprints.set(outgoingSessionId, composerDraftFingerprint(outgoingDraft))
+      saveComposerDraftToStorage(outgoingSessionId, outgoingDraft)
     }
 
     const storedIncomingDraft =
-      sessionDrafts.get(sessionId) ?? createEmptyComposerDraft(draftRevisions.get(sessionId) ?? 0)
+      sessionDrafts.get(sessionId) ??
+      loadComposerDraftFromStorage(sessionId) ??
```
zerob13, line 676: [P1] Restore persisted drafts on the real mount path

`loadComposerDraftFromStorage()` is only reached through `switchComposerSession()`. In `ChatPage.vue`, however, the immediate session watcher runs before `useComposerSubmit()` is constructed, so it calls the initial no-op `switchComposerSessionDraft`; the real function is assigned only afterward. `ChatTabView.vue` also keys the route shell by session ID, so every session switch creates a fresh `ChatPage` and that watcher never runs again on the instance. A saved draft is therefore not restored when the user reopens the session, and the next `dispose()` can flush the empty live state and remove the stored key. Initialize the draft from storage during composable construction (or explicitly invoke the assigned function once after construction) and test the actual `ChatPage` remount without manually calling `switchComposerSession()`.

src/renderer/src/features/chat-page/model/composerDraftPersistence.ts
```diff
@@ -0,0 +1,78 @@
...
+export function saveComposerDraftToStorage(sessionId: string, draft: ComposerSessionDraft): void {
+  if (!hasStorage() || !sessionId) {
+    return
+  }
+  try {
+    if (isComposerDraftEmpty(draft)) {
+      window.localStorage.removeItem(storageKey(sessionId))
+      return
+    }
+    window.localStorage.setItem(storageKey(sessionId), JSON.stringify(draft))
```
zerob13: [P1] Do not store raw attachment payloads in localStorage

`ComposerSessionDraft.files` contains full `MessageFile` objects. Image uploads populate both `content` with the complete base64 image and `thumbnail` with another base64 payload; prompt/file attachments may also contain their full text. Serializing that plus the TipTap document into synchronous localStorage can exceed the origin quota or stall the renderer. This catch then silently drops the entire write, including the user's text draft. Persist a bounded representation (for example text/document plus validated file descriptors) or use application storage designed for large records, and add an attachment/quota regression test.

src/dc-ui/components/button/DcCopyButton.vue
```diff
@@ -0,0 +1,79 @@
...
+          ),
+          label: accessibleName,
+          class: cn(
+            'shrink-0',
+            copied.value ? 'text-emerald-600 dark:text-emerald-400' : '',
+            inheritedAttrs.class
+          ),
+          onClick: copyText
+        },
+        slots
```
zerob13, line 93: [P2] Preserve the visible label fallback

The previous template rendered `<slot>{{ label }}</slot>`. Passing the raw `slots` object here removes that fallback. `CodeArtifact`, `McpJsonViewer`, `McpServers`, and `TraceDialog` pass `label` without a default slot, so their copy controls change from icon + visible text to icon-only. Preserve `label` as the default-slot fallback (or retain the template implementation) and cover one text-button consumer.

src/dc-ui/components/button/DcCopyButton.vue
```diff
@@ -0,0 +1,79 @@
...
+      const inheritedAttrs = attrs as Partial<DcButtonProps> & Record<string, unknown>
+      const { copyText: _copyText, ...buttonProps } = props
+      const accessibleName = inheritedAttrs.label ?? inheritedAttrs.tooltip ?? _copyText
+
+      return h(
+        DcButton,
+        {
+          ...inheritedAttrs,
+          ...buttonProps,
+          key: icon.value,
```
zerob13, line 76: [P2] Keep keyboard focus while changing the success icon

Using `icon.value` as the component key makes Vue unmount and recreate the entire `DcButton` whenever `copied` changes. A keyboard-activated copy therefore replaces the focused DOM node and moves focus to `<body>`; the next keyboard action and tooltip focus state are lost. The button and its icon already react to the `icon` prop, so this component-level key is unnecessary. Remove it and add a focus-retention assertion.

src/renderer/src/components/message/MessageToolbar.vue
```diff
@@ -64,35 +60,24 @@
...
               size="icon-sm"
               icon="lucide:chevron-right"
-              icon-size="3"
               :tooltip="t('thread.toolbar.nextVariant')"
               :tooltip-delay-duration="200"
               class="text-muted-foreground hover:text-primary hover:bg-transparent transition-colors duration-[var(--dc-motion-fast)] ease-[var(--dc-ease-out-soft)]"
               @click="emit('next')"
             />
-            <DcButton
+            <DcCopyButton
```
zerob13, line 68: [P2] Preserve the toolbar's 12 px icon contract

This hunk removes `icon-size="3"` from nearly every toolbar action, including the new copy control. `DcButton` defaults to `iconSize: '4'`, while the scoped toolbar CSS still constrains the button itself to 16 px. The result is a 16 px icon filling a 16 px control instead of the documented 12 px icon inside the 16 px exception; the edit action also remains at 12 px, making the row inconsistent. Restore `icon-size="3"` for all toolbar actions and keep the existing density test/contract.

yyhhyyyyyy (commented):
I reviewed the latest head (`557b875`). The latest commit fixes several earlier issues, including the required `copyText` contract, inherited click handlers/modifiers, the `MessageToolbar` copy event, storage getter failures, nested payload validation, and test teardown.

However, I still do not think this PR is safe to merge as-is. I found no P0 issues, but there are three P1 blockers and several P2 regressions.

P1 blockers:

1. Persisted drafts are not restored on the real keyed `ChatPage` mount path.

`ChatPage.vue` runs its immediate session watcher before `useComposerSubmit()` is constructed, so the watcher calls the initial no-op `switchComposerSessionDraft`. The real `switchComposerSession` function is assigned only afterward. Because `ChatTabView.vue` keys the route shell by session ID, every session switch creates a fresh `ChatPage`; the watcher does not run again on that instance.

As a result, a stored draft is not restored when reopening a session. A later `dispose()` can then flush the empty live composer and remove the previously stored draft. The current tests manually call `switchComposerSession()` and therefore do not cover the actual mount path. Please seed storage and test a real `ChatPage` mount without manually invoking the composable method.

2. Full attachment payloads are synchronously serialized into `localStorage`.

`ComposerSessionDraft.files` contains complete `MessageFile` objects. Image attachments contain the full base64 source in `content` and another base64 payload in `thumbnail`; file attachments may also contain their full text. Every draft save performs `JSON.stringify(draft)` and synchronous `localStorage.setItem()` on the renderer thread.

A normal screenshot or a few attachments can exceed the origin quota, and every later edit will retry the same large synchronous serialization. The exception is swallowed, so the text draft can silently fail to persist while the renderer may repeatedly stall. Please persist a bounded representation, or move full attachment persistence behind an asynchronous preload/IPC-backed storage service. This also needs oversized-attachment/quota coverage.

3. Session deletion does not remove persisted draft data.

`clearComposerDraftFromStorage()` currently has no production caller. Deleting a session removes the database session and messages, but leaves `deepchat.composerDraft.v1.<sessionId>` behind indefinitely, including pasted image/file contents. Cleanup needs to be part of the central session-removal lifecycle and must also prevent the active `ChatPage` unmount from writing the deleted draft back afterward.

Additional P2 regressions:

- `disposeComposerSubmit()` runs from the parent `onUnmounted` hook, after the child composer ref has been cleared. A rapid keyed unmount before the debounce fires can therefore persist raw text while losing the latest TipTap document and active skills. Flush before child teardown or retain the latest complete snapshot in the composable.
- The old `<slot>{{ label }}</slot>` fallback was removed. `CodeArtifact`, `McpJsonViewer`, `McpServers`, `TraceDialog`, and tool-call controls pass `label` without a default slot, so text buttons become icon-only. When no label/tooltip is provided, using the full `copyText` as `aria-label` can also make a screen reader announce the complete artifact source.
- `key: icon.value` is applied to the entire `DcButton`. The copy/check transition therefore replaces the focused button and drops keyboard focus. The inner icon key is sufficient; the button itself should remain stable.
- Copying now has two owners. `DcCopyButton` first uses the web Clipboard API, then `@copied` causes the message item to copy again through the Electron preload bridge. On unsupported web Clipboard APIs, VueUse may resolve without copying and show a false success state; on permission rejection, `copied` is not emitted, so the reliable preload path is skipped. Please use one clipboard adapter exactly once and enter the success state only after that adapter succeeds.
- Most `icon-size="3"` props were removed from `MessageToolbar`, so icons grow from 12 px to the `DcButton` default of 16 px inside 16 px controls. The edit action remains 12 px, making the toolbar inconsistent. Please restore the existing density contract.

The copy-button fix is valid and necessary, but the composer-persistence feature is substantial and unrelated. I recommend splitting this into a focused copy-button PR and a separate composer-persistence PR with the recovery, storage-boundary, deletion, quota, and keyed-remount cases covered before merge.

PR #2132, opened 2026-08-10
fix(chat): render resolved permission outcomes instead of generic Continued

### LINKED ISSUE #2117: [bug] Render resolved permission outcomes instead of generic Continued
## Summary

Resolved tool-permission action blocks are rendered with the generic label **Continued** regardless of whether the request was granted or denied. The resolution should be shown as part of the associated tool card using the actual persisted outcome.

**Priority:** P1

## Current behavior

The runtime correctly records permission resolution:

- granted request: `block.status = 'granted'` and `grantedPermissions` is populated;
- denied request: `block.status = 'denied'` and the request is no longer actionable.

See `markPermissionResolved()` in
`src/main/agent/deepchat/runtime/interactionProjection.ts` and the equivalent streaming helper in `src/main/agent/deepchat/runtime/process.ts`.

`src/renderer/src/components/message/MessageBlockAction.vue` does not inspect that outcome. Apart from rate-limit and legacy `needContinue` branches, every resolved action renders the same checkmark and `components.messageBlockAction.continued` string.

As a result, a denial can be visually described as “Continued.” Dedicated granted/denied i18n strings already exist but are not used by this component.

- `docs/architecture/tool-system.md` defines one main-process `ToolPermissionBroker` as the authority for host-tool consent.
- `docs/architecture/session-management.md` requires renderer and other entry points to share one permission semantic rather than invent independent meanings.
- `docs/architecture/local-control-plane/spec.md` treats approval result and redacted presentation as a typed state-machine outcome.

The renderer should faithfully project the broker-owned decision.

## Impact

- Users cannot tell whether they approved or rejected a tool after the interaction closes.
- “Continued” describes control flow, not the security decision the user made.
- Denial history can be actively misleading.
- The separate action card adds visual noise and is disconnected from the tool whose execution it controls.

## Proposed experience

```text
Before
[check] Continued
[Run command ...]  Completed

After grant
[Run command ...]  Allowed -> Running -> Completed

After denial
[Run command ...]  Denied  [Retry]
```

- Merge the resolved permission status into the associated tool card where possible.
- Use subtle `Allowed` / `Denied` labels derived from `block.status`.
- Preserve an audit trace without rendering a second large action card.
- Keep unresolved permission controls and read-only history behavior unchanged.

## Acceptance criteria

- [ ] Granted requests render as Allowed/Granted, never Continued.
- [ ] Denied requests render as Denied and cannot appear successful.
- [ ] The status is associated with the correct tool call.
- [ ] Pending permissions remain actionable and accessible.
- [ ] Question actions and legacy continue actions retain their distinct semantics.
- [ ] Renderer tests cover pending, granted, denied, read-only, and missing-tool-card fallbacks.

## User benefit

Permission history becomes accurate, compact, and trustworthy. Users can understand exactly what they allowed without reconstructing it from later tool behavior.

zhangmo8:
Closes #2117

## Summary

Resolved tool-permission action blocks previously rendered the generic checkmark + "Continued" label regardless of the persisted outcome, so a denial could look successful. The renderer now faithfully projects the broker-owned decision (`block.status = 'granted' | 'denied'`) and merges it into the associated tool card.

- `displayMessage.ts`: new `getResolvedPermissionStatus()` and `buildResolvedPermissionStatusByToolCallId()` helpers; the map only includes permission blocks whose tool card exists in the same message.
- `MessageItemAssistant.vue`: resolved permission action blocks with a matching tool card are no longer rendered as a separate card; their outcome is passed to `MessageBlockToolCall` via the new `permissionStatus` prop. Missing-tool-card cases fall back to a compact standalone label.
- `MessageBlockToolCall.vue`: renders an `Allowed` (emerald) / `Denied` (red) badge on the tool pill, styled like the existing RTK badge.
- `MessageBlockActivityGroup.vue`: forwards the permission map so tool cards inside collapsed activity groups show the badge too.
- `MessageBlockAction.vue`: fallback rendering uses shield-check "Permission granted" / shield-x "Permission denied" (existing i18n keys) instead of "Continued". Rate-limit, legacy needContinue and other action types keep their behavior.
- i18n: added `toolCall.badge.allowed` / `toolCall.badge.denied` to all 20 locales.

## Layout

```text
BEFORE                          AFTER (grant)                  AFTER (deny)
[check] Continued               [run_command  Allowed >]       [run_command  Denied >]
[run_command  >]  Completed
```

## Acceptance criteria

- [x] Granted requests render as Allowed/Granted, never Continued
- [x] Denied requests render as Denied and cannot appear successful
- [x] The status is associated with the correct tool call (matched by `tool_call.id`)
- [x] Pending permissions remain actionable and accessible (overlay flow untouched)
- [x] Question actions and legacy continue actions retain their distinct semantics
- [x] Renderer tests cover pending, granted, denied, read-only, and missing-tool-card fallbacks

## Verification

- `vitest run test/renderer/components/message` — 16 files, 151 tests passed
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run format`, `pnpm run i18n` all pass

## Summary by CodeRabbit

- **New Features**
  - Tool-call cards and activity history now display clear “Allowed” or “Denied” permission badges with status-specific icons and colors.
  - Permission outcomes appear directly on standard and live tool-call cards.
  - Pending or unmatched permission requests remain visible without misleading resolution labels.

- **Localization**
  - Added translated permission-status labels across supported languages.

- **Tests**
  - Added coverage for granted, denied, pending, unmatched, read-only, live, and error-state permission displays.

### DIFF
--- a/src/renderer/src/components/message/MessageBlockToolCall.vue
+++ b/src/renderer/src/components/message/MessageBlockToolCall.vue
@@ -44,6 +44,21 @@
         </span>
       </div>
       <span
+        v-if="permissionStatus"
+        data-testid="tool-call-permission-badge"
+        :data-permission-status="permissionStatus"
+        :class="[
+          'shrink-0 rounded border px-1.5 py-0.5 text-[10px] font-medium',
+          permissionStatus === 'granted'
+            ? 'border-emerald-500/20 bg-emerald-500/10 text-emerald-700 dark:text-emerald-300'
+            : 'border-red-500/20 bg-red-500/10 text-red-700 dark:text-red-300'
+        ]"
+      >
+        {{
+          permissionStatus === 'granted' ? t('toolCall.badge.allowed') : t('toolCall.badge.denied')
+        }}
+      </span>
+      <span
         v-if="showRtkBadge"
         data-testid="tool-call-rtk-badge"
         class="shrink-0 rounded border border-emerald-500/20 bg-emerald-500/10 px-1.5 py-0.5 text-[10px] font-semibold uppercase tracking-[0.08em] text-emerald-700 dark:text-emerald-300"
@@ -263,6 +278,7 @@ const props = defineProps<{
   threadId?: string
   readOnly?: boolean
   renderMode?: 'full' | 'tool-only' | 'app-only'
+  permissionStatus?: 'granted' | 'denied'
 }>()
 
 type ExpansionSource = 'auto' | 'manual' | null
--- a/src/renderer/src/components/message/MessageBlockToolCall.vue
+++ b/src/renderer/src/components/message/MessageBlockToolCall.vue
@@ -8,6 +8,7 @@
       :details-id="detailsId"
       :details-expanded="isExpanded"
       :read-only="readOnly"
+      :permission-status="permissionStatus"
       @toggle-details="toggleExpanded"
     />
     <button
--- a/src/renderer/src/components/message/MessageItemAssistant.vue
+++ b/src/renderer/src/components/message/MessageItemAssistant.vue
@@ -48,6 +48,7 @@
                 :reasoning-count="item.reasoningCount"
                 :tool-call-count="item.toolCallCount"
                 :read-only="isReadOnly"
+                :permission-status-by-tool-call-id="permissionStatusByToolCallId"
                 @toggle-collapse="handleCollapseToggle"
               />
               <MessageBlockToolCall
@@ -88,6 +89,11 @@
                 :thread-id="currentThreadId"
                 :read-only="isReadOnly"
                 :render-mode="item.block.tool_call?.mcpResult?.app ? 'tool-only' : 'full'"
+                :permission-status="
+                  item.block.tool_call?.id
+                    ? permissionStatusByToolCallId[item.block.tool_call.id]
+                    : undefined
+                "
               />
               <MessageBlockQuestionRequest
                 v-else-if="
@@ -217,7 +223,9 @@ import { ref, computed, watch } from 'vue'
 import {
   type DisplayAssistantMessage,
   type DisplayAssistantMessageBlock,
+  buildResolvedPermissionStatusByToolCallId,
   filterRenderableAssistantBlocks,
+  getResolvedPermissionStatus,
   isInternalAssistantToolCallBlock
 } from '@/features/chat-page/model/displayMessage'
 import MessageBlockContent from './MessageBlockContent.vue'
@@ -437,9 +445,23 @@ const shouldGroupActivity = computed(() => {
   return currentMessage.value.status !== 'pending'
 })
 
+const permissionStatusByToolCallId = computed(() =>
+  buildResolvedPermissionStatusByToolCallId(currentContent.value)
+)
+
+// Resolved permission outcomes merge into their tool card; the standalone
+// action card only remains as a fallback when the tool card is missing.
+const currentVisibleContent = computed(() =>
+  currentContent.value.filter((block) => {
+    const status = getResolvedPermissionStatus(block)
+    const toolCallId = block.tool_call?.id
+    return !(status && toolCallId && permissionStatusByToolCallId.value[toolCallId])
+  })
+)
+
 const currentRenderItems = computed(() =>
   buildAssistantRenderItems({
-    blocks: currentContent.value,
+    blocks: currentVisibleContent.value,
     messageId: currentMessage.value.id,
     messageUpdatedAt: currentMessage.value.updatedAt,
     shouldGroup: shouldGroupActivity.value,
--- a/src/renderer/src/i18n/en-US/toolCall.json
+++ b/src/renderer/src/i18n/en-US/toolCall.json
@@ -13,7 +13,9 @@
   "imagePreviewCount": "No image previews | One image preview | {count} image previews",
   "terminalOutput": "Terminal Output",
   "badge": {
-    "rtk": "RTK"
+    "rtk": "RTK",
+    "allowed": "Allowed",
+    "denied": "Denied"
   },
   "replacementsCount": "Completed {count} replacements",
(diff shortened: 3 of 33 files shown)


src/renderer/src/components/message/MessageBlockToolCall.vue
```diff
@@ -43,6 +43,21 @@
           {{ summaryText }}
         </span>
       </div>
+      <span
```
zerob13, line 47: [P1] Preserve permission outcomes for live-delegation cards

This badge is inside the ordinary tool-call `button`, but trusted live-delegation spawn blocks take the preceding `LiveDelegationToolCallCard` branch. In explicit orchestration mode, `agentToolManager` requires `subagentStart` permission for these spawns; `MessageItemAssistant` then removes the resolved action as soon as a matching tool-call ID exists. The dedicated card never receives or renders `permissionStatus`, so grants show only the delegation state and denials collapse to a generic tool error—there is no Allowed/Denied audit outcome. Render the permission state in the live-delegation card (or keep its standalone action) and cover both decisions with a regression test.

src/renderer/src/i18n/en-US/toolCall.json
```diff
@@ -13,7 +13,9 @@
   "imagePreviewCount": "No image previews | One image preview | {count} image previews",
   "terminalOutput": "Terminal Output",
   "badge": {
-    "rtk": "RTK"
+    "rtk": "RTK",
+    "allowed": "Allowed",
```
zerob13, line 17: [P3] Regenerate the committed i18n type declaration

`src/types/i18n.d.ts` is generated from the zh-CN locale and committed, but this PR leaves `toolCall.badge` with only `rtk`. Run `pnpm i18n:types` and include the generated `allowed` / `denied` members so the source messages and `DefineLocaleMessage` contract stay in sync.

zerob13 (commented):
Reviewed the current head. Required GitHub checks are green, but I found one user-visible permission-history gap and one generated i18n contract drift; details are inline.

src/renderer/src/components/message/MessageBlockToolCall.vue
zhangmo8: Fixed in 9969d7701. `LiveDelegationToolCallCard` now accepts `permissionStatus` and renders the Allowed/Denied badge next to the delegation status badge, and `MessageBlockToolCall` forwards the prop to the card branch. Both decisions are covered by regression tests: granted (spawn with a delegation response) and denied (errored spawn without a delegation, which still routes to the card).

src/renderer/src/i18n/en-US/toolCall.json
zhangmo8: Regenerated in 11809d2c9 via `pnpm i18n:types` — `toolCall.badge` now includes `allowed`/`denied`. The run also synced a few other keys the committed declaration had drifted on (commandShell, outputLimits, etc.).

src/renderer/src/components/message/MessageItemAssistant.vue
```diff
@@ -88,6 +89,11 @@
                 :thread-id="currentThreadId"
                 :read-only="isReadOnly"
                 :render-mode="item.block.tool_call?.mcpResult?.app ? 'tool-only' : 'full'"
+                :permission-status="
+                  item.block.tool_call?.id
+                    ? permissionStatusByToolCallId[item.block.tool_call.id]
+                    : undefined
+                "
```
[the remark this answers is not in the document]
zhangmo8: Verified against the current code — this scenario cannot occur, so no change is needed. A tool call with `mcpResult.app` always renders its tool pill alongside the `mcp-app` item: ungrouped, it renders through the normal tool-call branch with `render-mode="tool-only"` which already receives `permission-status` (MessageItemAssistant.vue L85-97); grouped, it renders inside `MessageBlockActivityGroup`, which also forwards the status (L54-55). The badge therefore always has a home on the pill. The proposed fix would also be a no-op: in `app-only` mode `MessageBlockToolCall` renders only `McpAppView` — the badge lives in the pill button, which is `v-else-if="renderMode !== 'app-only'"`.

test/renderer/components/message/MessageItemAssistant.test.ts
```diff
@@ -742,4 +742,87 @@ describe('MessageItemAssistant', () => {
...
+      const wrapper = mountWith([
+        createPermissionActionBlock('denied'),
+        createToolCallBlock({ tool_call: { id: 'tc1', name: 'run_command' } })
+      ])
+
+      expect(wrapper.findComponent({ name: 'MessageBlockAction' }).exists()).toBe(false)
+      expect(
+        wrapper.find('[data-testid="tool-call-stub"]').attributes('data-permission-status')
+      ).toBe('denied')
+    })
```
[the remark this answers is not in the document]
zhangmo8: Added in 9969d7701 — `merges the outcome only into the tool call with the matching id` mounts two tool calls with distinct ids plus one resolved permission action and asserts only the matching stub receives `permissionStatus` while the other stays unset.

zhangmo8:
Heads-up on the extra `ci:` commit (88a169154): `test-renderer` OOM'd twice in a row on the merge ref (a straight rerun reproduced it — [run](https://github.com/ThinkInAIXYZ/deepchat/actions/runs/31455699016)). The GC log shows a single reused jsdom worker breaching the 4GB Node default heap cap (`Mark-Compact 4081.2 (4106.2)` → fatal). The branch itself is not the cause — the full suite passes locally on this branch alone — but dev merged several PRs with new renderer suites after this PR's last green run (#2129, #2131, #2133, #2134, #2138), and the merge-ref accumulation now crosses the 4GB line, which will start blocking other PRs' merge CIs too. Raised the job's heap ceiling to 6GB (standard 16GB runner, `maxWorkers: 2`). Happy to split this into its own PR if you prefer.

PR #2135, opened 2026-08-10
fix(cli): make model outputs JSON-safe

### LINKED ISSUE #2120: [bug] Make model list and config-get outputs JSON-safe
## Summary

The current `dev` build returns `internal_error: Route returned an invalid result` for `model list` and `model config-get`, including across unrelated providers. The route result passes its domain schema but fails the CLI server's pure-JSON validation.

**Priority:** P1 for CLI administration and diagnostics

## Reproduction

The following public commands fail after successful provider/model resolution:

```text
deepchat model list --provider <provider> --json
deepchat model config-get --provider <provider> --model <model> --json
```

The behavior reproduces with more than one built-in provider, so it is not provider-specific.

## Root cause

The failure occurs in the boundary between route schemas and the CLI JSON envelope:

- `modelsListRuntimeRoute` in `src/shared/contracts/routes/models.routes.ts` contains model objects with many optional fields.
- `src/main/provider/routes.ts` returns `providerRuntime.getModelList()` through that schema.
- `PublicModelConfigSchema` also contains optional configuration fields.
- `CliServer.parseRouteOutput()` in `src/main/cli/server.ts` first parses the route output and then parses the result with `JsonValueSchema`.

Zod preserves an optional key when the input object explicitly contains that key with value `undefined`. Such an object passes the model/config schema but is not a valid JSON value, causing the second parse to fail.

This explanation matches both failing routes and was independently reproduced with the same Zod optional-field behavior. The exact offending key should still be captured in a safe regression test rather than logged with values.

- `docs/architecture/local-control-plane/spec.md` requires canonical typed routes, deterministic JSON/JSONL, and stable machine-mode envelopes.
- It explicitly requires safe/redacted public DTOs and rejects a second loosely defined CLI-only schema.
- Public model configuration is intended to reject or strip unsupported fields while remaining JSON serializable.

## Impact

- CLI users cannot enumerate runtime models or inspect effective model configuration.
- Provider diagnosis and benchmark setup are blocked.
- The generic error does not identify a schema path, making regressions hard to isolate.
- Renderer IPC can appear healthy while the public CLI contract is broken.

## Proposed direction

- Introduce an explicit JSON-safe DTO projection that omits every `undefined` property before public route output.
- Prefer field-by-field public mapping over generic serialization that could accidentally expose extra provider metadata.
- Add secret-safe development logging of failing schema paths/types, never values.
- Audit other CLI routes that reuse optional-rich renderer/domain schemas.

## Acceptance criteria

- [ ] `model list` succeeds when optional model fields are explicitly undefined.
- [ ] `model config-get` succeeds with sparse effective configuration.
- [ ] Returned output passes both the canonical route schema and `JsonValueSchema`.
- [ ] No provider secrets, raw auth state, or unknown internal fields are added to the DTO.
- [ ] Tests cover undefined optional fields, empty model lists, custom models, and multiple providers.
- [ ] A generic route-boundary test detects non-JSON values before release.

## User benefit

Model administration and diagnostics work reliably from scripts, with a stable and safely redacted JSON contract.

yyhhyyyyyy:
## Summary

- Project runtime model entries into an explicit JSON-safe DTO before returning `models.listRuntime`.
- Project public model configuration field by field for both `models.getPublicConfig` and `models.setPublicConfig`.
- Omit explicitly `undefined` properties, including nested image, video, and TTS options.
- Keep the CLI route boundary fail-closed for schema-valid outputs that are not valid JSON.
- Log bounded validation metadata without exposing route values or dynamic record keys.

## Root Cause

Zod preserves optional properties when the input explicitly contains those properties with an `undefined` value.

The affected model routes passed their domain schemas, but their parsed results then failed `JsonValueSchema` validation in the CLI server because `undefined` is not a valid JSON value. As a result, `model list` and `model config-get` returned `internal_error`.

## Compatibility

- Preserves all existing renderer-visible runtime model fields, including falsy values.
- Does not normalize, clamp, or add defaults to model configuration values.
- Continues stripping private and unknown configuration fields.
- Does not change the public CLI protocol or command syntax.

## Testing

- Covered empty model lists, multiple providers, and custom models.
- Covered every currently exposed optional runtime model field.
- Covered sparse public configuration for both config-get and config-set.
- Covered nested image, video, and TTS properties containing explicit `undefined`.
- Covered unknown/private field redaction.
- Added a generic CLI boundary test for schema-valid but non-JSON output.
- Verified validation logging does not expose dynamic output keys.

Validation completed:

- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`
- 40 focused Vitest tests

Fixes #2120

## Summary by CodeRabbit

* **Bug Fixes**
  * Model configuration responses now expose only approved, JSON-safe settings, excluding private or unknown fields for both retrieval and updates.
  * Runtime model listings no longer expose internal credentials or runtime-only fields.
  * Invalid route responses are rejected safely with improved diagnostics, without leaking sensitive values or being treated as successful responses.

* **Tests**
  * Expanded coverage for configuration redaction, runtime model responses, JSON safety, and invalid route output handling.

### DIFF
--- a/src/main/cli/providerModelAdminRoutes.ts
+++ b/src/main/cli/providerModelAdminRoutes.ts
@@ -11,7 +11,7 @@ import {
   type PublicProviderSummary,
   type SettingsActivityInput
 } from '@shared/contracts/routes'
-import type { LLM_PROVIDER } from '@shared/types/provider'
+import type { LLM_PROVIDER, ModelConfig } from '@shared/types/provider'
 import type { ProviderRuntime } from '@/provider'
 import type { ProviderQueryScheduler } from '@/provider/providerService'
 import type { ProviderSettingsPort } from '@/provider/settings'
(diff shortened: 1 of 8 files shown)


src/main/cli/providerModelAdminRoutes.ts
```diff
@@ -74,6 +74,129 @@ export function toPublicProviderSummary(provider: LLM_PROVIDER): PublicProviderS
   })
 }
 
+function toPublicModelConfig(config: ModelConfig) {
```
zerob13: [P2] Keep the Zod schema as the only public field list

This mapper duplicates `PublicModelConfigSchema` down to every nested optional field. It is correct today, but the next public schema field will be silently omitted until this function and its tests are updated, effectively creating the second CLI-only schema that the control-plane contract is meant to avoid. Parse with `PublicModelConfigSchema` first so unknown/private fields are stripped, then apply JSON serialization semantics to that parsed DTO to omit explicit `undefined` values, and validate with the route output schema. That preserves the redaction boundary while deleting this hand-maintained allowlist.

src/main/provider/routes.ts
```diff
@@ -74,6 +75,33 @@ import { ProviderService, type ProviderQueryScheduler } from './providerService'
 import type { ProviderRuntime } from '.'
 import { CliRequestError } from '@/cli/errors'
 
+function toRuntimeModelDto(model: MODEL_META): MODEL_META {
```
zerob13: [P2] Do not mirror `ProviderModelSummarySchema` here

`modelsListRuntimeRoute.output` already owns the public field allowlist and strips unknown provider metadata. Normalize the schema-parsed result to JSON to remove explicit `undefined` properties instead of copying every `MODEL_META` member here. This helper otherwise silently drops any future public model field and requires a second field-by-field test matrix for the same contract.

zerob13 (commented):
The current behavior is covered and CI is green, but both DTO helpers duplicate the canonical route schemas and create avoidable field-drift risk. Details are inline.

src/main/cli/providerModelAdminRoutes.ts
yyhhyyyyyy: Fixed in eeb0ed0e5: public model configs now use the canonical route output schema for projection, JSON normalization, and final validation, removing the manual field mapper.

src/main/provider/routes.ts
yyhhyyyyyy: Fixed in eeb0ed0e5: runtime models now use the canonical route output schema before JSON normalization and revalidation, removing the duplicated field allowlist.

PR #2136, opened 2026-08-10
fix(cli): wait for detached run terminal state

### LINKED ISSUE #2113: [bug] Do not terminate run watch on provider stream completion
## Summary

`deepchat run watch` can exit when one provider stream finishes even though the detached Agent run is still active or waiting for a user interaction.

**Priority:** P1

## Reproduction

Using a sanitized detached run:

1. Start an Agent run that reaches a tool requiring approval.
2. Start `deepchat run watch --run <run-id>`.
3. The provider round emits `chat.stream.completed` before the permission interaction is resolved.
4. `run watch` exits successfully.
5. An immediate `run get` still reports an active Session with a pending assistant message.

The permission wait is valid behavior. The defect is that the watcher reports terminal completion too early.

## Root cause

`RunService.isTerminalEvent()` in `src/main/cli/runService.ts` treats either of these root-Session events as terminal:

- `chat.stream.completed`
- `chat.stream.failed`

Those events describe one stream/provider-round boundary, not necessarily the entire durable run lifecycle. A tool loop, follow-up provider round, permission interaction, question interaction, pending-input handoff, or resume can still follow.

The existing tests in `test/main/cli/runService.test.ts` explicitly expect a root `chat.stream.completed` event to resolve the watcher, so the incorrect lifecycle assumption is currently encoded as a contract.

- `docs/architecture/local-control-plane/spec.md` requires `--jsonl` to emit one terminal result/error record and says thin CLI invocations must exit after that terminal result or EOF.
- The same document defines detached runs as recoverable from Session state and event cursors after disconnection.
- `docs/architecture/durable-execution-journal/spec.md` makes `execution/run_terminal` the authoritative physical-run terminal fact before terminal projections.

The watcher should terminate on the durable run/session contract, not a provider-stream implementation event.

## Impact

- Automation can start downstream work while the Agent is still running.
- Benchmarks record incomplete latency and output.
- Scripts may omit required interaction handling or cancellation.
- Users interpret a normal interaction pause as an unexplained stop.

## Proposed direction

- Stop using `chat.stream.completed/failed` as whole-run terminal signals.
- Prefer an explicit run-terminal event or a final Session state transition to `idle`/`error` with matching run ownership.
- Add a public phase such as `running | awaiting_interaction | terminal`; do not represent an interaction wait as ordinary active generation only.
- Preserve cursor catch-up, descendant-run filtering, overflow handling, and already-terminal fast paths.

## Acceptance criteria

- [ ] A provider stream completion does not terminate `run watch` while another round or interaction remains possible.
- [ ] `awaiting_interaction` is observable without treating the permission/question wait as an error.
- [ ] The watcher emits exactly one terminal result for the owned run.
- [ ] Descendant Session events cannot terminate the root watcher.
- [ ] Already-terminal snapshots still return immediately.
- [ ] Tests cover provider-round completion, permission pause/resume, question pause/resume, cancellation, failure, and final completion.

## User benefit

CLI users and harnesses can trust that “watch finished” means the run actually reached a terminal state.

yyhhyyyyyy:
## Summary

Fix `deepchat run watch` exiting after a single provider round while the detached Agent run is still active or waiting for user interaction.

- Treat root Session transitions to `idle` or `error` as the authoritative run terminal state.
- Keep watching after `chat.stream.completed` and `chat.stream.failed`.
- Preserve initial-turn failure termination, cursor recovery, overflow handling, and already-terminal fast paths.
- Prevent descendant Session terminal events from terminating the root watcher.
- Expose `running | awaiting_interaction | terminal` through `runs.get` and run snapshots.
- Detect direct permission/question waits, current live-delegation waits, and compatible legacy subagent waits.
- Derive run phase independently of transcript pagination and cursors.

## Root Cause

`chat.stream.completed` and `chat.stream.failed` describe a single provider-stream boundary, not the lifecycle of the complete detached run.

A tool loop, permission or question interaction, subagent execution, pending-input handoff, or another provider round may still follow. Treating these events as terminal caused `run watch` to return success before the owned root Session actually finished.

## Compatibility

`phase` is a required field in the strict public run snapshot schema. The local-control surface version is therefore bumped from `1` to `2` so incompatible clients and servers fail closed.

The local-control protocol remains version `1`, and the existing `/v1/...` transport paths are unchanged.

## Test Coverage

Added regression coverage for:

- provider completion without run termination
- provider failure without premature termination
- permission and question pause/resume
- current live-delegation interaction waits
- legacy interaction recovery
- malformed interaction progress
- cursor replay and catch-up ordering
- limited transcript pages
- already-terminal snapshots
- initial-turn startup failure
- descendant Session terminal filtering
- cancellation behavior
- surface-version compatibility

## Validation

- 29 test files passed
- 357 tests passed
- Node and renderer typechecks passed
- Lint passed with 0 warnings and 0 errors
- Format check passed
- i18n validation passed

Fixes #2113

## Summary by CodeRabbit

* **New Features**
  * Added explicit run phases: running, awaiting interaction, and terminal.
  * Run status now identifies when user input or approval is required.
  * Waiting interactions from delegated tasks are reflected in run status.

* **Improvements**
  * `run watch` continues through provider stream completion and ends when the root run becomes idle or errors.
  * Improved recovery and monitoring for detached runs, including incomplete or malformed progress data.
  * Updated the CLI and local-control protocol to version 2.

### DIFF
--- a/src/main/session/data/tables/deepchatMessages.ts
+++ b/src/main/session/data/tables/deepchatMessages.ts
@@ -32,20 +32,9 @@ export interface DeepChatAssistantMessageIdentityRow {
   updated_at: number
 }
 
-const PENDING_ASSISTANT_INDEX_SQL = `
-  CREATE INDEX IF NOT EXISTS idx_deepchat_messages_pending_assistant
-    ON deepchat_messages(session_id, order_seq, id)
-    WHERE role = 'assistant' AND status = 'pending';
-`
-
 export class DeepChatMessagesTable extends BaseTable {
   constructor(db: Database.Database) {
     super(db, 'deepchat_messages')
-  }
-
-  override createTable(): void {
-    super.createTable()
-    this.db.exec(PENDING_ASSISTANT_INDEX_SQL)
   }
 
   getCreateTableSQL(): string {
@@ -63,7 +52,6 @@ export class DeepChatMessagesTable extends BaseTable {
         updated_at INTEGER NOT NULL
       );
       CREATE INDEX IF NOT EXISTS idx_deepchat_messages_session ON deepchat_messages(session_id, order_seq);
-      ${PENDING_ASSISTANT_INDEX_SQL}
     `
   }
 
--- a/src/main/session/data/tables/deepchatMessages.ts
+++ b/src/main/session/data/tables/deepchatMessages.ts
@@ -32,9 +32,20 @@ export interface DeepChatAssistantMessageIdentityRow {
   updated_at: number
 }
 
+const PENDING_ASSISTANT_INDEX_SQL = `
+  CREATE INDEX IF NOT EXISTS idx_deepchat_messages_pending_assistant
+    ON deepchat_messages(session_id, order_seq, id)
+    WHERE role = 'assistant' AND status = 'pending';
+`
+
 export class DeepChatMessagesTable extends BaseTable {
   constructor(db: Database.Database) {
     super(db, 'deepchat_messages')
+  }
+
+  override createTable(): void {
+    super.createTable()
+    this.db.exec(PENDING_ASSISTANT_INDEX_SQL)
   }
 
   getCreateTableSQL(): string {
@@ -52,6 +63,7 @@ export class DeepChatMessagesTable extends BaseTable {
         updated_at INTEGER NOT NULL
       );
       CREATE INDEX IF NOT EXISTS idx_deepchat_messages_session ON deepchat_messages(session_id, order_seq);
+      ${PENDING_ASSISTANT_INDEX_SQL}
     `
   }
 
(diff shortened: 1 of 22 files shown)


src/main/session/data/tables/deepchatMessages.ts
```diff
@@ -32,11 +32,22 @@ export interface DeepChatAssistantMessageIdentityRow {
   updated_at: number
 }
 
+const PENDING_ASSISTANT_INDEX_SQL = `
```
zerob13: [P2] Avoid building a global index for this bounded CLI lookup

`createTable()` runs during application startup, and this synchronous `better-sqlite3` statement will scan the entire existing `deepchat_messages` table on the first launch after upgrade. The new query is only used for a generating run when the requested transcript page is paginated, while `idx_deepchat_messages_session(session_id, order_seq)` already bounds that lookup to one Session. Keep `getPendingAssistantBySession()`, but remove this partial index and the `createTable()` override unless profiling shows the per-Session scan is material; the current optimization trades a rare bounded lookup for an unbounded main-thread startup cost.

zerob13 (commented):
The lifecycle fix and phase contract are coherent and CI is green. One added optimization expands startup risk without evidence that the existing session index is insufficient; details are inline.

src/main/session/data/tables/deepchatMessages.ts
yyhhyyyyyy: Removed the startup partial index and `createTable()` override. The lookup now relies on the existing per-Session index.

PR #2137, opened 2026-08-10
fix(chat): retry released queue inputs

### LINKED ISSUE #2112: [bug] Expose explicit retry for released queued inputs
## Summary

When a queued input is claimed and then released before its user fact is committed, the runtime intentionally leaves it at the queue head for an explicit retry. The renderer does not expose that retry action for an ordinary pending row.

**Priority:** P1

## Current behavior

`PendingInputPump.scheduleNextIfReady()` in
`src/main/agent/deepchat/runtime/pendingInputPump.ts` checks whether the just-released row is still pending. If it is, the pump deliberately does not schedule the next drain because the row is waiting for an explicit retry.

The regression test in
`test/main/agent/deepchat/runtime/pendingInputPump.test.ts` codifies this behavior as:

> leaves a released input at the head for explicit retry

However, `src/renderer/src/components/chat/PendingInputLane.vue` only renders **Retry** and **Send without image content** for attachment-blocked rows. A normal pending row exposes edit, steer, reorder, and delete, but no retry control and no “retry required” status.

Editing the row may indirectly cause another schedule, but that is an undocumented side effect rather than the explicit retry contract required by the pump.

- `docs/architecture/session-management.md` defines Queue as a mutable draft before claim and states that pending-input persistence is authoritative.
- `docs/features/im-style-steer-messages/spec.md` keeps Queue in the pending lane and distinguishes Queue release/retry semantics from the irreversible Steer claim boundary.
- `docs/architecture/deepchat-agent-harness-boundaries/spec.md` assigns release and subsequent drain ownership to `PendingInputPump`.

The backend and tests agree that explicit retry is required; the UI contract is incomplete.

## Impact

- A recoverable row can stay at the queue head indefinitely.
- Later rows cannot drain because FIFO ordering preserves the released head.
- Users receive no explanation that the row needs action.
- Editing text appears to “magically” fix the problem, creating an unreliable mental model.

## Proposed direction

- Project a clear retry-required state for a released Queue row without inventing a second source of truth.
- Add an explicit **Retry** action for that state.
- Route retry through the existing pending-input coordinator so capacity, attachment, interaction, and single-flight gates are re-evaluated.
- Keep ordinary queued drafts unchanged when they have not been released from a failed claim.

## Acceptance criteria

- [ ] A released Queue head is visibly labeled as requiring retry.
- [ ] The row exposes an explicit Retry action without requiring an edit.
- [ ] Retry schedules only that Session and remains idempotent under repeated clicks.
- [ ] A later Queue row is not admitted ahead of the released FIFO head.
- [ ] Attachment-blocked actions retain their existing behavior.
- [ ] Renderer and main-process tests cover release, visible retry, successful re-admission, and repeated retry.

## User benefit

Recoverable queued messages become understandable and actionable instead of silently blocking the queue.

yyhhyyyyyy:
## Summary

- Add a durable `retry_required` state for Queue inputs released after a pre-user-fact failure or transcript rollback.
- Expose an explicit per-item Retry action in the pending input lane.
- Preserve FIFO ordering, restart-held Queue behavior, ACP compatibility, and edit-as-retry behavior.
- Add regression coverage across persistence, runtime admission, routes, restart recovery, ACP, and renderer behavior.

Closes #2112.

## Root Cause

PR #2023 centralized pending input claim, release, and single-flight draining in `PendingInputPump`.

When a claimed Queue input failed before its user fact was committed, or after the fact was rolled back, it was released to the FIFO head as an ordinary `pending` row. The pump intentionally did not replay that row automatically, but the renderer could not distinguish it from a normal draft and exposed no explicit recovery action.

PR #2129 added restart recovery and Resume Queue behavior, which made the issue more visible, but did not introduce the original defect.

## Changes

- Add `retry_required` to `PendingSessionInputState`.
- Transition failed DeepChat Queue settlements from `claimed` to `retry_required`.
- Keep intentional Queue-to-Steer mutation and restart repair releases as ordinary `pending`.
- Keep retry-required rows in FIFO ordering and capacity accounting while excluding them from automatic dispatch.
- Add a typed, session-gated `sessions.retryPendingQueueInput` route.
- Validate session ownership, retry state, and FIFO-head position inside the persistence transaction.
- Return separate `accepted` and `started` results so an accepted retry delayed by a runtime gate is not reported as a failure.
- Preserve retry acceptance when event publication or the first immediate drain fails after persistence.
- Keep ACP pending-input semantics unchanged.
- Preserve `attachmentFallbackPolicy` when editing a Queue item.
- Add localized Retry labels and descriptions for all supported locales.

No database migration is required because the pending input state column is unconstrained text.

### Before

```text
┌──────────────────────── Queue ────────────────────────┐
│ Failed message                              [Steer][×] │
└───────────────────────────────────────────────────────┘
```

The failed item remained at the Queue head without explaining why the Queue was blocked or how to continue it.

### After

```text
┌──────────────────────── Queue ────────────────────────┐
│ Failed message   This message wasn't sent             │
│                  [Retry required]          [Retry][×]  │
└───────────────────────────────────────────────────────┘
```

Retry-required rows cannot be reordered or converted to Steer. They can be retried, edited and retried, or removed.

## Safety and Compatibility

- Later Queue items cannot bypass a retry-required FIFO head, including through the question-follow-up fast path.
- Duplicate Retry requests produce at most one state transition and turn.
- Restart-held Queue rows retain the existing Resume Queue contract.
- A retry-required head prevents a restart-held tail from being resumed out of order.
- ACP continues releasing failed Queue claims to ordinary `pending`.
- Retry diagnostics do not include message content, attachment paths, or payload data.
- Publication and immediate-drain failures after a successful state transition remain recoverable.

## Validation

- [x] `pnpm run format`
- [x] `pnpm run i18n`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck:node`
- [x] `pnpm run typecheck:web`
- [x] Main pending-input, route, session, backend, and ACP tests — 210 passed
- [x] Renderer pending-input tests — 29 passed
- [x] DeepChat harness failure and rollback boundaries — 4 passed

Twelve native SQLite tests were skipped by the repository harness because the locally installed native module ABI does not match the Node ABI used by Vitest. Equivalent store, FIFO, retry, and restart contracts passed in the non-native suites.

## Summary by CodeRabbit

- **New Features**
  - Added explicit retry handling for queued messages that cannot proceed automatically.
  - Retry-required items remain at the queue head until manually retried.
  - Added Retry actions with loading, failure, and recovery feedback.
  - Editing a retry-required message now retries it safely.
  - Added localized retry messaging across supported languages.

- **Bug Fixes**
  - Improved restart and queue-resume behavior for interrupted messages.
  - Prevented duplicate retries and preserved queued message settings during edits.
  - Prevented later queued messages from bypassing items requiring retry.


src/main/session/data/pendingInputStore.ts
```diff
@@ -461,7 +474,7 @@ export class SessionPendingInputStore {
 
   private getWaitingQueueRows(sessionId: string): DeepChatPendingInputRow[] {
     return this.getQueueRows(sessionId).filter(
-      (row) => row.state === 'pending' || row.state === 'blocked'
+      (row) => row.state === 'pending' || row.state === 'blocked' || row.state === 'retry_required'
```
zerob13: [P1] Preserve the claimed Queue slot during every resequence. `getWaitingQueueRows()` excludes the active `claimed` row, while move/delete/conversion renumber the visible tails from 1. Reproduced on this head: reorder a newer item to the head, let the pump claim it, reorder an older tail, then release the claimed item to `retry_required`; both rows get `queue_order = 1`, the older tail wins `created_at`, and `getNextPendingQueueInput()` returns that tail instead of `null`. A later wake can therefore send the wrong item before the required retry. Include active claimed rows when resequencing (or preserve their occupied slots) and cover move/delete/conversion during a claim followed by retry release.

src/main/session/data/pendingInputStore.ts
```diff
@@ -334,19 +337,29 @@ export class SessionPendingInputStore {
...
     const row = this.requireRow(itemId)
     if (row.mode !== 'queue') {
       throw new Error(`Pending input ${itemId} is not a queue item.`)
     }
-    if (row.state !== 'claimed') {
-      return this.toRecord(row)
+    if (row.state !== 'retry_required') {
+      throw new Error(`Pending queue item ${itemId} does not require retry.`)
+    }
+    if (this.getWaitingQueueRows(row.session_id)[0]?.id !== itemId) {
```
zerob13: [P2] Do not require a retry-required row to already be the FIFO head. `shouldClaimImmediately()` intentionally lets an explicit Send bypass restart-held Queue drafts (and the question-follow-up fast path can bypass ordinary drafts), so that claimed row receives a later `queue_order`. If it fails before the user fact, this PR turns it into a non-head `retry_required` row; the renderer shows Retry, but this check makes every click silently return `accepted: false` until unrelated earlier rows are cleared. Transitioning that row to `pending` is safe because the existing dispatcher still enforces FIFO. Let the transactional operation return accepted/no-op without this head restriction, delete the duplicated coordinator-side sort/preflight, and add a held-head + failing fast-path Send + Retry regression.

zerob13 (commented):
Two actionable state-machine issues remain. The first is reproduced by a failing regression on the current head; the second combines two already-supported paths that are only tested separately.

docs/issues/pending-input-explicit-retry/spec.md
```diff
@@ -0,0 +1,93 @@
...
+that a released Queue head waits for explicit retry, but release persisted the same `pending` state
+used by ordinary drafts and the renderer contract was not extended.
+
+PR https://github.com/ThinkInAIXYZ/deepchat/pull/2129 did not introduce the defect. It added the
+separate restart-held Queue and Resume Queue semantics that this fix must preserve.
+
+## Fix Design
+
+- Add durable `retry_required` to `PendingSessionInputState`. SQLite already stores unconstrained
+  text, so no schema migration is required.
```
zerob13: [P2] Treat `retry_required` as a persisted contract change even though the SQLite column needs no DDL. A downgrade binary still sees schema v46 and reads this row, but its waiting-row filter excludes the state while its renderer presents ordinary edit/reorder/Steer controls; those mutations reject the unknown state, leaving deletion as the only recovery. The repository SDD contract requires upgrade/rollback considerations for stored user data. Add an explicit downgrade strategy or schema/version fence, and cover/document the behavior of active retry rows when rolling back.

zerob13 (commented):
One additional persisted-data compatibility issue:

src/main/session/data/pendingInputStore.ts
yyhhyyyyyy: Fixed by allowing per-item retry behind earlier Queue rows while leaving execution order to the FIFO pump.

src/main/session/data/pendingInputStore.ts
yyhhyyyyyy: Fixed by preserving claimed Queue slots during atomic move, delete, and conversion resequencing, with regression coverage.

docs/issues/pending-input-explicit-retry/spec.md
yyhhyyyyyy: Fixed with schema v67 and a downgrade-safe blocked-state marker, plus migration and recovery coverage.

PR #2142, opened 2026-08-11
feat(chat): dedupe workspace labels, split sidebar

### LINKED ISSUE #2121: [feature] Disambiguate duplicate workspace names in the sidebar
## Summary

The sidebar uses only the final path segment as a workspace label. Two managed directories with the same basename are therefore visually indistinguishable.

**Priority:** P2 UX

## Current behavior

`getProjectGroupLabel()` in `src/renderer/src/stores/ui/session.ts` derives the label with a basename-style split and does not use parent context.

This creates common collisions, for example:

```text
.../team-a/app
.../archive/app
```

Both groups are rendered as `app`. Cross-platform history, nested copies, monorepo checkouts, and similarly named client projects make this a normal case rather than an edge case.

- `docs/features/complete-directory-management/spec.md` treats each normalized directory path as a distinct managed environment.
- `docs/architecture/session-management.md` states that Project owns directory lifecycle while Session stores only `projectDir`.

The identity is path-based, but the current presentation removes enough path information to make distinct identities ambiguous.

## Impact

- Users can open or move a conversation under the wrong workspace.
- Drag/reorder actions have no reliable visual target.
- Missing and cross-platform environments are especially difficult to identify.
- Full paths in every row would solve ambiguity but add unnecessary visual noise.

## Proposed direction

- Keep the basename for unique labels.
- Detect collisions among currently visible/active groups.
- For collisions only, append the shortest parent suffix that makes the label unique, for example:

```text
app · team-a
app · archive
```

- Keep the normalized full path in an accessible tooltip/details surface.
- Recompute disambiguation when environment or filter state changes.

## Acceptance criteria

- [ ] Unique basenames retain their current compact label.
- [ ] Duplicate basenames receive deterministic minimal parent context.
- [ ] Labels remain unique when more than two paths share multiple parent segments.
- [ ] Windows and POSIX separators are handled consistently.
- [ ] Full normalized paths are available to keyboard and assistive-technology users.
- [ ] Tests cover same basename, nested duplicates, cross-platform path strings, and renamed/removed environments.

## User benefit

Users can reliably identify the intended project without turning the entire sidebar into a list of long absolute paths.

zhangmo8:
## Summary

- **Closes #2121** — duplicate workspace basenames in the sidebar now get the shortest parent suffix that makes them unique (`app · team-a` / `app · archive`), computed reactively over the currently visible group set (session groups **and** environment-only groups). Unique basenames keep their compact label; when a colliding workspace is archived/removed, the survivor shrinks back automatically. The full normalized path is exposed on the group header via `title` and an `sr-only` span for keyboard/AT users.
- **Visual separation of Chat vs Workspace** — the chat section gets a subtle `bg-muted/30` rounded card and the workspace title a hairline top divider (shown only when there is content above), so the two identical-looking lists read as separate zones. Class-only change, no DOM restructuring.
- **Refactor: `WindowSideBar.vue` 2513 → 1199 lines** — the script is split into seven focused composables under `composables/sidebar/`; raw `window.*` / `document.*` usage is replaced with VueUse primitives (`useEventListener`, `useDocumentVisibility`, `useResizeObserver`, `useTimeoutFn`, `usePreferredReducedMotion`, `tryOnScopeDispose`). Template and DOM structure are unchanged.

## BEFORE / AFTER (sidebar layout)

```text
BEFORE                              AFTER
──────────────────────────         ──────────────────────────
📌 Pinned                          📌 Pinned
   session row                        session row
Chats                    [+]       ╭────────────────────────╮
   session row                     │ Chats              [+] │  ← bg-muted/30 card
Workspace          [📁+] [⊞]       │   session row          │
  📂 app                 [+]       ╰────────────────────────╯
   session row                     ──────────────────────────  ← divider
  📂 app                 [+]       Workspace          [📁+] [⊞]
   session row                       📂 app · team-a       [+]
       ↑ indistinguishable            session row
                                     📂 app · archive      [+]
                                       session row
```

## New composables

| File | Responsibility |
|---|---|
| `useSidebarWorkspaceGroups` | grouping pipeline, environment merge, label disambiguation, collapse state |
| `useSessionPinFlight` | pin flight animation + feedback, reduced-motion aware |
| `useSessionListAutoFill` | scroll pagination + viewport auto-fill (#1762 behavior preserved) |
| `useSidebarSessionShortcuts` | Cmd/Alt+digit shortcuts and badges |
| `useProjectGroupReorder` | drag & menu reordering of workspace groups |
| `useSidebarWorkspaceActions` | add / archive / reveal workspace |
| `useSidebarRemoteControl` | remote status polling with backoff + visibility pause |

## Verification

- `test/main/shared/workspaceLabels.test.ts` — 6 unit tests (same basename, nested duplicates, cross-platform separators, parentless paths, collision removal)
- `WindowSideBar.test.ts` 65/65, `WindowSideBarSessionItem` + `App.startup` 24/24, `test/main/shared` 112/112; full renderer suite passes except 4 files already failing on `dev` (reproduced on a clean tree, unrelated)
- Two tests that reached into component internals via `wrapper.vm` were rewritten against observable behavior (real window keydown, poll timer)
- `oxfmt`, `oxlint`, `typecheck:web`, `typecheck:node`, `pnpm run lint`, `pnpm run i18n` all clean
- Verified in the built app (Playwright `_electron`): disambiguated labels, sr-only paths, section separation render correctly with zero renderer console errors
- Behavior double-checked against `dev`: cold-start pagination (30-per-page; sessions of later groups appear after scrolling) is identical before/after this PR — pre-existing behavior, see note below

## Note for reviewers

While verifying, we confirmed a pre-existing UX quirk (unchanged by this PR): in project grouping, managed-environment group headers always render, but their sessions only appear once pagination reaches them, so a group can look temporarily empty on cold start. Worth a separate issue; happy to file it.

## Summary by CodeRabbit

* **New Features**
  * Reorder project workspace groups by drag-and-drop or menu actions.
  * Use keyboard shortcuts to quickly select visible sessions.
  * Add and archive workspaces directly from the sidebar.
  * Automatically load more sessions as you scroll.

* **Improvements**
  * Added smoother pinning and unpinning animations.
  * Improved remote-control status updates and navigation.
  * Clarified duplicate workspace names with directory context.
  * Enhanced workspace grouping, labels, tooltips, and accessibility.

* **Tests**
  * Expanded coverage for workspace labels, shortcuts, scrolling, pinning, and remote-control updates.

### DIFF
--- a/src/renderer/src/components/WindowSideBar.vue
+++ b/src/renderer/src/components/WindowSideBar.vue
@@ -412,64 +412,72 @@
                       : ''
                   ]"
                 >
-                  <button
-                    type="button"
-                    class="flex min-w-0 flex-1 items-center gap-1.5 rounded-md px-2 py-1.5 text-left"
-                    :class="
-                      isProjectGroupReorderTarget(group) && canReorderProjectGroups
-                        ? 'sidebar-project-folder-target cursor-grab active:cursor-grabbing'
-                        : ''
-                    "
-                    :data-group-id="getGroupIdentifier(group)"
-                    :title="isProjectDirectoryGroup(group) ? getGroupIdentifier(group) : undefined"
-                    :aria-expanded="getWorkspaceGroupAriaExpanded(group)"
-                    @click="handleWorkspaceGroupClick(group)"
-                  >
-                    <span class="shrink-0 size-6 flex items-center justify-center">
-                      <Icon
-                        :icon="getGroupIcon(group)"
-                        :data-icon="getGroupIcon(group)"
-                        data-testid="window-sidebar-group-icon"
-                        class="size-4"
-                      />
-                    </span>
-                    <span class="truncate">
-                      {{ getGroupLabel(group) }}
-                    </span>
-                    <span v-if="isProjectDirectoryGroup(group)" class="sr-only">
+                  <Tooltip>
+                    <TooltipTrigger as-child>
+                      <button
+                        type="button"
+                        class="flex min-w-0 flex-1 items-center gap-1.5 rounded-md px-2 py-1.5 text-left"
+                        :class="
+                          isProjectGroupReorderTarget(group) && canReorderProjectGroups
+                            ? 'sidebar-project-folder-target cursor-grab active:cursor-grabbing'
+                            : ''
+                        "
+                        :data-group-id="getGroupIdentifier(group)"
+                        :aria-expanded="getWorkspaceGroupAriaExpanded(group)"
+                        @click="handleWorkspaceGroupClick(group)"
+                      >
+                        <span class="shrink-0 size-6 flex items-center justify-center">
+                          <Icon
+                            :icon="getGroupIcon(group)"
+                            :data-icon="getGroupIcon(group)"
+                            data-testid="window-sidebar-group-icon"
+                            class="size-4"
+                          />
+                        </span>
+                        <span class="truncate">
+                          {{ getGroupLabel(group) }}
+                        </span>
+                        <span
+                          v-if="isTrueEmptyWorkspaceGroup(group)"
+                          data-testid="window-sidebar-empty-workspace-label"
+                          class="ms-auto shrink-0 text-[10px] font-normal text-muted-foreground/70"
+                        >
+                          {{ t('chat.sidebar.emptyWorkspace') }}
+                        </span>
+                        <span
+                          v-if="isWorkspaceUnavailable(group)"
+                          class="ms-auto flex shrink-0 items-center"
+                          :title="
+                            t('chat.input.workspaceUnavailableTooltip', {
+                              path: getWorkspacePath(group)
+                            })
+                          "
+                        >
+                          <Icon
+                            icon="lucide:circle-alert"
+                            data-testid="window-sidebar-workspace-unavailable"
+                            aria-hidden="true"
+                            class="size-3.5 text-amber-500"
+                          />
+                        </span>
+                        <span v-if="isWorkspaceUnavailable(group)" class="sr-only">
+                          {{
+                            t('chat.input.workspaceUnavailableTooltip', {
+                              path: getWorkspacePath(group)
+                            })
+                          }}
+                        </span>
+                      </button>
+                    </TooltipTrigger>
+                    <TooltipContent
+                      v-if="isProjectDirectoryGroup(group)"
+                      side="right"
+                      data-testid="workspace-path-tooltip"
+                      class="max-w-72 break-all"
+                    >
                       {{ getGroupIdentifier(group) }}
-                    </span>
-                    <span
-                      v-if="isTrueEmptyWorkspaceGroup(group)"
-                      data-testid="window-sidebar-empty-workspace-label"
-                      class="ms-auto shrink-0 text-[10px] font-normal text-muted-foreground/70"
-                    >
-                      {{ t('chat.sidebar.emptyWorkspace') }}
-                    </span>
-                    <span
-                      v-if="isWorkspaceUnavailable(group)"
-                      class="ms-auto flex shrink-0 items-center"
-                      :title="
-                        t('chat.input.workspaceUnavailableTooltip', {
-                          path: getWorkspacePath(group)
-                        })
-                      "
-                    >
-                      <Icon
-                        icon="lucide:circle-alert"
-                        data-testid="window-sidebar-workspace-unavailable"
-                        aria-hidden="true"
-                        class="size-3.5 text-amber-500"
-                      />
-                    </span>
-                    <span v-if="isWorkspaceUnavailable(group)" class="sr-only">
-                      {{
-                        t('chat.input.workspaceUnavailableTooltip', {
-                          path: getWorkspacePath(group)
-                        })
-                      }}
-                    </span>
-                  </button>
+                    </TooltipContent>
+                  </Tooltip>
 
                   <DcButton
                     v-if="canStartConversationInProjectGroup(group)"
@@ -666,7 +674,12 @@ import draggable from 'vuedraggable'
 import { Icon } from '@iconify/vue'
 import { DcButton } from '@dc-ui/components/button'
 import { DcEmpty } from '@dc-ui/components/empty'
-import { TooltipProvider } from '@shadcn/components/ui/tooltip'
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger
+} from '@shadcn/components/ui/tooltip'
 import { Input } from '@shadcn/components/ui/input'
 import { Skeleton } from '@shadcn/components/ui/skeleton'
 import {
--- a/src/renderer/src/components/WindowSideBar.vue
+++ b/src/renderer/src/components/WindowSideBar.vue
@@ -284,7 +284,7 @@
             </div>
           </div>
 
-          <div v-if="chatSectionGroup" class="pt-4">
+          <div v-if="chatSectionGroup" class="mt-4 rounded-lg bg-muted/30 p-1">
             <div
               class="group flex w-full items-center gap-1 rounded-md pr-1 text-xs font-semibold text-muted-foreground transition-colors duration-150 hover:bg-accent/40 hover:text-foreground focus-within:bg-accent/40 focus-within:text-foreground"
               :class="
@@ -343,7 +343,14 @@
             </TransitionGroup>
           </div>
 
-          <div class="flex items-center justify-between gap-2 px-2 pb-1 pt-4">
+          <div
+            class="flex items-center justify-between gap-2 px-2 pb-1"
+            :class="
+              pinnedSessions.length > 0 || chatSectionGroup
+                ? 'mt-3 border-t border-border/60 pt-3'
+                : 'pt-4'
+            "
+          >
             <div class="min-w-0 truncate text-xs font-semibold text-muted-foreground">
               {{ t('chat.sidebar.workspace') }}
             </div>
@@ -414,6 +421,7 @@
                         : ''
                     "
                     :data-group-id="getGroupIdentifier(group)"
+                    :title="isProjectDirectoryGroup(group) ? getGroupIdentifier(group) : undefined"
                     :aria-expanded="getWorkspaceGroupAriaExpanded(group)"
                     @click="handleWorkspaceGroupClick(group)"
                   >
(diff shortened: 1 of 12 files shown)


src/renderer/src/composables/sidebar/useProjectGroupReorder.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    for (const path of nextVisiblePaths) {
+      if (!nextOrder.includes(path)) {
+        nextOrder.push(path)
+      }
+    }
+
+    await projectStore.reorderEnvironments(nextOrder)
+  }
```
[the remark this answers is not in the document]
zhangmo8: Fixed in a0ea81cb9 — `commitVisibleProjectGroupOrder` now warns and returns early when `nextVisiblePaths` doesn't match the visible group count, so a short list can no longer persist a duplicated/dropped order.

src/renderer/src/composables/sidebar/useSessionPinFlight.ts
```diff
@@ -0,0 +1,330 @@
...
+      applyPinFeedback(session.id, nextPinned)
+      pinFlightSessionId.value = null
+      await nextTick()
+    } finally {
+      if (pinDockedSessionId.value === session.id) {
+        pinDockedSessionId.value = null
+      }
+      pinFlightSessionId.value = null
+      clone.remove()
+    }
```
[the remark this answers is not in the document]
zhangmo8: Fixed in a0ea81cb9 — the `pinFlightSessionId` resets are now guarded by session id, mirroring the `pinDockedSessionId` checks. Also applied the same guard to the third reset site in the no-target fallback path (the early-return branch), which had the identical clobbering issue.

src/renderer/src/composables/sidebar/useSidebarRemoteControl.ts
```diff
@@ -0,0 +1,197 @@
...
+      return
+    }
+    const backoffMs = Math.min(30_000, 2_000 * 2 ** statusRefreshErrors)
+    scheduleStatusRefresh(
+      hasEnabledRemoteChannelStatus()
+        ? refreshed
+          ? REMOTE_STATUS_ACTIVE_POLL_MS
+          : backoffMs
+        : REMOTE_STATUS_IDLE_POLL_MS
+    )
```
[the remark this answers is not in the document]
zhangmo8: Fixed in a0ea81cb9 — `runStatusRefresh` now increments a composable-local sequence token and passes it down; a superseded request no longer replaces the snapshot, counts toward backoff, or schedules another poll (which could previously also fork the poll chain). Added a regression test that overlaps a slow poll refresh with a hide/show-triggered refresh and asserts the stale result is discarded — verified it fails without the fix.

src/shared/utils/workspaceLabels.ts
```diff
@@ -0,0 +1,65 @@
...
+        .split(/[\\/]+/)
+        .filter(Boolean)
+        .slice(0, -1)
+    )
+    const maxDepth = Math.max(...parentSegments.map((segments) => segments.length))
+
+    let contexts = parentSegments.map(() => '')
+    for (let depth = 1; depth <= maxDepth; depth += 1) {
+      contexts = parentSegments.map((segments) => segments.slice(-depth).join('/'))
+      if (new Set(contexts).size === bucket.length) {
```
zerob13: [P2] Compute the shortest suffix per workspace

This loop chooses one depth for the entire collision bucket. For `/root/team/shared/app`, `/root/archive/shared/app`, and `/root/client/unique/app`, the last label becomes `app · client/unique`, although `app · unique` is already unique. That violates #2121's deterministic minimal-parent-context criterion. Resolve each entry at the first depth where its own suffix occurs once, and add this asymmetric case to the unit test.

src/renderer/src/components/WindowSideBar.vue
```diff
@@ -414,6 +421,7 @@
                         : ''
                     "
                     :data-group-id="getGroupIdentifier(group)"
+                    :title="isProjectDirectoryGroup(group) ? getGroupIdentifier(group) : undefined"
```
zerob13: [P2] Expose the full path on keyboard focus

A native `title` is not reliably exposed when a sighted keyboard user focuses this button, while the `sr-only` span only serves assistive technology. The keyboard half of #2121's full-path acceptance criterion therefore remains unmet. Use the existing tooltip primitive with hover and focus behavior, associate its content with the button, and avoid announcing the same path again through both `title` and the accessible name.

src/renderer/src/composables/sidebar/useSessionPinFlight.ts
```diff
@@ -0,0 +1,336 @@
...
+    const preservedScrollTop = sessionListRef.value?.scrollTop ?? null
+
+    if (!sourceElement || !sourceRect || sourceRect.width === 0 || sourceRect.height === 0) {
+      await commitPinToggle(session, nextPinned)
+      return
+    }
+
+    const clone = createPinFlightClone(sourceElement, sourceRect)
+    document.body.appendChild(clone)
+    pinFlightSessionId.value = session.id
```
zerob13, line 240: [P2] Serialize pin flights or track every active ID

The guarded cleanup prevents an older flight from clearing a newer ID, but this single assignment still drops ownership of the older flight immediately. Pin A, then pin B before A's ~460 ms animation finishes: the value becomes B, so A's `hero-hidden`/placeholder props turn off while clone A is still animating, rendering the real row and clone together; `pinDockedSessionId` has the same problem for unpins. Either reject/queue toggles while a flight is active or model the active flight/docked IDs as sets, and cover two overlapping sessions.

zerob13 (commented):
Targeted checks pass (`workspaceLabels`, `WindowSideBar`, the scroll-architecture test, both typechecks, Oxfmt, and Oxlint). I found three additional issues below. The existing unresolved threads about reorder-path membership and the guarded remote snapshot result also still need attention.

src/shared/utils/workspaceLabels.ts
```diff
@@ -0,0 +1,65 @@
...
+      contextCounts.set(context, (contextCounts.get(context) ?? 0) + 1)
+    }
+
+    bucket.forEach((item, index) => {
+      const context = contexts[index]
+      // Identical parent chains cannot be separated by a suffix; fall back to the full
+      // normalized path so every rendered label stays unique.
+      const uniqueContext = contextCounts.get(context) === 1 ? context : item.id
+      if (uniqueContext) {
+        overrides.set(item.id, `${item.label} · ${uniqueContext}`)
```
yyhhyyyyyy: [P2] Check uniqueness after generating overrides

This only buckets original labels. For `/work/team-a/app`, `/work/archive/app`, and `/work/app · team-a`, the rendered labels become `app · team-a`, `app · archive`, and `app · team-a`. Two visible workspaces remain indistinguishable, defeating the feature’s uniqueness guarantee.

src/renderer/src/composables/sidebar/useProjectGroupReorder.ts
```diff
@@ -0,0 +1,180 @@
...
+      : projectReorderableGroups.value.map(options.getWorkspacePath)
+  }
+
+  const commitVisibleProjectGroupOrder = async (nextVisiblePaths: string[]) => {
+    const currentOrder = getCurrentProjectOrderPaths()
+    const previousVisiblePaths = projectReorderableGroups.value.map(options.getWorkspacePath)
+    if (nextVisiblePaths.length !== previousVisiblePaths.length) {
+      console.warn('[WindowSideBar] Skipping project group reorder: visible group count mismatch')
+      return
+    }
```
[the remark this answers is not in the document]
zhangmo8: Fixed in d6bb5057d — the guard now rejects any `nextVisiblePaths` that is not a permutation of the visible set: lengths must match, both lists must be duplicate-free, and every next path must exist in the previous set.

src/renderer/src/composables/sidebar/useSidebarRemoteControl.ts
```diff
@@ -0,0 +1,210 @@
...
+      const descriptors = await remoteControlClient.listRemoteChannels()
+      const channels = descriptors.map((descriptor) => descriptor.id)
+      const statuses = await Promise.all(
+        channels.map((channel) => remoteControlClient.getChannelStatus(channel))
+      )
+      if (requestId !== statusRefreshSequence) {
+        return true
+      }
+      pluginCatalogStore.replaceRemoteSnapshot(descriptors, statuses, version)
+      return true
```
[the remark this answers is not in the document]
zhangmo8: Fixed in d6bb5057d — `refreshRemoteControlStatus` now returns a tri-state outcome (`applied` / `discarded` / `failed`). A discarded snapshot (superseded request or store version conflict) no longer resets `statusRefreshErrors`; it also doesn't count toward backoff since the fetch itself succeeded — the next poll is scheduled at the normal cadence to reconcile quickly after a user mutation. Only genuine fetch errors grow the backoff.

src/shared/utils/workspaceLabels.ts
zhangmo8: Fixed in d6bb5057d. `disambiguateWorkspaceLabels` now resolves each workspace at the first depth where its own suffix occurs once among the workspaces still colliding, so one deep collision no longer lengthens every label in the bucket. Your asymmetric case is covered in `workspaceLabels.test.ts` (`resolves each workspace at its own shortest unique suffix`): `/root/team/shared/app` → `app · team/shared`, `/root/archive/shared/app` → `app · archive/shared`, `/root/client/unique/app` → `app · unique`. Full-path fallback for identical parent chains and the single-parentless-duplicate compact-label behavior are preserved.

src/renderer/src/components/WindowSideBar.vue
zhangmo8: Fixed in d6bb5057d. The group header button is now wrapped in the existing reka-ui tooltip primitives (`Tooltip`/`TooltipTrigger as-child`/`TooltipContent`), which open on hover and on keyboard focus and associate the content with the button via `aria-describedby` while open. The native `title` and the `sr-only` path span are removed so the path is announced once, through the tooltip association, instead of through both `title` and the accessible name. `as-child` keeps the DOM structure unchanged (no wrapper element).

src/renderer/src/composables/sidebar/useSessionPinFlight.ts
zhangmo8: Fixed in d6bb5057d — went with the serialize option: `handleTogglePin` now queues toggles on a promise chain, so a toggle requested mid-flight starts only after the previous flight fully settles and the single flight/docked ownership refs are never contested (`nextPinned` is computed at execution time, so queued same-session toggles see the committed state). The session-id guards stay as defense in depth. Added `serializes overlapping pin toggles across two sessions` to `WindowSideBar.test.ts`, which holds the first commit open and asserts the second toggle does not start until it resolves — verified it fails without the queue.

yyhhyyyyyy (approved):
LGTM

PR #2143, opened 2026-08-12
fix(project): prepend new workspaces

### LINKED ISSUE #2116: [bug] Insert newly added workspaces at the top of persisted order
## Summary

After a user has persisted a custom workspace order, every newly added workspace is placed after all existing workspaces. The expected product behavior is to place the newly added workspace first while preserving the relative order of existing entries.

**Priority:** P1

## Root cause

The order is not controlled by `new_projects.last_accessed_at` in this UI path. It is controlled by environment preferences:

- `reorderActive()` in
  `src/main/project/data/tables/newEnvironmentPreferences.ts` writes explicit `sort_order = 0..N` values.
- `markActive()` delegates to `setStatus()`, which inserts a new row with `DEFAULT_ENVIRONMENT_SORT_ORDER` (the maximum/default sentinel).
- `ProjectService.compareEnvironmentSummaries()` in `src/main/project/index.ts` always places explicit orders before the default sentinel.

Once a user has reordered the list, all existing workspaces have explicit values. The next newly selected directory therefore deterministically appears at the end.

- `docs/features/complete-directory-management/spec.md` defines persisted workspace ordering and says restored directories return near the top.
- The same contract distinguishes explicit user order from recent-usage fallback.
- Issue #1785 expected existing users to receive a reasonable default order after migration.

The current insert behavior does not define how a new active environment joins an already explicit order.

## Impact

- The workspace the user just added can be outside the visible viewport.
- The user must immediately drag it from the bottom to the top.
- The behavior feels like the add action failed when the list is long.
- “Recently added” and persisted manual ordering interact inconsistently.

## Proposed direction

Perform a transactional prepend when a previously unknown environment becomes active:

1. Shift existing active explicit orders by one.
2. Insert the new environment at `sort_order = 0`.
3. Preserve the existing entries' relative order.

Restoring an archived environment should follow an explicit documented rule:

- restore to its previous position; or
- treat restore as re-activation and prepend it.

The code and maintained directory contract must agree on one behavior.

## Acceptance criteria

- [ ] A newly added workspace appears first in active environment order.
- [ ] Existing workspaces retain their relative order.
- [ ] The update is atomic and cannot leave duplicate/gapped effective positions after failure.
- [ ] Re-selecting an already active workspace does not reorder it unexpectedly.
- [ ] Archive/restore behavior is explicitly specified and covered by tests.
- [ ] Sidebar and Settings show the same order after restart.

## User benefit

The directory the user just added is immediately visible and ready to use, without an extra drag operation.

zhangmo8:
## Summary

- prepend newly activated workspaces with an explicit `sort_order = 0`
- shift existing explicit active orders in the same SQLite transaction
- keep re-selected active workspaces in place and make Settings restore follow the documented prepend rule
- cover successful prepend, active re-selection, restore, and rollback on failure

## Root cause

Persisted custom ordering assigns explicit values to existing workspaces, while a newly activated
workspace previously received the default sentinel (and later a negative minimum in the sidebar
flow). That did not satisfy the maintained dense prepend contract and left the restore rule
ambiguous.

## User impact

The workspace a user just adds or restores is immediately visible at the top, while every existing
workspace keeps its relative position. A failed insert rolls back the preceding order shift
atomically.

## BEFORE / AFTER

```text
BEFORE                    AFTER
0  workspace-a            0  new-workspace
1  workspace-b    -->     1  workspace-a
2  workspace-c            2  workspace-b
                          3  workspace-c
```

## Validation

- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`
- `ELECTRON_RUN_AS_NODE=1 pnpm exec electron node_modules/vitest/vitest.mjs run --config vitest.config.ts test/main/project/data/tables/newEnvironmentPreferencesTable.test.ts test/main/project/projectService.test.ts` (39 tests)

Closes #2116

## Summary by CodeRabbit

* **New Features**
  * Restored environments now appear at the top of the active directory list.
  * Existing active directories retain their relative order when a directory is restored or activated.
  * Directory ordering now remains dense and nonnegative.

* **Bug Fixes**
  * Re-activating an already active directory no longer changes its position.
  * Failed activation updates now roll back without altering the existing directory order.

### DIFF
--- a/src/main/project/data/tables/newEnvironmentPreferences.ts
+++ b/src/main/project/data/tables/newEnvironmentPreferences.ts
@@ -75,16 +75,9 @@ export class NewEnvironmentPreferencesTable extends BaseTable {
     }
 
     const now = Date.now()
-    const getStatus = this.db.prepare(
-      'SELECT status FROM new_environment_preferences WHERE path = ?'
-    )
-    const shiftActiveOrder = this.db.prepare(
-      `UPDATE new_environment_preferences
-       SET sort_order = sort_order + 1
-       WHERE status = 'active' AND sort_order < ${DEFAULT_ENVIRONMENT_SORT_ORDER}`
-    )
-    const activate = this.db.prepare(
-      `INSERT INTO new_environment_preferences (
+    this.db
+      .prepare(
+        `INSERT INTO new_environment_preferences (
           path,
           status,
           sort_order,
@@ -94,30 +87,27 @@ export class NewEnvironmentPreferencesTable extends BaseTable {
         ) VALUES (
           ?,
           'active',
-          0,
+          COALESCE((
+            SELECT MIN(sort_order) - 1
+            FROM new_environment_preferences
+            WHERE status = 'active' AND sort_order < ${DEFAULT_ENVIRONMENT_SORT_ORDER}
+          ), 0),
           NULL,
           NULL,
           ?
         )
         ON CONFLICT(path) DO UPDATE SET
           status = 'active',
-          sort_order = excluded.sort_order,
+          sort_order = CASE
+            WHEN new_environment_preferences.status = 'active'
+              THEN new_environment_preferences.sort_order
+            ELSE excluded.sort_order
+          END,
           archived_at = NULL,
           removed_at = NULL,
           updated_at = excluded.updated_at`
-    )
-
-    this.db.transaction(() => {
-      const existing = getStatus.get(normalizedPath) as
-        | Pick<NewEnvironmentPreferenceRow, 'status'>
-        | undefined
-      if (existing?.status === 'active') {
-        return
-      }
-
-      shiftActiveOrder.run()
-      activate.run(normalizedPath, now)
-    })()
+      )
+      .run(normalizedPath, now)
   }
 
   markArchived(environmentPath: string): void {
--- a/src/main/project/data/tables/newEnvironmentPreferences.ts
+++ b/src/main/project/data/tables/newEnvironmentPreferences.ts
@@ -75,9 +75,16 @@ export class NewEnvironmentPreferencesTable extends BaseTable {
     }
 
     const now = Date.now()
-    this.db
-      .prepare(
-        `INSERT INTO new_environment_preferences (
+    const getStatus = this.db.prepare(
+      'SELECT status FROM new_environment_preferences WHERE path = ?'
+    )
+    const shiftActiveOrder = this.db.prepare(
+      `UPDATE new_environment_preferences
+       SET sort_order = sort_order + 1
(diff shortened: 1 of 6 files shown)


src/main/project/data/tables/newEnvironmentPreferences.ts
```diff
@@ -75,9 +75,16 @@ export class NewEnvironmentPreferencesTable extends BaseTable {
...
     const now = Date.now()
-    this.db
-      .prepare(
-        `INSERT INTO new_environment_preferences (
+    const getStatus = this.db.prepare(
+      'SELECT status FROM new_environment_preferences WHERE path = ?'
+    )
+    const shiftActiveOrder = this.db.prepare(
+      `UPDATE new_environment_preferences
+       SET sort_order = sort_order + 1
```
zerob13: [P1] Handle negative orders already persisted by dev

The current dev implementation of activateAtTop legitimately stores MIN(sort_order) - 1, so an existing database can contain active explicit orders such as -1 and 0. This UPDATE turns them into 0 and 1, and the following upsert writes the activated path at 0, creating a duplicate. compareEnvironmentSummaries breaks equal explicit orders by lastUsedAt, so a newly added or restored path is no longer guaranteed to appear first. There is no migration in this PR. Please normalize the existing active explicit rows to contiguous values starting at 1 before writing the target at 0, and add a regression test seeded with negative pre-PR orders.

src/main/project/data/tables/newEnvironmentPreferences.ts
```diff
@@ -87,27 +94,30 @@ export class NewEnvironmentPreferencesTable extends BaseTable {
...
-          END,
+          sort_order = excluded.sort_order,
           archived_at = NULL,
           removed_at = NULL,
           updated_at = excluded.updated_at`
-      )
-      .run(normalizedPath, now)
+    )
+
+    this.db.transaction(() => {
```
zerob13: [P2] Avoid adding a multi-statement invariant unless the product needs it

The persisted contract currently consumes only relative order: compareEnvironmentSummaries accepts any explicit integer, and the base branch single UPSERT with MIN(sort_order) - 1 already prepends atomically while preserving the order of existing rows and active duplicates. Replacing it with a status read, a full active-row UPDATE, and an upsert rewrites every explicitly ordered workspace and creates the partial-failure mode that the trigger-based rollback test then has to defend, without a user-visible benefit.

Unless zero-based, nonnegative, dense storage is made an explicit invariant with migration and normalization, please keep the existing single-statement implementation and limit this change to making restoreEnvironment call activateAtTop. In that version the rollback trigger test and this defensive machinery can be removed. If dense storage is intentional, document that invariant and reindex legacy/gapped rows rather than merely incrementing them.

PR #2147, opened 2026-08-13
feat(skills): add progressive disclosure

yyhhyyyyyy:
## Summary

- Add a bounded Skill routing catalog with adaptive token budgets and paginated `skill_list` discovery.
- Unify Skill activation around canonical effective content: `SKILL.md` body plus runtime instructions.
- Keep `@skill` and Skills-panel selections scoped to the current turn, including the first message.
- Return `skill_view` content through its tool result without duplicating it in the system prompt.
- Preserve legacy session-active Skills as visible, removable compatibility state.
- Materialize all provider-visible Skill instructions into Session Tape facts before projection.
- Improve context-window preflight, overflow recovery, provider-limit observations, and token-category diagnostics.

## Architecture

Skills now follow three explicit stages:

1. **Route** — a bounded system catalog containing compact routing cards.
2. **Discover** — `skill_list(query, cursor, limit)` searches enabled Skills without expanding bodies.
3. **Activate** — full effective content is projected according to activation scope:
   - Message selection → current active turn
   - `skill_view` → current tool result/tool loop
   - Session-active compatibility state → stable system section

Each request projects a given complete Skill body at most once. Historical message and root-view projections retain bounded markers instead of full Skill instructions.

## Tape guarantees

- Provider-visible Skill content is backed by content-addressed Session Tape facts.
- View manifests reference source entries and hashes rather than becoming a second content source.
- Continuations and recovery reuse the original materialization references.
- New executions fresh-resolve Skill sources and record version drift explicitly.
- Materialization, manifest, and integrity failures fail closed before provider execution.
- Skill context facts remain excluded from transcript projection, Tape search/context, Memory ingestion and recall, and fork merges.
- Runtime components receive narrow Tape capabilities rather than unrestricted append access.

## Context handling

- Keep `ContextCoordinator` as the authoritative admission boundary.
- Add a conservative non-persistent early guard before Skill materialization.
- Record provider-reported prompt/total limits with their correct scope.
- Recover context overflows when either the projected messages or effective output limit changes.
- Report request composition through an attributed context ledger.
- Recognize common OpenAI, Anthropic, and Gemini context-limit error formats.

### Before

```text
New conversation
└─ Select Skill
   └─ First turn could silently persist it for the whole session

Historical skill_view
└─ Full Skill body remained in later provider requests
```

### After

```text
New or existing conversation
└─ @Skill / panel selection
   └─ Current turn only
      ├─ Tape stores the verified materialization
      └─ History keeps only [Used skill: name]

Legacy session-active Skill
└─ Remains persistent
   ├─ Clearly visible
   └─ Can be removed
```

## Validation

- `pnpm run test:main` — 7,209 passed
- `pnpm run test:renderer` — 2,133 passed
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run build`

## Summary by CodeRabbit

- **New Features**
  - Added bounded, searchable, paginated skill discovery.
  - Added progressive skill activation for message and session scopes.
  - Added session skill indicators with per-skill removal controls.
  - Improved skill execution safeguards, cancellation, output limits, and environment validation.
  - Added clearer context-window overflow diagnostics and recovery behavior.

- **Bug Fixes**
  - Prevented internal skill context data from appearing in transcripts, search, memory, replay, or forks.
  - Improved cleanup of session skills and temporary execution files.

- **Documentation**
  - Added comprehensive progressive disclosure architecture and implementation documentation.


src/main/skill/index.ts
```diff
@@ -4234,12 +5176,26 @@ export class SkillService implements SkillServicePort {
...
+    // tombstone until this explicit lifecycle boundary rather than retaining every deleted ID for
+    // the lifetime of the process.
+    this.retiredSessionSkillScopes.delete(conversationId)
   }
 
   async revalidateActiveSkillsForAgent(conversationId: string, agentId: string): Promise<string[]> {
+    if (this.retiredSessionSkillScopes.has(conversationId)) return []
     const persisted = this.getPersistedNewSessionSkills(conversationId)
     const valid = await this.validateSkillNames(agentId, persisted)
+    if (this.retiredSessionSkillScopes.has(conversationId)) return []
```
zerob13: [P2] Serialize agent revalidation with Session Skill mutations

`revalidateActiveSkillsForAgent` still runs outside `runActiveSkillMutation`: it reads the persisted list, awaits catalog validation, and may then write that old snapshot back. During the await, the new `skills.removeActive` route can remove a different still-valid Skill. If the snapshot also contains a Skill that is invalid for the newly assigned Agent, this branch writes the filtered stale snapshot and restores the concurrently removed Skill. The route guard only rejects `generating`; it does not exclude the Agent-reassignment window. Please run this whole read/validate/write sequence through the same per-Session queue (with the tombstone checks inside it) and cover the interleaving.

zerob13 (commented):
Found one concurrency issue in Session Skill persistence.

PR #2150, opened 2026-08-13
feat(skills): unify shared skill management

zerob13:
## Summary

- store mutable Skills once in a global catalog and derive per-DeepChat-Agent access through logical bindings
- migrate legacy Agent-scoped packages and state with restartable recovery while excluding ACP Agents
- make Plugins / Skills the sole management surface with global cards, preview-based Agent enablement, external Agent import, and secondary sync-directory access
- keep the task-completion Skill Draft suggestion setting above search and actions
- enforce the shared catalog across prompt assembly, tools, scripts, sessions, and filesystem authorization

## UI

Before:

```text
Settings / Skills
+----------------------------------------------------------+
| Agent selector | Assigned / Unassigned states            |
+----------------------------------------------------------+
| Mixed-size cards, source badges, and scoped actions      |
+----------------------------------------------------------+
```

After:

```text
Plugins / Skills
+----------------------------------------------------------+
| Suggest Skill Drafts                               [off] |
+----------------------------------------------------------+
| Search Skills                                            |
+----------------------------------------------------------+
| Sync directory | Import from external Agent              |
+----------------------------------------------------------+

+----------------------------+  +----------------------------+
| Skill name                 |  | Skill name                 |
| Two-line description...    |  | Two-line description...    |
+----------------------------+  +----------------------------+

Open card -> Preview -> Enabled Agents [+ Add Agent]
```

## Validation

- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run architecture:renderer-baseline:check`
- `pnpm run icons:check`
- Main Vitest: 561 files / 7,167 tests passed
- Renderer Vitest: 254 files / 2,118 tests passed
- Electron production build
- focused Playwright Skills/settings smoke and visual checks at 1280x900 and 760x720

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Skills page under Plugins.
  * Manage shared Skills, assign them to Agents, duplicate or delete them, and view enabled Agents.
  * Import Skills from external tools with previews, conflict warnings, reuse detection, and partial-result reporting.
  * Added guided onboarding support for the new Skills location.
* **Bug Fixes**
  * Improved runtime handling for explicitly selected Skills and protected inactive Skill packages.
* **Updates**
  * Refreshed supported model and external tool catalog information.
  * Updated Skills documentation and translations across supported languages.
  * Streamlined Skill synchronization to focus on discovery and external imports.


src/main/skill/index.ts
```diff
@@ -717,334 +695,449 @@ export class SkillService implements SkillServicePort {
...
+    if (!fs.existsSync(journalPath)) return targets
 
-    const rootExists = fs.existsSync(root)
-    const committedMigrationRoot = this.isCommittedAgentSkillMigrationRoot(
-      root,
-      normalizedAgentId,
-      selected
-    )
-    const preexistingIndependentRoot = rootExists && !committedMigrationRoot
+    const journal = JSON.parse(fs.readFileSync(journalPath, 'utf-8')) as {
```
zhangmo8, line 945: `readMigrationRecoveryTargets` parses the journal with no tolerance for a torn write, while the journal itself is produced by a plain `fs.writeFileSync` (L900-904). A crash or ENOSPC mid-write leaves a truncated `journal.json`; from then on every launch hits `JSON.parse` → throw → `migrateSharedSkills` fails, and since a failed migration is deliberately startup-fatal (per the comment in `composition.ts`), the app can no longer start until the user manually deletes `~/.deepchat/skills/.library-migration-v3`. The disk-full case is the nasty one: freeing space afterwards doesn't unbrick anything because the corrupt journal persists.

Writing the journal atomically (temp name in the same directory + `renameSync` over `journal.json`) closes this; with that in place the strict parse here is fine. A tolerant "treat corrupt as absent" fallback is *not* a good alternative — retries rewrite the journal after commits from a previous attempt may already exist, so dropping it could reallocate variant names.

src/main/skill/index.ts
```diff
@@ -5514,11 +6018,18 @@ export class SkillService implements SkillServicePort {
     const skillName =
       this.findSkillNameByPath(filePath, agentId) ?? path.basename(path.dirname(filePath))
     this.getMetadataCacheForAgent(agentId).delete(skillName)
-    this.getContentCacheForAgent(agentId).delete(skillName)
+    this.invalidateSkillContent(skillName)
+    const state = this.getStoredManagementState()
+    const affectedAgentIds = Object.entries(state.agents)
+      .filter(([, agent]) => agent.bindings[skillName]?.assigned === true)
+      .map(([assignedAgentId]) => assignedAgentId)
+    delete state.skills[skillName]
```
zhangmo8: A single watcher `delete` event now permanently erases the skill's `state.skills` entry plus every Agent's binding (assigned flag, per-Agent extension env, `runtimeBindingId`) — before this PR the handler only dropped caches. Two ways this fires for a skill that isn't actually gone:

- rename-based atomic saves (vim's default backup strategy, some sync tools) emit unlink→create for `SKILL.md`. The follow-up `handleSkillFileAdded` restores metadata and `state.skills` via `reconcileSkillManagementState`, but the per-Agent bindings are unrecoverable — the skill silently ends up unassigned for every Agent with its env config gone.
- when `findSkillNameByPath` misses (exact `metadata.path` match only), the `basename(dirname())` fallback can resolve a nested `<skill>/templates/<name>/SKILL.md` deletion to an unrelated live skill `<name>` and wipe its bindings.

Suggest gating the persistent prune on (a) the deleted path actually matching a cached skill's `metadata.path` and (b) `!fs.existsSync(filePath)` at handling time, and keeping plain cache invalidation for everything else — startup `reconcileSkillManagementState` already covers true deletions eventually.

src/main/skill/index.ts
```diff
@@ -1509,62 +1621,176 @@ export class SkillService implements SkillServicePort {
   }
 
   async setSkillDisabledForAgent(agentId: string, name: string, disabled: boolean): Promise<void> {
+    await this.setSkillAssignment(agentId, name, !disabled)
```
zhangmo8, line 1622: With `setSkillDisabledForAgent` delegating to the assignment path, nothing emits `skills.catalog.changed` with `reason: 'disabled-updated'` (or a `disabled` field) anymore — I checked all 20 emit sites. The renderer's targeted-update branch in `skillsStore.ts` (L398, `payload.reason === 'disabled-updated' && payload.disabled !== undefined`) is now dead code, so every toggle falls through to a full `loadSkills(agentId)` reload and the optimistic `applySkillDisabled` the store just did gets overwritten by the refetch. `'management-state-updated'` in the event contract is likewise orphaned (no emitter, no listener). Either re-emit a targeted reason from `setSkillAssignmentForAgentsUnlocked` or drop the dead enum values + store branch.

src/main/skill/index.ts
```diff
@@ -4331,115 +4651,272 @@ export class SkillService implements SkillServicePort {
...
       try {
-        this.deleteSkillManagementItem(name, agentId)
-      } catch (error) {
-        logger.warn('[SkillService] Failed to delete skill management state after uninstall', {
-          name,
-          error
+        const state = this.getStoredManagementState()
+        const bindings = this.getAgentBindingState(state, normalizedAgentId).bindings
+        bindings[targetName] = { assigned: true, extension }
+        bindings[name] = { assigned: false, extension }
```
zhangmo8: `duplicateSkillForAgent` doesn't just create a copy — it also flips the *source* skill to `assigned: false` for that Agent (fork-and-switch). Nothing in the renderer calls this route yet (only the main-process test does), so before a UI adopts it: is the swap the intended contract? If yes, the eventual action copy should say so, because under a plain "Duplicate" label unassigning the original will read as a bug. Side note: the source binding is rewritten without its `runtimeBindingId`, so a later re-assign regenerates the env binding — probably fine, just checking it's deliberate.

src/main/skill/index.ts
```diff
@@ -717,334 +695,449 @@ export class SkillService implements SkillServicePort {
...
+          }
+          this.saveManagementState(resumedState)
+        }
+      } else if (stored.migration?.status === 'planned') {
+        throw new Error('Shared Skills migration state was committed before its packages')
+      }
+      await this.pruneInactiveAgentBindings()
+      this.reconcileSkillManagementState()
+      await this.materializeProviderBindingsForExistingAgents()
+      fs.rmSync(path.join(this.skillsDir, SHARED_SKILL_MIGRATION_DIR), {
```
zhangmo8, line 717: After migration completes, the legacy per-Agent packages under `.agent-scopes/` stay on disk forever — only `.library-migration-v3` is removed. They're excluded from discovery and the watcher, and the FS handler denies Agent access, so from here on they're pure dead weight (a full duplicate of every migrated skill). Intentional as a rollback safety net? If so, a documented cleanup path (after the migration has been `completed` for a while, or a maintenance action) would keep long-lived profiles from carrying the duplicates indefinitely.

src/main/skill/skillTools.ts
```diff
@@ -35,7 +39,18 @@ export class SkillTools {
       return buildSkillListResult([], [], [], input)
     }
     const agentId = resolvedAgentId
-    const allSkills = await this.skillService.getMetadataList(agentId)
+    const assignedSkills = await this.skillService.getMetadataList(agentId)
+    const allSkills = [...assignedSkills]
+    if (activeSkillNames !== undefined) {
+      const listedNames = new Set(allSkills.map((skill) => skill.name))
+      const activeNames = new Set(activeSkillNames)
+      for (const skill of await this.skillService.getAllSkills()) {
```
zhangmo8, line 47: The snapshot semantics this PR adds keep an explicitly-activated skill usable after it's unassigned mid-session: the list result re-adds it here via `getAllSkills`, prompt assembly and `getActiveSkillsAllowedTools` trust the override, and `AgentToolManager` resolves its root through `getAllSkills`. The one path still enforcing assignment is the read step — `viewSkillInAgentScope` returns "Skill not found" for unassigned skills (`index.ts` L2411), and the view tool reaches it with the session's `conversationId`. So the model still sees the skill listed and keeps its allowed tools, but progressive disclosure (re-reading SKILL.md / referenced files) breaks. `viewSkillInAgentScope` already fetches the session's active list right below that check — treating session-active skills as visible there would close the gap.

src/main/skill/routes.ts
```diff
@@ -362,7 +395,6 @@ export function createSkillRoutes(deps: {
           action: 'updated',
           targetType: 'active-skills',
           targetLabel: 'active skills',
```
zhangmo8, line 386: `routeName` was removed from this `recordActivity` call (and from `cli/skillService.ts`), but the same cleanup was missed in `skillsRemoveActiveRoute` below — L415 still records `routeName: 'settings-skills'`, which this PR deletes from the settings router and contracts. The activity contract types `routeName` as a plain nullable string, so nothing catches the stale literal, the record persists to the activity table, and `SettingsOverview.openActivity` does an unguarded `router.push({ name: activity.routeName })` — clicking that entry throws vue-router's "No match" and navigates nowhere. Suggest dropping the field at L415 like the others, and adding a `router.hasRoute` guard in `openActivity` since already-persisted records from older builds hit the same dead link.

src/main/skill/agentSkillImportService.ts
```diff
@@ -203,170 +164,114 @@ export class AgentSkillImportService {
...
       return false
+    } finally {
+      await fs.promises
+        .rm(temporaryDirectory, { recursive: true, force: true })
+        .catch(() => undefined)
     }
   }
 
+  private async createDirectoryHash(root: string): Promise<string> {
+    const directoryHash = createHash('sha256')
```
zhangmo8, line 246: `createDirectorySnapshot` materializes the whole tree in memory as base64 (~2.7× folder size per side, two sides per comparison) and runs per conflicting skill — twice per import flow, since preview and execute both go through `resolveExternalItems`. Hashing each file (sha256) into the snapshot string, or short-circuiting on a file-list/size mismatch first, keeps the semantics and drops the spike.

src/renderer/settings/lib/guidedOnboardingSettings.ts
```diff
@@ -48,13 +48,22 @@ export async function continueGuidedOnboardingFromSettings(options: {
 
   const target = resolveGuidedOnboardingStepTarget(stepId)
 
+  if (target?.surface === 'plugins' && target.routeName === 'plugins-skills') {
+    if (router.hasRoute(target.routeName)) {
+      await router.push({ name: target.routeName })
+      return
+    }
+
+    persistGuidedOnboardingResumeIntent({
```
zhangmo8: This handoff can't work across windows: `persistGuidedOnboardingResumeIntent` writes `window.sessionStorage` (`onboardingResume.ts`), and sessionStorage is per-WebContents in Electron — the settings window is its own `BrowserWindow`, so the intent written here is invisible to the main window. The only consumer is `ChatMainApp.handleResumeGuidedOnboarding`, which reads the *main window's* sessionStorage on `appRuntime.windowFocused`. Net effect: after finishing the MCP step in the settings window, `focusMainWindow()` fires but the main window finds no intent and never navigates to Plugins > Skills — the guided flow silently stalls (it only "works" if a stale intent from WelcomePage happens to linger in the main window, and any interim focus consumes that).

This is a regression introduced by removing `settings-skills`: on dev the skills step resolved to a same-window `router.push`. Suggest routing the handoff through the main process (IPC to the main window, like `focusMainWindow` itself) or any storage both windows can see, rather than sessionStorage.

src/renderer/src/pages/plugins/SkillsPluginsPage.vue
```diff
@@ -1,7 +1,488 @@
...
+  try {
+    await skillClient.setSkillAssigned(skill.name, enabled, agentId)
+    const enabledAgentIds = new Set(skill.assignedAgentIds)
+    if (enabled) enabledAgentIds.add(agentId)
+    else enabledAgentIds.delete(agentId)
+    const updatedSkill = {
+      ...skill,
+      assigned: enabledAgentIds.size > 0,
+      assignedAgentIds: Array.from(enabledAgentIds)
+    }
```
zhangmo8, line 203: This success path writes `selectedSkill.value = updatedSkill` (built from the `skill` captured before the `await`) without checking that the selection still points at the same skill. `handleOpenChange` only blocks closing on `props.saving` — `agentUpdatePendingId` doesn't — so: open skill A → toggle an Agent → close the dialog before the IPC returns → open skill B → A's response lands and overwrites `selectedSkill`. The dialog then renders B's detail while enabled-Agents, `mutable`, and the Delete action all operate on A (`deleteSkill` reads `selectedSkill`). Narrow window, but the failure mode is deleting the wrong skill — worth the same staleness guard `openSkill` already uses (`selectedSkill.value?.name === skill.name` before writing).

src/renderer/src/pages/plugins/SkillsPluginsPage.vue
```diff
@@ -1,7 +1,488 @@
...
+      skillClient.getAllSkills(),
+      configClient.listAgents({ agentType: 'deepchat' })
+    ])
+    if (requestId !== loadRequestId) return
+
+    skills.value = nextSkills
+    agents.value = nextAgents.filter((agent) => (agent.agentType ?? agent.type) === 'deepchat')
+    if (selectedSkill.value) {
+      const refreshedSkill = nextSkills.find((skill) => skill.name === selectedSkill.value?.name)
+      if (refreshedSkill) selectedSkill.value = refreshedSkill
```
zhangmo8, line 92: When a background catalog refresh drops the currently-open skill, this force-closes the detail dialog and bypasses the dirty-draft confirm the dialog itself implements (`handleOpenChange` → `discardConfirmOpen`) — unsaved edits are discarded without a prompt. Edge case (external delete / sync import while editing), but since the component went to the trouble of a discard confirmation, this path contradicts it; keeping the dialog open in a "skill removed" state or routing through the same confirm would be more consistent.

src/renderer/src/pages/plugins/skills/ImportSkillsFromAgentDialog.vue
```diff
@@ -0,0 +1,404 @@
...
+  else if (failedStage.value === 'execute') void execute()
+}
+
+const retryFailedItems = () => {
+  const failedNames = new Set(result.value?.failed.map((failure) => failure.skillName) ?? [])
+  result.value = null
+  void loadPreview(failedNames)
+}
+
+const formatAgentNames = (agentIds: string[] | undefined) => (agentIds ?? []).join(', ')
```
zhangmo8: `formatAgentNames` joins raw `affectedAgentIds` into the user-facing `overwriteImpact` message, but custom DeepChat Agent ids look like `deepchat-Xk3v9Qw2` (`deepchat-${nanoid(8)}`). The sibling surfaces in this PR both map id → display name (`SkillsPluginsPage.enabledAgentNames`, `SkillDetailDialog.agentNameById`); this dialog should get the same mapping (it can receive the agents list like the detail dialog does).

src/renderer/src/i18n/da-DK/settings.json
```diff
@@ -1889,7 +1889,7 @@
       "successMessage": "Færdighed {name} er blevet slettet succesfuldt",
       "title": "Slet skill"
     },
-    "description": "Administrér og konfigurér AI-assistentens skill-modul",
+    "description": "Gennemse alle Skills. Åbn en Skill for at se den og administrere aktiverede Agents.",
```
zhangmo8, line 1892: The reworded `settings.skills.description` / `settings.skills.emptyHint` (all 16 non-English/non-Chinese locales) and `skills.agentImport.menuItem` / `.description` (11 of them) land as en-US text, replacing existing translations — dev had e.g. da-DK "Administrér og konfigurér AI-assistentens skill-modul", ja-JP "AI アシスタントのスキルモジュールの管理と設定". The source copy did change meaning, so keeping the stale translations wasn't an option either — but these are the Skills page header and the import menu entry, and combined with the 21 brand-new keys shipping as en-US placeholders it means the feature is effectively localized only for en-US and zh-CN. Related staleness: `skills.agentImport.title` was updated in en-US/zh-CN only, so ko/ms/pl/pt/ru still show the *old* English string "Import Skills from Agent". Is a translation sweep planned before or right after merge?

zhangmo8 (commented):
Went through the main-process catalog/migration rewrite, the runtime/tool integration, contracts + API clients, the new Plugins/Skills surface, and the i18n/docs/registry sweep.

The core design holds up well: the plan → stage → journal → commit migration with startup-fatal gating plus the route-level `ensureInitialized()` wrapper means no pre-migration write can clobber legacy v1/v2 state; the v1/v2 → v3 `assigned` derivation preserves legacy allow-list semantics; and the trusted-override chain (validated once in `TurnCoordinator`, then honored by prompt assembly, allowed-tools and FS authorization) is consistent end to end. Deleted routes/events/components leave no dangling references I could find, and the registry refreshes look like normal build output.

What I'd want addressed before merge, in rough priority order:

1. The migration journal is written non-atomically and parsed with zero tolerance — a torn write (ENOSPC / crash) permanently bricks startup (inline at `src/main/skill/index.ts`).
2. A watcher `delete` event now permanently wipes every Agent's binding for that skill — transient unlink→create sequences (rename-based atomic saves) silently lose assignments + env config (inline).
3. The guided-onboarding Skills step hands off from the settings window to the main window via `sessionStorage`, which doesn't cross windows in Electron — the step silently stalls (inline at `guidedOnboardingSettings.ts`).
4. `skillsRemoveActiveRoute` still records the deleted `settings-skills` route name; clicking that activity entry in Settings Overview throws in vue-router (inline at `src/main/skill/routes.ts`).
5. i18n: the reworded Skills-page copy lands as English in the 16 locales that previously had translations, and zh-HK/zh-TW miss six of the new keys (inline).

The remaining inline notes are smaller: a dialog race on `selectedSkill`, background refresh bypassing the dirty-draft confirm, raw Agent IDs in the import conflict hint, the dead `disabled-updated` event branch, dead `skills.sources.*` keys, a memory-heavy directory compare, and two questions (duplicate's fork-and-switch semantics, `.agent-scopes` cleanup).

src/renderer/src/i18n/zh-HK/settings.json
```diff
@@ -2020,46 +2020,30 @@
...
         "skip": "略過",
         "rename": "重新命名",
         "overwrite": "覆寫"
       },
-      "renameTarget": "將匯入為 {name}",
-      "executeError": "匯入 Skills 失敗",
-      "importing": "正在匯入...",
-      "importSelected": "匯入所選 {count} 個"
+      "overwriteImpact": "覆寫此 Skill 會影響已啟用的 Agent：{agents}。",
+      "resultSummaryV3": "已匯入 {imported} 個、重用 {reused} 個、略過 {skipped} 個、失敗 {failed} 個。",
```
zhangmo8, line 2045: zh-HK/zh-TW translated nearly all of the new keys, except `agentImport.resultSummaryV3` + `agentImport.execute` (L2045-2046 in both files) and the four `skills.sources.*` values — zh-CN has all of these translated, so this looks like an oversight rather than policy. Also `agentImport.title` still reads 「從其他 Agent 匯入 Skills」 in both, which is now misleading given Agent-to-Agent import was removed in this PR.

src/renderer/src/i18n/en-US/settings.json
```diff
@@ -2304,15 +2303,21 @@
   },
   "skills": {
     "title": "Skills",
-    "description": "Manage and configure AI assistant skills",
+    "description": "Browse all Skills. Open a Skill to preview it and manage enabled Agents.",
+    "sources": {
```
zhangmo8: `settings.skills.sources.{plugin,builtin,external,local}` are added to all 20 locales, but nothing in the PR tree references `skills.sources` (no dynamic `sources.${…}` construction either), and `docs/features/deepchat-skills-management/spec.md` explicitly says cards "never display source". Leftover from an earlier iteration? If they're not reserved for a concrete follow-up, drop them — same for the now-unreferenced `settings.skills.card.enabled/disabled` and `importExport.includeDisabled` keys that stayed behind.

PR #2151, opened 2026-08-13
feat(chat): dock plan/question floats to slim bar

zhangmo8:
## Summary

The plan float and the question/permission panel used to stack into one large combined overlay above the composer, which got noisy when both were active. This PR replaces the floating surfaces with a slim interaction dock:

- New `ChatInteractionDock` renders a 40px bar with a `Plan` chip and a `Question` chip; each chip expands its panel above the bar, and only one panel can be open at a time (accordion).
- A pending question with no plan still opens expanded by default.
- Plan panels now default to collapsed (`agentPlan` store default), so an active plan shows as a chip until expanded.
- When plan and question are triggered together, both start docked and the user picks which panel to expand; the question chip shows a pulse dot while docked so the blocked turn stays discoverable.
- The combined-panel glass styles moved from `ChatPage.vue` into the dock; the read-only session overlay is unchanged.

## BEFORE

```
┌──────────────────────────────────────────────┐
│ ┌──────────────────────────────────────────┐ │
│ │ ☑ Plan (2/5)                           ✕ │ │
│ │  ✓ step 1                                │ │
│ │  ● step 2                                │ │
│ ├──────────────────────────────────────────┤ │
│ │ ? Question                               │ │
│ │  Which option should we use?             │ │
│ │  [Option A] [Option B] [Other]           │ │
│ └──────────────────────────────────────────┘ │
│  (composer hidden while question pending)    │
└──────────────────────────────────────────────┘
```

## AFTER

Both active (default state — everything docked):

```
┌──────────────────────────────────────────────┐
│ ▸ Plan 2/5   ▸ Question •                    │ ← 40px dock bar
└──────────────────────────────────────────────┘
```

One chip expanded (panels swap, never stack):

```
┌──────────────────────────────────────────────┐
│ ┌──────────────────────────────────────────┐ │
│ │ ? Question                               │ │
│ │  Which option should we use?             │ │
│ │  [Option A] [Option B] [Other]           │ │
│ └──────────────────────────────────────────┘ │
│ ▸ Plan 2/5   ▾ Question                      │
└──────────────────────────────────────────────┘
```

## Testing

- `vitest run test/renderer/components/ChatPage.test.ts test/renderer/stores/agentPlanStore.test.ts` — 102 passed
- `vitest run test/renderer/components/AgentProgressFloat.test.ts test/renderer/components/message/MessageBlockBasics.test.ts` — 13 passed
- `oxfmt`, `oxlint` + guards, `vue-tsc --noEmit`, and `pnpm run i18n` all pass (run on Node 24.18)

## Summary by CodeRabbit

* **New Features**
  * Added a shared expandable dock for viewing agent plans and responding to pending questions or permissions.
  * Added reusable radio-button and checkbox groups with labels, descriptions, disabled states, and accessibility support.

* **Improvements**
  * New agent-plan panels start collapsed while preserving explicit preferences.
  * Plan and question panels are mutually exclusive and adapt automatically to activity.
  * Improved interaction overlays with clearer selection controls, custom responses, focus handling, responsive scrolling, and reduced-motion support.
  * Refined light and dark theme styling across chat interaction panels.


zerob13 (commented):
Found two user-visible issues in the interaction dock: the expanded controls are behind their trigger in keyboard order, and the glass decoration is anchored to the outer float layer instead of the panel.

src/renderer/src/components/chat/ChatInteractionDock.vue
```diff
@@ -0,0 +1,492 @@
...
+          v-else-if="interaction"
+          :embedded="true"
+          :interaction="interaction"
+          :processing="processing"
+          @respond="emit('respond', $event)"
+        />
+      </div>
+    </Transition>
+
+    <div
```
zerob13, line 5: [P2] Preserve a forward-tab path into the expanded panel

The panel is before this bar in DOM order. In the default plan + question state, a keyboard user tabs to the Question chip and activates it, but the newly rendered option buttons are inserted *before* the focused chip. Pressing Tab therefore skips the answer controls and leaves the dock; the composer is also hidden/inert while the interaction is pending. Render the bar before the panel and use CSS order to keep the panel visually above it, or explicitly move focus into the opened panel.

src/renderer/src/components/chat/ChatInteractionDock.vue
```diff
@@ -0,0 +1,492 @@
...
+    opacity: 0.7;
+  }
+  70%,
+  100% {
+    transform: scale(2.4);
+    opacity: 0;
+  }
+}
+
+.interaction-dock-panel {
```
zerob13, line 408: [P2] Make the panel the containing block for its decoration

The `::before`, `::after`, and backdrop layers are absolutely positioned with `inset`, but this panel remains `position: static`. At rest they therefore resolve against ChatPage's full-width absolute float layer; only the Transition's temporary transform makes the panel their containing block during animation. The clipped gradients and inset borders jump/misalign when the transition finishes. Add `position: relative` here (or the equivalent utility on the panel).

src/renderer/src/components/chat/ChatInteractionDock.vue
```diff
@@ -0,0 +1,492 @@
...
+    }
+  },
+  { immediate: true }
+)
+
+watch(hasPlan, (now, before) => {
+  if (now && !before && interactionKey.value) {
+    collapseBoth()
+  }
+})
```
yyhhyyyyyy, line 227: [P2] Expand the remaining interaction when the plan disappears. This watcher only handles a plan being added. If the user opens and dismisses the plan while a question is pending, `hasPlan` becomes false but `questionExpanded` stays false, leaving the blocking interaction collapsed.

src/renderer/src/components/chat/ChatInteractionDock.vue
```diff
@@ -0,0 +1,492 @@
...
+        data-testid="agent-interaction-dock-panel"
+      >
+        <div class="interaction-dock-panel__backdrop" aria-hidden="true" />
+        <AgentProgressFloat
+          v-if="expandedPanel === 'plan'"
+          :snapshot="planSnapshot"
+          :collapsed="false"
+          :embedded="true"
+          @dismiss="emit('dismiss-plan')"
+          @toggle-collapse="setPlanExpanded(false)"
```
zerob13: [P2] Restore focus after the embedded plan collapses

`AgentProgressFloat`'s header is itself a focusable disclosure trigger. This handler unmounts the whole outer panel, including that focused trigger, so keyboard focus falls back to the document instead of the dock's Plan chip. The standalone component previously hid only its body and kept its trigger mounted.

```text
Current:  [inner Plan trigger]* --Enter--> (panel removed) --> <body>*
Expected: [inner Plan trigger]* --Enter--> [Plan chip]*
```

Return focus to the Plan chip after the state update, or remove/disable the nested collapse trigger in embedded mode.

zerob13 (changes requested):
Requesting changes for the current dock behavior:

- Opening a docked panel does not provide a forward-Tab path into its controls.
- Collapsing the embedded plan from its inner header removes the focused element without restoring focus.
- Dismissing the plan can leave the sole blocking interaction collapsed.
- The panel decoration is positioned against the outer float layer instead of the panel.
- The ChatPage plan-store mock uses the opposite default from production, so the green component test does not protect these transitions.

The implementation and checks are otherwise clean, but these user-visible state, focus, and layout regressions should be fixed before merge.

PR #2156, opened 2026-08-14
fix(agent): harden tape context recovery

yyhhyyyyyy:
## Summary

Improve long-context reliability while preserving Tape’s append-only model. Context compaction now reduces the provider-visible View without requiring semantic summary generation to succeed.

## Changes

- Decouple reconstruction-boundary advancement from summary generation.
- Fall back to boundary-only anchors when summaries are unavailable.
- Rebuild and apply recovery only when the resulting View is strictly smaller.
- Limit recovery sequences while allowing recovery again after a successful provider response.
- Compact closed oversized tool-result units before semantic recovery.
- Preserve open tool units, runtime Skill authority, provider replay identity, and the latest closed unit when possible.
- Anchor context estimates to provider-reported usage and estimate only the exact-prefix suffix.
- Invalidate usage anchors when the provider, model, generation configuration, tool surface, or message prefix changes.
- Integrate recovery with per-request Tool Surface snapshots and capabilities.
- Document Tape context-compaction and recovery invariants.

## Safety

- Tape entries remain append-only and available for recall, replay, and audit.
- Cancellation cannot commit a reconstruction boundary.
- Active turns are never recovered by replaying the original user prompt.
- Tool call/result protocol pairing is preserved.
- Recovery retries require measurable progress and have a Run-level ceiling.
- Provider errors are redacted before logging or persistence.

## Validation

- 627 context-recovery and agent regression tests passed.
- 187 Tool Surface integration tests passed.
- `pnpm format`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm i18n`

## Summary by CodeRabbit

* **New Features**
  * Improved handling of long conversations with automatic context compaction and recovery.
  * Preserves system instructions, current requests, skills, and tool interactions during recovery.
  * Condenses large completed tool outputs while retaining relevant details and session history.
  * Supports recovery when summary generation is unavailable.
  * Limits repeated recovery attempts and retries with smaller requests when needed.

* **Bug Fixes**
  * Prevents ineffective compaction and improves restoration of active conversations after context limits are reached.

* **Documentation**
  * Added architecture and implementation guidance for context compaction and overflow recovery.


src/main/agent/deepchat/loop/contextCoordinator.ts
```diff
@@ -1170,6 +1385,9 @@ export class DeepChatContextCoordinator {
               ) {
                 retryDecision = 'context_recovery_exhausted'
                 contextRecoveryAction = 'fail'
+              } else if (providerToolResultCompactionAttempted) {
```
zerob13: [P1] Carry preflight tool compaction into provider recovery

`providerToolResultCompactionAttempted` is only set in `recoverProviderContextOverflow`, while the preflight `compactToolResults(...)` calls above do not record their phase. If older-only preflight compaction makes the local estimate fit but the provider still rejects the request, this flag remains false and the `preflightContextRecoveryAttempted` branch jumps directly to `strict_retry`. The newest closed result therefore never gets its required fallback compaction, and an unchanged strict projection fails even though a smaller protocol-valid projection exists. Record which preflight stage ran, or route the rejection back through `recoverProviderContextOverflow`.

src/main/agent/deepchat/runtime/toolOutputGuard.ts
```diff
@@ -85,6 +90,96 @@ type AgentOutputLimitsResolver = (
...
+
+    for (const [toolCallId, resultIndex] of unit.resultIndexes) {
+      const toolMessage = messages[resultIndex]
+      if (
+        protectedToolCallIds.has(toolCallId) ||
+        toolMessage.provider_replay ||
+        getProviderProjectionIdentities(toolMessage).length > 0 ||
+        typeof toolMessage.content !== 'string' ||
+        toolMessage.content.length <= TOOL_OUTPUT_VIEW_COMPACTION_THRESHOLD ||
+        toolMessage.content.startsWith(TOOL_OUTPUT_OFFLOAD_MARKER) ||
```
zerob13: [P2] Do not use untrusted tool text as compaction provenance

Tool result content is arbitrary tool-controlled text. A large inline result can legitimately or maliciously begin with either marker, so these prefix checks classify a raw 8K+ result as already offloaded/compacted and skip the only model-free active-turn reduction. The internally generated projections are normally already below the threshold; if idempotence still needs provenance, store it out of band (or validate the complete bounded stub) instead of trusting a prefix.

zerob13 (commented):
Found two correctness issues in the context-recovery path.

src/main/agent/deepchat/runtime/toolOutputGuard.ts
```diff
@@ -313,8 +405,7 @@ export class ToolOutputGuard {
       if (
         current.isError ||
         current.requiresInline ||
-        current.offloadPath ||
-        current.responseText.startsWith(TOOL_OUTPUT_OFFLOAD_MARKER)
+        current.offloadPath
```
zerob13: [P1] Apply this provenance check to resumed single results

The batch path now correctly treats `current.offloadPath` as trusted provenance, but the single deferred/resume path does not: `ResumeBudgetToolCall` carries the guard-created `offloadPath`, while `TurnCoordinator.fitResumeBudgetForToolCall` passes only `existingOffloadPath`. After removing the prefix check, a tight resume budget calls `offloadToolOutput` with the existing stub as `rawContent`; `resolveToolOffloadPath(sessionId, toolCallId)` resolves the same file, so the full result is overwritten by its stub (and cancellation cleanup can then remove it). Thread the guard-owned offload metadata through `fitExistingToolOutput` and skip re-offloading that projection.

zerob13 (commented):
The original two findings are fixed, but the provenance change exposes one deferred-resume regression.

PR #2162, opened 2026-08-16
feat(agent): harden tape context compaction

yyhhyyyyyy:
## Summary

- Enforce strictly shrinking compaction checkpoints and preserve summary-gap provenance.
- Pin the effective first user message without counting it as hidden context.
- Reconcile interrupted compaction markers across restarts, runtime replacement, and database reopen.
- Record provider pressure and compaction usage as append-only Tape evidence with narrow writers.
- Rebuild compaction cost projections from Tape while preserving partial provider usage.
- Exclude compaction-derived events from the default model View and recall search.
- Add revision-safe compaction state synchronization and context occupancy reporting.
- Surface compacting, compacted, and boundary-only states in the chat UI.
- Remove obsolete compaction UI and unversioned state-reading paths.

## Tape semantics

This keeps the Tape as the source of truth:

- Facts remain append-only.
- Compaction changes the default read set, not history.
- Summaries, usage records, and occupancy are derived evidence.
- Derived compaction events stay out of the default View/search.
- Checkpoints, pins, markers, and usage records retain provenance.
- Internal compaction facts use explicit narrow-writer enforcement.

## UI

BEFORE

```text
┌──────────────────────────────┐
│ Conversation                 │
│                              │
│ No compaction status         │
│ No context occupancy signal  │
└──────────────────────────────┘
```

AFTER

```text
┌──────────────────────────────┐
│ Conversation                 │
│ ── Context compacted ─────── │
│ Boundary-only reason, if any │
│ Context usage: 62%           │
└──────────────────────────────┘
```

Stale occupancy evidence is shown as stale rather than presenting an outdated percentage as current.

## Reliability

- Provider pressure observations never replay a completed response.
- Pressure recovery rebuilds the canonical View from Tape.
- Snapshot synchronization uses subscribe → snapshot → `emitSeq` merge with bounded retry.
- Invalid preload event payloads are isolated from subsequent events.
- Forked sessions do not inherit compaction UI markers.
- Invalid zero-length context windows are rejected at the settings boundary.

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run i18n`
- Main tests: 577 files, 7,938 passed
- Renderer tests: 253 files, 2,148 passed

## Summary by CodeRabbit

- **New Features**
  - Added context-window occupancy indicators with usage, capacity, measurement source, estimates, and stale status.
  - Added compaction snapshots and dashboard breakdowns for chat, compaction, and unknown usage.
  - Preserved the initial user instruction during compaction and recovery.

- **Bug Fixes**
  - Invalid context-length values are rejected or safely defaulted.
  - Improved recovery after restarts, database reopening, and interrupted compaction.
  - Added clearer messages when summaries are unavailable or oversized.

- **Changes**
  - Removed the auto-compaction settings controls.


zerob13:
## Manual review

I reviewed the whole diff against `dev`, read the plan/spec, ran the focused main-process suites
(`test/main/tape`, `test/main/session/data`, `test/main/agent/deepchat`: 86 files, 2101 passed,
132 skipped), and checked i18n key consistency across all 20 locales (complete, incl. the removed
`autoCompaction` keys).

### What this PR does

- Decouples durable boundary progress from summary success; enforces a commit-time shrink proof for
  semantic checkpoints (`next checkpoint < current checkpoint + newly hidden visible turns`);
  CAS-loss/abort remain fail-closed.
- Pins the effective first user message in the `cache-aware-v2` View (never counted as hidden
  context, never given the derivative-content fence).
- Reconciles interrupted compaction markers across restarts / instance replacement / DB reopen via
  `compactionAttemptId` + anchor lookup, with a partial-index recovery query.
- Records silent provider pressure on `provider/attempt_completed` (schema v3) and compaction model
  calls as narrow-writer Tape events, plus a rebuildable `deepchat_usage_stats` projection
  (v68 migration, `chat`/`compaction` categories).
- Adds a context-occupancy read model (latest valid `view/assembled` + final physical attempt,
  `current | stale | unavailable`, provider vs estimated) and a race-free renderer sync
  (subscribe → snapshot → `emitSeq` merge).
- Removes the dead auto-compaction settings UI and the unused `useContextLength` helper.

### Over-engineering verdict

Mostly proportionate: the append-only discipline, narrow writers, CAS + shrink proof, and the
`emitSeq` merge protocol are all paid for by the spec's race/recovery requirements. Four spots are
not justified — see inline comments on the dead `LEGACY_SCHEMA_VERSION = 1` branch
(`compactionUsage.ts:6`), the V1/V2 policy duplication (`tapeViewPolicy.ts:157`), the
`EXPLAIN QUERY PLAN` test copies (`deepchatTapeEntriesTable.test.ts:156`), and the
`createTable()` override side-loading a partial index outside the migration machinery
(`deepchatMessages.ts:74`).

### Issues not covered by inline comments

1. **Branch is stale — merge would silently revert two dev fixes.** The diff includes unrelated
   MiniMax-M3 and chat-navigation changes that this branch never made; see the follow-up comment below (GitHub cannot anchor an inline comment there because
   the file is not part of this PR's diff).
2. `resolveProjection` writes `instance.setCompactionState(persistedState)`
   (`compactionRuntimeCoordinator.ts:168`) without emitting `sessions.compaction.changed` and
   without advancing `emitSeq`. After DB reopen / runtime replacement, a live renderer keeps the
   old state until the next activation pull or real event — the merge protocol's premise
   (every state change has a sequence) is violated on this path.
3. Dashboard summary mixes categories while calendar doesn't: `getSummary()`
   (`deepchatUsageStats.ts:366`) sums tokens across `chat` + `compaction`, while
   `message_count`/`session_count` and `getDailyCalendarRows` are chat-only. The plan blesses the
   summary behavior, but the card and the calendar now can't be reconciled without the breakdown
   card; worth a UI note ("includes compaction") or aligning the calendar.
4. Backfill `processedCount` counts attempted writes, not applied ones
   (`usageStatsService.ts:211`): the v68 upsert is a no-op for existing `live` rows
   (`WHERE source != 'live' OR excluded.source = 'live'`), so a re-run reports N processed with
   0 rows changed.
5. `state?.summaryText` in `summaryUpdatedAtFromCompactionAnchor` (`transcript.ts:78`) is a dead
   branch — anchor state carries `summary`, never `summaryText`.
6. `schemaCatalog.ts` repair column for `usage_category` lacks the `CHECK (usage_category IN
   ('chat','compaction'))` constraint present in the real DDL, so the repair path can produce a
   weaker table than the migration path.
7. "Strictly shrinking compaction checkpoints" overstates the invariant: the formula guarantees
   the total View shrinks (current checkpoint + newly hidden turns), not that checkpoints shrink
   monotonically — rolling summaries can grow across rounds. The code matches the spec formula;
   the PR description should say "strictly smaller View" to avoid future confusion.
8. Renderer: `synchronizeCompactionState`'s failure path calls `applyBufferedEvents()` after the
   retry loop without re-checking `ownsSync()` (`session.ts:497-501`). Safe today only because
   nothing awaits between the check and the call — fragile.
9. `messageContentHashByMessageId` is typed optional but required by the pin paths
   (`viewManifest.ts:70` vs `:516`/`:637`); the thrown error then says "no longer matches its Tape
   source" instead of "missing source map". Either make it required or give the pin path a
   dedicated missing-map error.
10. `TAPE_VIEW_CONTEXT_BUILDER_VERSION` was renamed to `'cache-aware-v2'` while only re-exported by
    the deprecated `session/data/tapeViewManifest` module (`viewManifest.ts:52`) — a latent
    mis-default if a v1-policy caller ever consumes the constant again. Consider removing the
    export instead of updating it.

### Test quality

Strong regression protection in the durable spots: `compactionUsage.test.ts` (atomic/idempotent
writes, rollback, backfill rebuild), `providerAttempt.test.ts` (v1/v2/v3 compat, malformed pressure
degradation), `tapeViewManifest.test.ts` (hash integrity, pin provenance), pressure-detector
boundary table in `contextCoordinator.test.ts`, and the `sessionStore` sync-race tests.

Please remove or rework these implementation-coupled tests:

- `compactionService.test.ts:1345` — tautological self-comparison (pre-existing on `dev` at
  lines 969-970, not introduced here; worth cleaning up while touching this file).
- `compactionService.test.ts:741-797` — asserts the exact positional call signature of
  `buildHistoryTurns`; any added parameter breaks the test without a behavior change.
- `compactionService.test.ts:1817-1869` — tests private `summarizeBlocks`/`buildSummaryPrompt` via
  `as any`; the observable contracts (boundary_only outcome, prompt-injection allowlist) are
  already covered elsewhere through public paths.
- `deepChatAgentHarness.test.ts:2922-2946` — call-count test for reconciliation, asserts method
  invocations instead of the reconciled result.
- `deepChatAgentHarness.test.ts:684-707` — fake hardcodes `schemaVersion === 3` + pressure-kind
  predicate, duplicating the production predicate inside a test double; will drift silently.

Missing coverage: the pin-vs-resume-owner edge case (it would have caught the crash flagged
inline), and a positive-payload test for `sessions.compaction.changed` at the bridge level
(only the invalid-payload drop is tested).

Overall this is careful, spec-faithful work; the pin/rebuild crash and the stale base are the two
things I'd block on.

src/main/agent/deepchat/runtime/contextBuilder.ts
```diff
@@ -1678,10 +1735,73 @@ function buildCacheAwareMetadata(input: {
...
+      candidate.orderSeq < record.orderSeq ||
+      (candidate.orderSeq === record.orderSeq &&
+        Buffer.compare(Buffer.from(candidate.id), Buffer.from(record.id)) < 0)
+    ) {
+      record = candidate
+    }
+  }
+  if (!record) {
+    if (expected) {
+      throw new Error('Pinned initial user instruction is no longer present in the effective View.')
```
zerob13, line 1770: The `expected` pin and `excludedRecordId` filters are mutually exclusive when the resume owner IS the pinned first user: candidates with `id === excludedRecordId` are skipped, and every other candidate fails `candidate.id !== expected.messageId`. For a single-turn tool loop (the exact case this PR hardens), the first user is both the pin and the active-turn owner, so `buildPinnedFirstUser` can never select a record and throws here.

`spec.md` states the intended behavior: "A resume whose protected active owner is itself the first user does not add a second prefix copy; the active turn remains the protected authority for that edge case." The rebuild path passes `expected` (`deepChatLoopRunner.ts`: `runPinnedFirstUser: params.pinnedFirstUser ?? null`), so it throws instead of falling back to the active turn — turning pressure recovery into a hard request failure. The normal resume path (`turnCoordinator.ts`) doesn't pass `runPinnedFirstUser` and silently handles this case, which is why existing tests miss it.

Suggested: when `excludedRecordId === expected.messageId`, skip the pin (active turn stays authoritative) instead of throwing; add a regression test for first-turn pressure recovery.

src/main/agent/deepchat/runtime/compactionRuntimeCoordinator.ts
```diff
@@ -329,54 +368,83 @@ export class CompactionRuntimeCoordinator {
...
+            compactionMessageId,
+            compactionAttemptId: intent.compactionAttemptId
+          })
+      )
     } catch (error) {
-      this.assertCurrent(sessionId, expectedInstance)
       this.deps.messageStore.deleteMessage(compactionMessageId)
       this.deps.messageProjection.refresh(sessionId, compactionMessageId)
-      this.emit(sessionId, this.fromSummary(intent.previousState), expectedInstance)
+      this.emit(
```
zerob13: Commit 979aad113 intentionally removed `assertCurrent` before `deleteMessage` so a replaced instance can still retract its marker (per the plan: "Marker finalization/retraction proceeds after an originating instance is replaced"). But `emit()` below still calls `assertCurrent`, so when the instance was replaced mid-compaction the retraction succeeds and then emit throws, masking the original abort/provider error with an instance-fencing assertion. Either skip the emit when the instance is no longer current, or make `emit` tolerate replacement here.

src/main/session/data/tables/deepchatMessages.ts
```diff
@@ -52,9 +67,18 @@ export class DeepChatMessagesTable extends BaseTable {
         updated_at INTEGER NOT NULL
       );
       CREATE INDEX IF NOT EXISTS idx_deepchat_messages_session ON deepchat_messages(session_id, order_seq);
+      ${COMPACTION_RECOVERY_INDEX_SQL}
     `
   }
 
+  override createTable(): void {
```
zerob13, line 74: Overriding `createTable()` to side-load `COMPACTION_RECOVERY_INDEX_SQL` bypasses the versioned migration machinery (`getMigrationSQL` returns null for this table). Existing databases get this index through an unversioned side door that re-runs on every open, while every other schema change in this PR goes through `schema_versions`/repair paths. A proper migration version (or extending the catalog repair mechanism) keeps index provisioning auditable and consistent with the rest of the codebase.

src/main/tape/domain/compactionUsage.ts
```diff
@@ -0,0 +1,204 @@
+import type { DeepChatTapeEntryRow } from './entry'
+import { parseTapeJsonObject } from './effectiveSemantics'
+
+export const TAPE_COMPACTION_MODEL_CALL_EVENT_NAME = 'compaction/model_call_completed'
+export type TapeCompactionModelCallEventName = typeof TAPE_COMPACTION_MODEL_CALL_EVENT_NAME
+const TAPE_COMPACTION_MODEL_CALL_LEGACY_SCHEMA_VERSION = 1
```
zerob13: `TAPE_COMPACTION_MODEL_CALL_LEGACY_SCHEMA_VERSION = 1` plus the full v1 parse branch in `normalizeUsage` describe a historical format that can never exist: this is a brand-new event type introduced in this PR, and the only writer always emits v2. The v1 branch has no producer, can't be exercised by real data, and adds a compatibility surface that must be maintained forever. Delete it; a hypothetical v1 should fail closed to null like other malformed payloads.

docs/issues/tape-context-compaction/plan.md
```diff
@@ -84,6 +88,354 @@ review workflow; this plan records only the implementation and validation contra
...
+
+##### Silent Pressure State And Persistence Contract
+
+- The detector runs when one physical provider attempt settles, before its existing
+  `provider/attempt_completed` fact is appended. It uses that attempt's status, stop reason, and
+  usage plus the positive safe-integer effective context window captured from the same request
+  preflight. It emits `successful_prompt_overflow` only for `status=completed`,
+  `stopReason=complete`, and `inputTokens > contextWindowTokens`. It emits
+  `zero_output_length_at_limit` only for `status=completed`, `stopReason=max_tokens`,
+  `outputTokens=0`, and `inputTokens >= max(1, floor(contextWindowTokens * 0.99))`. Cache-read
```
zerob13: The implementation uses `Math.ceil(contextWindowTokens * 0.99)` (commit 6a694e4dc), but this line still documents `floor`. The code is the correct reading of the prose above ("input usage fills at least 99% of that window" — for window 101, floor gives 99 tokens = 98.02%, which is below 99%). Please update the formula here and in spec.md to `ceil` so the documented contract matches the detector; the `providerAttempt` domain validator accepts either form, so there is no data-compatibility concern.

test/main/session/data/tables/deepchatTapeEntriesTable.test.ts
```diff
@@ -95,6 +97,107 @@ describeIfSqlite('DeepChatTapeEntriesTable', () => {
...
+        'provider/attempt_completed',
+        'runtime_event',
+        'message-2',
+        2
+      )?.entry_id
+    ).toBe(finalAttempt.entry_id)
+
+    const viewPlan = db
+      .prepare(
+        `EXPLAIN QUERY PLAN SELECT *
```
zerob13, line 156: These `EXPLAIN QUERY PLAN` tests copy the production SQL verbatim and regex-match index names (`USING INDEX idx_deepchat_tape_entries_provider_context_pressure` etc.). They only prove "the SQL copied into the test uses the index the test names": if production SQL changes (even correctly) the test fails until the copy is updated; if the test copy is updated but production diverges, it passes. They're also fragile across SQLite versions. Index creation and migration coverage already lives in the mainDatabase tests — I'd drop the query-plan assertions here and in `deepchatMessagesTable.test.ts:211-281`, or replace them with behavioral assertions through the public reader methods.

src/main/agent/deepchat/runtime/tapeViewPolicy.ts
```diff
@@ -141,15 +146,61 @@ export const cacheAwareTapeViewPolicy: TapeViewPolicy = {
...
         historyRecords: input.historyRecords,
+        pinFirstUser: false,
         contextContributions:
           input.contextContributions ?? createEmptyContextRuntimeContributions()
       }
     )
   }
 }
 
+export const cacheAwareTapeViewPolicyV2: TapeViewPolicy = {
```
zerob13, line 157: `cacheAwareTapeViewPolicyV1` and `V2` are two full policy objects whose only differences are `id`, `contextBuilderVersion`, and `pinFirstUser: true|false`. That's a boolean flag duplicated into ~50 lines of policy surface; a third variant would copy it again. A single `cacheAwareTapeViewPolicy(pinFirstUser: boolean)` factory (or shared buildChat/buildResume wrappers parameterized by the flag) keeps both ids without duplicating the bodies.

zerob13:
Addendum to my review comment above:

1. **Stale base (blocking).** I could not anchor this inline because the affected files are not part of this PR's diff: the branch was cut from `dev` at `6726a44d` and misses two dev commits —
   `e9dc627ac fix(provider): omit MiniMax-M3 temperature (#2158)` and
   `fa769d4b0 fix(chat): restore parent session navigation (#2160)`. The diff against `dev`
   therefore shows #2158 and #2160 being reverted (MiniMax-M3 temperature omit, minimax-global
   adaptive thinking and its tests; parent-session navigation). This is merge-base drift, not an
   intentional change by this branch, but merging as-is would silently revert both fixes. Please
   rebase onto current `dev`.
2. **Tautological assertion is pre-existing.** The self-comparison in
   `test/main/agent/deepchat/runtime/compactionService.test.ts:1345-1347` already exists on `dev`
   (lines 969-970 there); this PR didn't introduce it, so my inline couldn't be anchored either.
   Since the file is already being touched heavily, removing it here is still cheap and worthwhile.

zerob13:
没啥了，解决完冲突合并吧

src/main/agent/deepchat/runtime/contextBuilder.ts
zerob13: Partially fixed — the build stage no longer throws, but the same first-turn pressure recovery still crashes one step later in `fit`.

Chain:
1. `recoverFromPressure` (contextCoordinator.ts:776-784) calls `rebuildAfterCompaction` and then unconditionally `fit(...)`.
2. After this fix, `buildPinnedFirstUser` returns `null` for the owner==pin case, so the rebuilt messages have no leading pin.
3. But the `fit` callback still passes `pinnedFirstUserContentHash: params.pinnedFirstUser?.contentHash` (deepChatLoopRunner.ts:2545-2554), and `fitCacheAwareMessagesToContextWindow` → `requirePinnedFirstUserMessage` throws `Pinned initial user instruction no longer matches the protected View prefix.` because the rebuilt View's leading user message is now the checkpoint, not the pin.
4. `recoverFromPressure` rethrows after restoring contributions, so first-turn tool-loop pressure recovery still fails hard.

I reproduced this with a throwaway test driving `buildTapeResumeView` (runPinnedFirstUser = owner, owner == only user) + the same merge logic as `rebuildAfterCompaction` + `fitRequestMessagesToContextWindow` with the run pin hash: rebuild succeeds without the pin, fit throws as above.

Suggested fix: let `rebuildAfterCompaction` propagate the rebuilt `metadata.pinnedFirstUser` (the assembler result already carries it) and pass its hash to `fit` — i.e. null when the pin was dropped — instead of always reusing the initial run pin hash. Also worth a regression test: first turn, tool loop, pressure recovery rebuild succeeds (pin absent, active turn authoritative) and the request fits without throwing.

zerob13:
Follow-up review after the fix commits (`4c088d9e0`..`f88e47ec7`). Verified against the refreshed branch; regression suites re-run: `contextCoordinator` / `compactionRuntimeCoordinator` / `compactionService` / `contextBuilder` / `harness` / `tape` — 13 files, 740 passed, 25 skipped.

### Fixed

- **Stale base** — `4c088d9e0` merges current `dev`; the #2158/#2160 reverts are gone from the diff (merge-base is now `fa769d4b0`).
- **Pin-vs-owner crash, build stage** — `a148f1570` makes owner==pin selectable and returns `null` (active turn stays authoritative), with a new regression test covering the non-duplication and the changed-hash throw. See the thread reply for the remaining `fit`-stage issue.
- **Catch-path emit masking errors** — `e126b02c7` gates `emit` on `scope.isCurrent()`; the replaced-instance tests now assert the original error surfaces and the replacement runtime is untouched.
- **ceil/floor spec drift** — `9c8004047` updates plan.md to `ceil`.
- **Dead v1 usage schema** — `f88e47ec7` removes the unpublished v1 branch; unknown schema versions now fail closed to `null`.

### Still open (from the earlier review, all non-blocking)

- `fit`-stage pin crash on first-turn pressure recovery (see thread on contextBuilder.ts:1770) — this one I'd still fix before merge since it's the same failure mode.
- `resolveProjection` mutating instance state without emitting / advancing `emitSeq` (compactionRuntimeCoordinator.ts:168) — live renderer stays stale after DB-reopen reconciliation.
- V1/V2 policy duplication (tapeViewPolicy.ts:157), `createTable()` override side-loading the partial index (deepchatMessages.ts:74), `EXPLAIN QUERY PLAN` test copies (deepchatTapeEntriesTable.test.ts:156 / deepchatMessagesTable.test.ts:211).
- Dashboard summary mixing chat+compaction tokens while the calendar is chat-only; backfill `processedCount` counting no-op upserts; `summaryText` dead branch; `usage_category` repair column missing the CHECK; `synchronizeCompactionState` failure path calling `applyBufferedEvents()` without an `ownsSync()` re-check; `messageContentHashByMessageId` optional-but-required in the pin path; `TAPE_VIEW_CONTEXT_BUILDER_VERSION` latent mis-default.
- Test cleanup: `buildHistoryTurns` positional-signature assertions, `as any` private-method tests, harness call-count test, fake hardcoding `schemaVersion === 3`.

None of those need to block; thanks for the fast turnaround on the blocking items.

PR #2167, opened 2026-08-17
fix(db): keep corrupt databases on startup

### LINKED ISSUE #1735: fix(db): Conversations lost after force-close — SQLite corruption detection and recovery
> **Source:** [Discussions #1226 — feedback from @c4-619-1](https://github.com/ThinkInAIXYZ/deepchat/discussions/1226)

## Description

When the application becomes unresponsive during a long-running task and the user force-closes it (kill process / Task Manager), some conversations disappear after restarting. The root cause is that the SQLite database may become corrupted during an abnormal exit, leading to data loss.

> 用户在对话过程中因应用卡顿强制关闭（kill 进程 / 任务管理器结束），重新打开后部分对话消失。原因是 SQLite 数据库在异常退出时可能损坏，导致数据丢失。

## Steps to Reproduce

1. Start a long-running conversation (e.g. heavy file operations, extended search).
2. Force-close the process when the app is clearly unresponsive (Ctrl+C / kill / Task Manager).
3. Restart DeepChat.
4. Check the conversation list.

> **复现步骤：**
> 1. 发起一个耗时较长的对话
> 2. 在应用卡顿时强制关闭进程
> 3. 重新启动 DeepChat
> 4. 观察对话列表

## Expected Behavior

Saved conversation data should be preserved even after an abnormal exit. If corruption occurs, the app should provide diagnosis and recovery mechanisms.

> **预期行为：** 即使异常退出，已保存的对话数据应尽可能保留。如果损坏，应提供诊断和修复机制。

## Actual Behavior

Some conversations are missing after restart. The user cannot see them in the client (though the conversation files may still exist on disk).

> **实际行为：** 部分对话在重启后消失，用户无法在客户端中看到。

## Root Cause Analysis

The current corruption handling (`DatabaseInitializer.ts` + `schemaRepair.ts`) only covers **schema-level** issues (missing tables, missing columns, type mismatches). It does **not** cover:

1. **WAL checkpoint interruption:** A force-close may prevent the WAL journal from being properly checkpointed into the main database file, causing data inconsistency.
2. **Page-level corruption:** A partial write during a crash can leave database pages incomplete.
3. **Index corruption:** Damaged B-tree index pages prevent queries from finding existing conversations.
4. **SQLITE_CORRUPT / SQLITE_NOTADB:** The current `isDestructiveDatabaseError` pattern matches these errors, but the handling strategy is to throw an exception rather than attempt recovery.

> **根因分析：**
> 当前的损坏处理仅覆盖 schema 级别的问题，未覆盖以下场景：
> 1. **WAL checkpoint 中断：** 强制关闭可能导致 WAL 日志未正确 checkpoint
> 2. **页级损坏：** 写入中途强杀导致部分页面写入不完整
> 3. **索引损坏：** B-tree 索引页损坏，查询无法找到已有对话
> 4. **SQLITE_CORRUPT：** 当前直接抛出异常，未尝试恢复

### Startup Detection & Repair / 启动时检测与修复

1. **Integrity check:** Run `PRAGMA integrity_check` on startup to detect page-level corruption.
   > 启动时执行 `PRAGMA integrity_check`，检测页级损坏。
2. **WAL recovery:** Ensure the WAL file is properly replayed on startup; run `PRAGMA wal_checkpoint(TRUNCATE)`.
   > 启动时确保 WAL 文件正确 replay，执行 `PRAGMA wal_checkpoint(TRUNCATE)`。
3. **Backup first:** Before attempting any repair, back up the current database file (`agent.db` → `agent.db.backup.{timestamp}`).
   > 修复前先备份当前数据库文件。

### Tiered Recovery Strategy / 分级恢复策略

1. **Light corruption (missing table/column):** Existing schema repair mechanism ✅
   > 轻度损坏（缺表/缺列）：已有 schema repair 机制
2. **Medium corruption (index damage):** Run `REINDEX`, then re-run `integrity_check`.
   > 中度损坏（索引损坏）：执行 `REINDEX`，再次 integrity_check
3. **Heavy corruption (page-level damage):** Use `.recover` mode to export recoverable data into a new database, or use `sqlite3 .dump` to salvage readable portions.
   > 重度损坏（页级损坏）：使用 `.recover` 模式导出可恢复数据，或用 `sqlite3 .dump` 导出能读取的部分

### User Communication / 用户告知

- If auto-repair fails, show a clear error message and manual repair instructions.
- Provide the database backup path so users can attempt recovery with third-party tools.
- Consider adding a "Database Repair" entry in Settings.
> - 自动修复失败时，向用户展示错误信息和手动修复指引
> - 提供数据库备份路径，方便使用第三方工具恢复
> - 考虑在设置中添加"数据库修复"入口

### Preventive Measures / 预防措施

- Evaluate enabling `PRAGMA synchronous=FULL` (trade performance for safety) or keeping `NORMAL` with periodic checkpoints.
- Use transactions for critical write operations to reduce the inconsistency window.
> - 评估 `PRAGMA synchronous=FULL` 的可行性
> - 对关键写操作使用事务，减少不一致窗口

yyhhyyyyyy:
## Summary

Force-close / `SQLITE_CORRUPT` no longer silently deletes `agent.db`.
Startup now quarantines the original files only after the user confirms
"Start empty" on splash.

Closes #1735

## What changed

- Remove backup → unlink → empty-db retry from `MainDatabase`
- Classify open failures: true corruption, unreadable / likely
  encrypted, leftover WAL without a main file
- Refuse to create a replacement DB when only `-wal` remains
- Splash recovery: Start empty (confirm) / password / Quit
- Keep encryption when a password is already proven; clear metadata
  only for the unreadable start-empty path
- Decrypted `SQLITE_CORRUPT` / malformed leaves the unlock loop and
  opens recovery; wrong-password `NOTADB` stays in the password loop
- Quarantine uses a precomputed exact path; a failed move returns to
  the same dialog instead of the English ErrorBox
- Headers shorter than 16 bytes classify as true corruption

## Out of scope

- Periodic WAL checkpoint / `synchronous=FULL`
- Startup `integrity_check`
- Structured recovery telemetry
- Copy-out salvage / REINDEX settings entry

## Summary by CodeRabbit

- **New Features**
  - Added startup recovery for corrupted, unreadable, or interrupted database files.
  - Added splash-screen options to retry passwords, start with an empty database, confirm recovery, or cancel.
  - Preserves recoverable encryption settings and safely quarantines damaged files.

- **Bug Fixes**
  - Prevents orphaned database journal files from causing unsafe startup behavior.
  - Improves handling of invalid passwords and database corruption errors.

- **Localization**
  - Added recovery preview labels across supported languages.


src/main/app/databaseStartup.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  while (true) {
+    if (!passwordResolved) {
+      try {
+        password = await input.security.resolveStartupPassword((request) =>
+          input.splash.requestDatabaseUnlock(request)
+        )
+        passwordResolved = true
+      } catch (error) {
+        const kind = classifyDatabaseStartupFailure({ error, dbPath, password })
```
zerob13, line 50: [P1] Preserve the proven password when decrypted corruption escapes unlock

When `resolveStartupPassword()` validates a wrapped or manually entered correct password and SQLite reports `SQLITE_CORRUPT` / `database disk image is malformed`, it rethrows before this assignment completes. The local `password` therefore remains `undefined`, so `classifyDatabaseStartupFailure()` reads the encrypted header and labels the file `unreadable` instead of `true-corruption`.

This loses the fact that decryption already succeeded. If the user then chooses Start empty, lines 119-120 clear the encryption metadata and the replacement DB is created unencrypted. Please propagate the proven password/corruption state across this boundary (for example with a typed result/error), set `passwordResolved`, and enter `true-corruption` recovery without asking for the password again. Add an integration-style test covering safeStorage/manual unlock -> decrypted corruption -> Start empty.

src/main/data/databaseStartupRecovery.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export function quarantineDatabaseFiles(dbPath: string, directory: string): string {
+  fs.mkdirSync(directory, { recursive: true })
+
+  for (const source of [sqliteShmPath(dbPath), sqliteWalPath(dbPath), dbPath]) {
+    if (!fs.existsSync(source)) {
+      continue
+    }
+
+    fs.renameSync(source, path.join(directory, path.basename(source)))
```
zerob13, line 108: [P1] Keep the database/WAL set intact when quarantine partially fails

The files are renamed one at a time with the WAL before the main DB. If the WAL rename succeeds and the final main-file rename throws (for example because of a transient lock or ACL), the caller re-displays recovery, but quitting or restarting leaves `agent.db` in place while its WAL exists only in the quarantine directory. `assertNoOrphanWalSidecar` does not catch this because the main file still exists, so SQLite can open a stale main file and ignore committed pages from the moved WAL.

Please make the failure state recoverable without leaving a stale-looking main DB. Moving the main file first is the smallest option: a main-file failure leaves the set untouched, while a later sidecar failure leaves no main file and is caught by the orphan-WAL guard. A rollback or copy-then-remove strategy also works. Add a failure-injection test for an error after the WAL has moved.

src/main/data/databaseConnection.ts
```diff
@@ -16,6 +17,7 @@ function ensureDatabaseDirectory(dbPath: string): void {
 
 export function openSQLiteDatabase(dbPath: string, password?: string): Database.Database {
   ensureDatabaseDirectory(dbPath)
+  assertNoOrphanWalSidecar(dbPath)
```
zerob13, line 20: [P2] Preserve the OCR cache rebuild path for orphan WAL files

`openSQLiteDatabase` is also used by `openPersistentBackend` in `ocrArtifactStore.ts`. If the derived OCR cache loses its main file but retains `-wal`, this new global guard throws `OrphanWalDatabaseError`. `shouldRebuildOcrCache` does not recognize `ORPHAN_WAL`, so the cache is not cleaned and rebuilt; `createBackend` instead falls back to memory for the rest of the process.

Either scope this guard to the main-database startup boundary, or explicitly treat `OrphanWalDatabaseError` as rebuildable in the OCR cache. Please cover the missing-cache-main plus existing-WAL case.

src/main/app/databaseStartup.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    try {
+      const initializer = new DatabaseInitializer({
+        password,
+        dbPath,
+        observe: input.observe
+      })
+      const database = await initializer.initialize()
+      await initializer.migrate()
```
zerob13: [P3] Remove the no-op migration seam

`DatabaseInitializer.initialize()` constructs `MainDatabase`, whose initialization already runs the actual migration. `DatabaseInitializer.migrate()` has this single production caller and only checks that a database exists before writing two log lines; its own comment says it exists for possible future needs.

Please remove this call, the `IDatabaseInitializer.migrate` surface, the no-op method, and its mocks until a genuinely separate migration phase exists. This removes roughly 28 lines without changing behavior.

zerob13 (commented):

PR #2168, opened 2026-08-17
fix(agent): declare exec shell dialect

### LINKED ISSUE #2125: [bug] Declare the actual shell dialect in Agent exec capabilities
## Summary

The Agent system prompt reports the operating-system platform, but the `exec` tool contract only says “shell command” and does not identify the actual shell dialect selected by the runtime. On Windows this encourages otherwise capable models to emit POSIX commands that fail and trigger avoidable retries or permission prompts.

**Priority:** P2 harness/tool UX

## Observed behavior

During a sanitized Windows Agent tool run, the model attempted Unix-style shell syntax before adapting to PowerShell. The resulting failures were not provider transport failures; they came from an ambiguous tool capability description.

## Code evidence

- `src/main/agent/deepchat/resources/systemEnvPromptBuilder.ts` emits a generic `Platform: win32` fact.
- The exec argument schema in `src/main/tool/agentTools/agentToolManager.ts` describes `command` as “The shell command to execute.”
- The tool definition says it executes a shell command but does not name PowerShell, `cmd.exe`, Bash, or the dynamically resolved shell.
- `src/main/tool/agentTools/agentBashHandler.ts` resolves and prepares the actual shell at execution time.

The nearest instruction to the model is the tool definition, and it lacks the dialect needed to construct a valid command.

- `docs/architecture/tool-system.md` makes the Tool catalog and source mapping authoritative for model-visible capabilities.
- `docs/architecture/deepchat-agent-harness-boundaries/spec.md` argues that execution safety/capability should be catalog-owned rather than inferred from tool names or conventions.
- Platform information in the system prompt is useful context, but it is not a substitute for a precise per-tool runtime contract.

## Impact

- Commands fail before doing useful work.
- The Agent spends extra provider rounds correcting syntax.
- Failed compound commands can generate additional approval requests.
- Users experience slower and less reliable tool execution.
- Hardcoding “PowerShell” globally would be wrong when a configured or packaged runtime uses another shell.

## Proposed direction

- Add a runtime-resolved shell capability to the model-visible exec tool description, for example shell name, platform, and a short dialect hint.
- Generate the description from the same shell resolver used by execution so prompt and runtime cannot drift.
- Keep the description content-free and do not expose user environment variables or executable search paths.
- Update the capability when the selected shell changes.

## Acceptance criteria

- [ ] On Windows, the model-visible exec contract identifies the actual PowerShell/cmd dialect selected by runtime.
- [ ] POSIX systems identify their actual shell contract without assuming Bash when it is not used.
- [ ] No user-specific paths, environment values, or command history are exposed.
- [ ] Tool definition and execution use one resolved source of truth.
- [ ] Tests cover Windows, POSIX, configured shell changes, and missing-shell fallback.

## User benefit

Models issue valid commands on the first attempt more often, reducing tool errors, retries, latency, and unnecessary permission interactions.

yyhhyyyyyy:
Closes #2125

  ## Problem

  When the model fills `exec.command`, the nearest contract it sees is the parameter schema — which still
  said a static `The shell command to execute`. The shell identity only lived in the tool-level description
  suffix and the system prompt, and those two were built by two independent formatters that had already
  drifted:

  - The tool description leaked internal enums (`Dialect: posix; path style: ...`), so fish — whose
  `dialect` field is `posix` for spawn/permission purposes — was actively presented to the model as a POSIX
  shell.
  - The system prompt claimed PowerShell 7 "does not support && or ||", which is wrong since pwsh 7.0
  (pipeline chain operators). Models were steered away from legal syntax.
  - The `$SHELL`-derived display name was sanitized in the system prompt but interpolated unsanitized into
  the tool description.

  On Windows this is the failure mode described in #2125: POSIX-first commands, a failed round, an extra
  approval.

  ## Changes

  - **`shared/commandShell.ts` becomes the single source for every model-visible shell description.** One
  per-profile dialect-hint switch feeds three views: the system prompt line, the `exec` tool description
  suffix, and a new dynamic `exec.command` parameter description. Display-name sanitization now applies to
  all of them. Internal `dialect`/`pathStyle` enums are no longer exposed; the MSYS path semantics stay in
  the Git Bash hint.
  - **`decorateExecForShell` now decorates the parameter layer too**:
  `parameters.properties.command.description` is replaced (not appended) with e.g. `The Windows PowerShell
  command to execute. It does not support && or ||; use ; for unconditional sequential execution.`
  Same-shell re-decoration is idempotent; definitions without a `command` property only get the description
  suffix.
  - **`systemEnvPromptBuilder` is now a pure caller** — its local switch and sanitizer are gone.
  - **Wording fixes**: PowerShell 7 supports `&&`/`||`; fish is flagged as non-POSIX with a one-line hint.

  Not touched, on purpose: `CommandShellService` resolution, `agentBashHandler`, the static zod schema
  (decoration overrides it per turn), the per-turn refresh (`resolveForTurn` → `configureToolMode` already
  re-decorates every turn), and the explicit-shell-unavailable error contract.

## Summary by CodeRabbit

* **Improvements**
  * Improved command execution guidance for PowerShell, CMD, Fish, Git Bash, and POSIX shells.
  * Shell prompts and command descriptions now provide clearer, profile-specific syntax and path guidance.
  * Shell names are displayed safely and consistently without exposing sensitive executable path details.
  * Execution tools now apply shell metadata consistently and update guidance when the selected shell changes.

* **Bug Fixes**
  * PowerShell Core checks now require version 7 or newer and reject invalid version information.

### DIFF
--- a/src/main/tool/codeMode/toolModeTools.ts
+++ b/src/main/tool/codeMode/toolModeTools.ts
@@ -1,5 +1,9 @@
 import { TOOL_EXECUTION, type MCPToolDefinition } from '@shared/types/mcp'
-import { formatCommandShellForModel, type ResolvedCommandShell } from '@shared/commandShell'
+import {
+  formatCommandShellForModel,
+  formatExecCommandDescription,
+  type ResolvedCommandShell
+} from '@shared/commandShell'
 import { CODE_MODE_TOOL_SERVER_NAME } from '@shared/codeModeProtocol'
 import { LIVE_DELEGATION_AGENT_TOOL_NAME } from '@shared/agentTools'
 import { UPDATE_PLAN_TOOL_NAME } from '@shared/types/agent-plan'
@@ -157,11 +161,27 @@ export function decorateExecForShell(
   shell: ResolvedCommandShell
 ): MCPToolDefinition {
   if (definition.function.name !== 'exec') return definition
+  const shellLine = formatCommandShellForModel(shell)
+  if (definition.function.description.endsWith(shellLine)) return definition
+  const { parameters } = definition.function
+  const command = parameters.properties.command
   return {
     ...definition,
     function: {
       ...definition.function,
-      description: `${definition.function.description}\n\n${formatCommandShellForModel(shell)}`
+      description: `${definition.function.description}\n\n${shellLine}`,
+      parameters: command
+        ? {
+            ...parameters,
+            properties: {
+              ...parameters.properties,
+              command: {
+                ...(command as Record<string, unknown>),
+                description: formatExecCommandDescription(shell)
+              }
+            }
+          }
+        : parameters
     }
   }
 }
--- a/src/main/tool/codeMode/toolModeTools.ts
+++ b/src/main/tool/codeMode/toolModeTools.ts
@@ -2,6 +2,7 @@ import { TOOL_EXECUTION, type MCPToolDefinition } from '@shared/types/mcp'
 import {
   formatCommandShellForModel,
   formatExecCommandDescription,
+  stripCommandShellLine,
   type ResolvedCommandShell
 } from '@shared/commandShell'
 import { CODE_MODE_TOOL_SERVER_NAME } from '@shared/codeModeProtocol'
@@ -161,24 +162,27 @@ export function decorateExecForShell(
   shell: ResolvedCommandShell
 ): MCPToolDefinition {
   if (definition.function.name !== 'exec') return definition
-  const shellLine = formatCommandShellForModel(shell)
-  if (definition.function.description.endsWith(shellLine)) return definition
+  const description = `${stripCommandShellLine(definition.function.description)}\n\n${formatCommandShellForModel(shell)}`
   const { parameters } = definition.function
-  const command = parameters.properties.command
+  const command = parameters.properties.command as Record<string, unknown> | undefined
+  const commandDescription = formatExecCommandDescription(shell)
+  if (
+    description === definition.function.description &&
+    (!command || command.description === commandDescription)
+  ) {
+    return definition
+  }
   return {
     ...definition,
     function: {
       ...definition.function,
-      description: `${definition.function.description}\n\n${shellLine}`,
+      description,
       parameters: command
         ? {
             ...parameters,
             properties: {
               ...parameters.properties,
-              command: {
-                ...(command as Record<string, unknown>),
-                description: formatExecCommandDescription(shell)
-              }
+              command: { ...command, description: commandDescription }
             }
           }
         : parameters
--- a/src/main/tool/codeMode/toolModeTools.ts
+++ b/src/main/tool/codeMode/toolModeTools.ts
@@ -2,7 +2,6 @@ import { TOOL_EXECUTION, type MCPToolDefinition } from '@shared/types/mcp'
 import {
   formatCommandShellForModel,
   formatExecCommandDescription,
-  stripCommandShellLine,
   type ResolvedCommandShell
 } from '@shared/commandShell'
 import { CODE_MODE_TOOL_SERVER_NAME } from '@shared/codeModeProtocol'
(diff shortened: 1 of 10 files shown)


src/main/tool/codeMode/toolModeTools.ts
```diff
@@ -157,11 +162,30 @@ export function decorateExecForShell(
   shell: ResolvedCommandShell
 ): MCPToolDefinition {
   if (definition.function.name !== 'exec') return definition
+  const description = `${stripCommandShellLine(definition.function.description)}\n\n${formatCommandShellForModel(shell)}`
+  const { parameters } = definition.function
+  const command = parameters.properties.command as Record<string, unknown> | undefined
```
zerob13, line 173: [P1] Scope shell decoration to the built-in exec tool

`configureToolMode()` receives the merged catalog, not an agent-only catalog. `collectToolDefinitions()` admits MCP definitions first and drops the built-in agent definition on a name collision, while `exec` is not reserved. Therefore an MCP server that exposes `exec({ command: ... })` reaches this new branch and has its own command description replaced with guidance for the local DeepChat shell, even if it executes remotely or uses a different DSL. This corrupts the model-visible MCP contract.

Please gate this decorator on the built-in identity (`source === 'agent'` and `server.name === 'agent-filesystem'`) before reading or rewriting `command`, and add an MCP `exec` regression case.

src/shared/commandShell.ts
```diff
@@ -167,19 +167,66 @@ export function normalizeAgentCommandShellConfig(value: unknown): AgentCommandSh
...
+    case 'windows-powershell':
+      return 'It does not support && or ||; use ; for unconditional sequential execution.'
+    case 'powershell-core':
+    case 'cmd':
+      return 'It supports && and ||.'
+    case 'git-bash':
+      return 'Use POSIX syntax. Use Windows-native paths with file tools; MSYS drive paths such as /c/... are for shell commands.'
+    case 'fish':
+      return 'Fish is not POSIX; bash idioms such as export do not work.'
+    case 'posix':
```
zerob13, line 204: [P1] Preserve the Fish hint for the default Auto path

The hint is keyed only by `profile`, but Auto on non-Windows always resolves through `resolveProfile('posix')`. The existing `CommandShellService` test at lines 600-619 explicitly proves that `/opt/homebrew/bin/fish` becomes `{ profile: 'posix', displayName: 'fish' }`. That common default configuration therefore reaches `case 'posix'` here and loses the new non-POSIX Fish guidance. The new Fish fixture only covers an explicitly selected `profile: 'fish'`.

Please derive the hint from the resolved shell identity, or normalize known Auto shell basenames to their specific profiles, and add coverage that composes the Auto-Fish resolver output with these formatters.

src/shared/commandShell.ts
```diff
@@ -167,19 +167,66 @@ export function normalizeAgentCommandShellConfig(value: unknown): AgentCommandSh
...
+
 function executableBasename(executable: string): string {
   return executable.split(/[\\/]/).filter(Boolean).at(-1) ?? executable
 }
 
+function commandShellDialectHint(profile: CommandShellProfile): string {
+  switch (profile) {
+    case 'windows-powershell':
+      return 'It does not support && or ||; use ; for unconditional sequential execution.'
+    case 'powershell-core':
```
zerob13, line 197: [P2] Do not promise pipeline-chain operators until the resolver enforces PowerShell 7

`assertPowerShellCoreAvailable()` currently accepts any probe output starting with `deepchat-pwsh:`, so a `pwsh` 6.x installation is classified as this profile. This unconditional hint then tells the model that `&&` and `||` are supported, although Microsoft documents those pipeline-chain operators as beginning in PowerShell 7.

Please parse the probed major version and reject versions below 7, or carry the resolved version into the formatting decision. Add a 6.x probe regression case.

src/main/tool/codeMode/toolModeTools.ts
```diff
@@ -157,11 +162,30 @@ export function decorateExecForShell(
   shell: ResolvedCommandShell
 ): MCPToolDefinition {
   if (definition.function.name !== 'exec') return definition
+  const description = `${stripCommandShellLine(definition.function.description)}\n\n${formatCommandShellForModel(shell)}`
```
zerob13: [P3] Remove the cross-shell re-decoration state machine

A repository-wide production call-site check does not find a lifecycle that feeds a decorated definition back into this function. `deepChatLoopRunner` and `turnCoordinator` pass definitions freshly returned by the raw tool catalog into `configureToolMode`; the SDK path separately decorates raw `executionTools`. `configureToolMode` is copy-on-write and stores its decorated result separately. Only the new tests pass this function output back into itself.

Please format once at the catalog boundary and remove `stripCommandShellLine`, the trailing regex, and the cross-shell re-decoration tests. Besides deleting roughly 30 lines, this avoids treating arbitrary trailing `Selected shell:` text as generated state.

zerob13 (commented):

PR #2171, opened 2026-08-17
fix(agent): surface guard-stop terminals

### LINKED ISSUE #2122: [bug] Surface guard-triggered Agent stops as actionable terminal states
## Summary

Safety guards such as the tool-call limit and repeated no-progress detection can stop an Agent correctly, but the resulting user/CLI state can look like an ordinary completion or an unexplained stop.

The guards should remain. Their terminal outcomes need explicit, actionable product semantics.

**Priority:** P2

## Current behavior

- `src/main/agent/deepchat/loop/deepChatLoopEngine.ts` enforces `MAX_TOOL_CALLS = 128`.
- `src/main/agent/deepchat/runtime/noProgressToolLoopGuard.ts` adds a correction after two identical tool batches and terminates after four.
- In `src/main/agent/deepchat/runtime/process.ts`, the max-tool path stamps `runOutcome = 'completed'` with `runStopReason = 'max_tool_calls'` and maps the plan terminal reason to `max_steps`.

The structured stop reason exists, but an ordinary “completed” outcome can dominate the renderer/CLI presentation. A user may conclude that the Agent chose to finish, even though a safety budget ended the run before the requested task was complete.

- `docs/architecture/durable-execution-journal/spec.md` requires authoritative terminal facts before terminal transcript/status/UI projection.
- `docs/architecture/local-control-plane/spec.md` requires terminal benchmark records to expose finish reason, retries, cancellation outcome, and stable machine semantics.
- `docs/architecture/deepchat-agent-harness-boundaries/spec.md` requires terminal persistence and public return values to remain explicit across lifecycle owners.

## Impact

- Users experience a protective stop as random Agent behavior.
- Automation may treat a partial task as successful completion.
- The user has no direct “continue with a fresh budget” action.
- Support cannot distinguish model choice from runtime guard activation without inspecting metadata.

## Proposed direction

- Define a user-visible terminal category such as `limit_reached`/`needs_continuation` while preserving the exact stop reason.
- Show concise reason-specific copy:
  - tool-call limit reached;
  - repeated identical tool batches/no progress;
  - provider-round limit reached.
- Offer safe actions such as **Continue**, **Change approach**, or **Review details** where applicable.
- Ensure CLI terminal results are non-success or explicitly partial according to a documented contract; do not overload ordinary `completed`.
- Keep no-progress and tool-call protections fail-safe.

## Acceptance criteria

- [ ] Guard-triggered stops cannot render as an indistinguishable ordinary completion.
- [ ] Renderer and CLI expose the exact structured stop reason.
- [ ] Continuing creates a new physical run and does not mutate the historical terminal fact.
- [ ] No-progress copy explains that repeated tool results caused the stop without exposing sensitive tool output.
- [ ] Existing safety thresholds and loop protections remain intact.
- [ ] Tests cover max tools, no progress, max provider rounds, user stop, and normal completion.

## User benefit

Users understand why the Agent stopped and can take a safe next action instead of assuming the model silently abandoned the task.

yyhhyyyyyy:
## Summary

Safety guards already stop an Agent correctly, but the product treated those stops as a normal finish or an unexplained failure. This change keeps the guards and journal outcomes as they are, and projects the existing `runStopReason` to the renderer and CLI.

- `max_tool_calls` still stamps `completed` so claimed queue input is
  consumed instead of rolled back.
- `no_progress` and `max_turns` stay `error`, but the UI no longer shows
  generic "request failed" plus a raw English string.
- Continue starts a new physical run through `sendMessage`. It does not
  retry or rewrite the historical `run_terminal`.
- CLI `PublicRunSnapshot` and `events.subscribe` gain an optional open
  `stopReason`. `run watch` prints it and exits `6` for guard stops.
- `LOCAL_CONTROL_SURFACE_VERSION` is not bumped. Same-release CLI/app
  binaries must ship together because the snapshot schema is `.strict()`.

Closes #2122

## UI

Before:

```
[assistant answer / tools]
Request failed...
  (expand) Agent stopped after four identical...
```

`max_tool_calls` looked like a normal completed turn.

After:

```
[assistant answer / tools]
The tool-call limit was reached before the task finished.
[Continue]          <- only on the latest assistant
```

Historical guard banners keep the reason and hide Continue.

## CLI

- `runs.get` / recovery `runs.snapshot`: optional `stopReason` from the
  latest assistant metadata, independent of the message page.
- `run watch`: `Run <id> stream ended at <cursor> (idle, max_tool_calls)`
- `run start` and `runs.get` stay exit `0`
- `run watch` exits `6` (`domain`) for `max_tool_calls` / `no_progress` /
  `max_turns`

JSONL still prints `ok: true` for a successful RPC. Exit `6` is a domain
result, not a transport failure. Read `stopReason` and the exit code
together.

If the Session is deleted after the stream ends, subscribe still returns
`runId` and `lastCursor` instead of `not_found`.

## Compatibility

- No new `ExecutionRunOutcome`
- No new `action_type`
- Protective thresholds unchanged (128 tools, 2/4 no-progress)
- Legacy `needContinue` actions, if any historical rows exist, now send a
  follow-up turn instead of a destructive retry

## Test plan

- [x] Renderer banner + Continue sends a new message
- [x] Continue is single-flight
- [x] Continue button only on the latest assistant
- [x] CLI snapshot stopReason ignores page cursor
- [x] Watch prints the reason and exits 6
- [x] Claimed queue consume stays true for completed tool-limit stops
- [x] Later run commit cannot rewrite a historical `max_tool_calls` terminal

## Summary by CodeRabbit

* **New Features**
  * Added guard-stop banners for tool-call, progress, and turn-limit stops.
  * Added localized Continue actions to resume from the latest stopped assistant message.
  * Added guard-stop translations across supported languages.

* **Bug Fixes**
  * Prevented continuation actions from appearing on older or unavailable messages.
  * Preserved historical terminal states and safely handled invalid stop metadata.

* **Tests**
  * Added coverage for guard-stop rendering, continuation behavior, duplicate requests, and terminal-state persistence.


src/shared/contracts/routes/runs.routes.ts
```diff
@@ -62,7 +64,8 @@ export const PublicRunSnapshotSchema = z
     updatedAt: TimestampMsSchema,
     messages: z.array(PublicRunMessageSchema).max(RUN_MAX_MESSAGE_PAGE_SIZE),
     nextCursor: MessagePageCursorSchema.nullable(),
-    hasMore: z.boolean()
+    hasMore: z.boolean(),
+    stopReason: PublicRunStopReasonSchema.optional()
```
zerob13: [P1] Negotiate this strict output contract change

`PublicRunSnapshotSchema` and `events.subscribe` are strict V2 contracts. Because `LOCAL_CONTROL_SURFACE_VERSION` remains 2, an older CLI passes descriptor/version negotiation, then rejects these new keys in `runCli` via `parsed.contract.output.safeParse(...)` and exits 8. This contradicts the accepted local-control contract, which freezes V2 shapes and explicitly keeps old clients on byte/shape-compatible V2 (`docs/architecture/local-control-plane/spec.md`, contract policy and V3 compatibility sections).

Please either preserve the V2 outputs or expose these fields through a newly negotiated public surface so an old client fails early with the documented version-mismatch exit code. Shipping the normal bundled pair together does not enforce this protocol invariant. The same issue applies to the new `status` and `stopReason` fields in `events.subscribe`.

src/renderer/src/features/chat-page/composables/useDisplayMessages.ts
```diff
@@ -139,7 +139,12 @@ export function useDisplayMessages(options: UseDisplayMessagesOptions) {
             ...baseMessage,
             ...(streamingRenderKey ? { renderKey: streamingRenderKey } : {}),
             role: 'assistant',
-            content: filterRenderableAssistantBlocks(messageStore.getAssistantMessageBlocks(record))
+            content: filterRenderableAssistantBlocks(
+              messageStore.getAssistantMessageBlocks(record)
+            ),
+            ...(metadata.runStopReason?.trim()
```
zerob13: [P2] Validate persisted stopReason before calling string methods

`getMessageMetadata()` only checks that the parsed JSON root is an object; it does not validate individual field types. A persisted or imported row such as `{"runStopReason":123}` therefore reaches this line and `.trim()` throws, aborting `displayMessages` for the entire session.

Please normalize this once with the already-added `readRunStopReason(metadata)` helper (or an explicit string guard), then spread the normalized value. Add a malformed-metadata regression case because this value crosses a persistence boundary.

src/main/agent/deepchat/runtime/claimedInputSettlement.ts
```diff
@@ -0,0 +1,10 @@
+import type { ProcessResult } from './types'
+
+export function shouldConsumeClaimedInput(
```
zerob13: [P3] Remove the test-only settlement abstraction

`shouldConsumeClaimedInput` has one production caller and only re-houses the unchanged four-clause condition that was already inline there. The new test never drives a claimed input or a `max_tool_calls` terminal; it only passes the generic status `completed`, so it remains green even if `TurnCoordinator` stops using this helper and does not protect the claimed-input regression.

Please restore the inline condition and remove this module/unit test. If this invariant needs durable regression coverage, assert the observed claim disposition at the existing coordinator or harness boundary instead.

src/renderer/src/components/chat/MessageList.vue
```diff
@@ -106,6 +110,11 @@ const shouldDisableMarkdownVirtualization = computed(
   () => props.disableMarkdownVirtualization || isCapturingValue.value
 )
 const allRenderedMessages = computed(() => props.messages)
+const latestAssistantMessageId = computed(() =>
```
zerob13: [P3] Drop the virtual-window fallback and the tri-state prop

`MessageList` has one production caller, and that caller must provide the full-transcript assistant ID because `messages` is a virtualized window. The optional `undefined | null | id` prop plus fallback scan exists only to make isolated mounts infer a value, while retaining the exact stale-window branch this PR had to fix when a caller omits the prop.

Make this required nullable or default-deny, compare the supplied `null | id` directly, and remove the fallback-only regression case. The full-transcript owner should remain the single source of truth.

zerob13 (commented):

docs/issues/guard-stop-terminal-states/spec.md
```diff
@@ -0,0 +1,66 @@
...
+2. Do not add a new `action_type`. Renderer banners project `runStopReason` from persisted message
+   metadata. No second copy is written into transcript blocks.
+3. Continue starts a new physical run through `sendMessage` with non-empty visible user text. It
+   must not call `retryMessage` or mutate the historical `run_terminal`.
+4. Public local-control V2 outputs stay shape-compatible. `PublicRunSnapshot` and
+   `events.subscribe` do not gain fields. `LOCAL_CONTROL_SURFACE_VERSION` stays 2.
+
+## Contract Surfaces
+
+Renderer projection of `runStopReason` is the user-visible contract. CLI V2 run routes keep their
```
zerob13, line 50: [P1] Keep #2122 open unless the CLI contract is delivered

Preserving the V2 shape fixes the compatibility break, but this update also removes every CLI projection added by the PR: `runs.get` exposes no exact reason, `run watch` prints no reason, and a guard stop still exits 0. That is the original bug for all automation users, and #2122 explicitly requires both renderer and CLI semantics. The PR body still claims the removed fields/exit-6 behavior and uses `Closes #2122`, so merging this would close the issue while one of its acceptance criteria remains unimplemented.

Please either add a newly negotiated surface/capability while continuing to serve the frozen V2 shape, or narrow the PR description, remove `Closes #2122`, and track the CLI work in an open follow-up issue. Keeping V2 unchanged is a compatibility invariant, not completion of the CLI requirement.

zerob13 (commented):
Follow-up review after the compatibility update.

PR #2179, opened 2026-08-17
fix(agent): stop silent empty file reads

### LINKED ISSUE #2173: [BUG] The read tool is unable to read certain contents.
### System details / 系统信息

AMD Ryzen 5 5600G with Radeon Graphics | win11 | DeepChat v1.1.0

### What's wrong? / 出了什么问题？

The "read" tool cannot read the file content when the chat box is not referenced. It will return an empty result.I confirm that the file is in the corresponding workspace.
Image
Image
When there is a citation, the file content can be returned.
Image
There is also a minor UI issue here where the left side is cut off.
Image

yyhhyyyyyy:
## Summary

`read` returned an empty `path:\n\n` body for workspace files such as `Cargo.toml` and `.rs`. The chat-box citation was a red herring: attachments already omit content and tell the model to `read`. The split was MIME type, not citation.

PR #1319 routed any MIME outside a small text allowlist through `prepareFileCompletely`. Unknown `application/*` types (`application/toml`, `application/rls-services+xml`, `application/sql`, …) hit `UnsupportFileAdapter` and collapsed to `''`.

`read` now treats text as the default. Images, office/PDF/RTF extraction, and known binaries stay special-cased. Failures return an explicit message.

Closes #2173

## Behavior

- Raw-read source and config files that mime-db maps to `application/*` (toml, Rust, SQL, PHP, …).
- Keep CSV/TSV on the raw path so `read` returns the full file, not a 10-row preview.
- Office/PDF/RTF still go through document adapters. Files above `maxFileSize` are rejected before extraction. Empty extraction reports the real file size and configured limit.
- Raw reads load at most the first 10MB from the same open fd (loop until the window is full or EOF). Larger files keep a `[first N of M bytes]` header so the model can continue.
- NUL-only files get an explicit binary or encoding hint instead of an empty body.

## Test plan

- [x] `vitest` `binaryReadGuard` / `agentToolManagerRead` / `agentFileSystemHandler`
- [x] `pnpm run typecheck:node`
- [x] `oxlint` on the touched files

## Summary by CodeRabbit

* **New Features**
  * Improved reading of documents, source files, and configuration files.
  * Added support for BOM-less UTF-16 content and more reliable text detection.
  * Large files are read in bounded sections with pagination and truncation guidance.
  * Document reads enforce configurable size limits and provide clearer guidance for oversized or empty files.

* **Bug Fixes**
  * Binary file errors no longer affect other files in the same read operation.
  * Improved handling of empty files, archives, and files containing embedded null characters.

### DIFF
--- a/src/main/file/mime.ts
+++ b/src/main/file/mime.ts
@@ -139,3 +139,22 @@ export const getMimeTypeAdapterMap = (): Map<string, FileAdapterConstructor> => 
 
   return map
 }
+
+const DOCUMENT_READ_ADAPTERS: ReadonlySet<FileAdapterConstructor> = new Set([
+  PdfFileAdapter,
+  DocFileAdapter,
+  ExcelFileAdapter,
+  PptFileAdapter,
+  OpenDocumentFileAdapter,
+  RtfFileAdapter
+])
+
+export const DOCUMENT_READ_MIMES: ReadonlySet<string> = new Set(
+  [...getMimeTypeAdapterMap().entries()]
+    .filter(([key, adapter]) => DOCUMENT_READ_ADAPTERS.has(adapter) && !key.includes('*'))
+    .map(([key]) => key)
+)
+
+export function isDocumentReadMime(mimeType: string): boolean {
+  return DOCUMENT_READ_MIMES.has(mimeType)
+}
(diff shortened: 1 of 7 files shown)


yyhhyyyyyy:
@coderabbitai review

PR #2182, opened 2026-08-18
fix(ai-sdk): honor model timeout instead of undici 300s headersTimeout

Steve0x2a:
Local Ollama (and any other slow first-token provider) was dying at about five minutes with `UND_ERR_HEADERS_TIMEOUT` / `AI_APICallError: Headers Timeout Error`, even when the model timeout was set to one hour.

Node's default undici Agent uses `headersTimeout` 300s, independent of the `AbortSignal` from `combineRequestSignal`. `src/main/platform/proxy.ts` only called `setGlobalDispatcher` when a proxy URL was set, so SYSTEM mode with no proxy (typical local Ollama) kept that 300s default.

This installs the process-wide dispatcher with `headersTimeout: 0` and `bodyTimeout: 0` in `clearProxy`, custom/system proxy setup, and SYSTEM mode with no detected proxy. `EnvHttpProxyAgent` still gets the global `NO_PROXY` list. The AI SDK factory no longer attaches a per-request Agent/ProxyAgent, so fetch uses that dispatcher. Retry policy is unchanged.

Credit: @jianpuwuwei

Fixes #2180

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for AI provider requests by preventing premature timeout failures.
  * Improved network routing across direct connections, system proxy settings, and custom proxy configurations.
  * Prevented unnecessary routing configuration from being applied to direct provider requests.
* **Tests**
  * Added coverage for timeout behavior, proxy selection, and direct provider requests.

PR #2185, opened 2026-08-18
fix(proxy): inherit global fetch timeouts on remaining fetches

yyhhyyyyyy:
## Summary
- Follow-up to #2182. Remaining provider fetches still created a private `ProxyAgent`, which overrode the process dispatcher and restored undici's 300s `headersTimeout` / `bodyTimeout` when a proxy was set.
- Codex, Grok, TTS, video generation, model listing, and VoiceAI now use the global dispatcher so they share `headersTimeout: 0`, `bodyTimeout: 0`, and `NO_PROXY`.
- `resolveProxy()` failures also install that dispatcher. If dispatcher setup throws, the serialized resolve queue still settles and later proxy updates keep running; a previously working proxy URL is not dropped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system proxy handling, including `NO_PROXY` support and recovery when proxy resolution fails.
  * Invalid or unavailable proxy addresses now fall back safely to direct connections.
  * Preserved working proxy connections when later resolution attempts fail.
  * Voice availability checks now time out reliably when requests hang.

* **Changes**
  * AI, voice, video, and Codex requests now use standard network connections without provider-specific proxy dispatchers.

* **Tests**
  * Added coverage for proxy fallback, recovery, environment merging, timeouts, and direct request behavior.

### DIFF
--- a/src/main/platform/proxy.ts
+++ b/src/main/platform/proxy.ts
@@ -105,6 +105,7 @@ export class ProxyConfig {
       } else {
         setGlobalDispatcher(createGlobalFetchDispatcher())
         this.proxyUrl = null
+        this.clearProxyEnv()
       }
       return true
     } catch (error) {
@@ -135,9 +136,7 @@ export class ProxyConfig {
     }
   }
 
-  private async clearProxy(): Promise<void> {
-    await session.defaultSession.setProxy({ mode: 'direct' })
-    this.proxyUrl = null
+  private clearProxyEnv(): void {
     delete process.env.http_proxy
     delete process.env.https_proxy
     delete process.env.HTTP_PROXY
@@ -146,6 +145,12 @@ export class ProxyConfig {
     delete process.env.grpc_proxy
     delete process.env.no_proxy
     delete process.env.NO_PROXY
+  }
+
+  private async clearProxy(): Promise<void> {
+    await session.defaultSession.setProxy({ mode: 'direct' })
+    this.proxyUrl = null
+    this.clearProxyEnv()
     setGlobalDispatcher(createGlobalFetchDispatcher())
   }
 
--- a/src/main/platform/proxy.ts
+++ b/src/main/platform/proxy.ts
@@ -35,20 +35,19 @@ export function createGlobalFetchDispatcher(proxy?: {
   return new Agent(FETCH_DISPATCHER_TIMEOUTS)
 }
 
-// 合并系统和自定义的 no_proxy 设置
-function mergeNoProxy(defaultNoProxy: string): string {
-  const systemNoProxy = process.env.no_proxy || process.env.NO_PROXY || ''
-  logger.info('systemNoProxy', systemNoProxy)
-  if (!systemNoProxy) {
+// Merge app defaults with the process inherited no_proxy snapshot, not live env.
+// resolve can write NO_PROXY then later clear it; live env would lose the original.
+function mergeNoProxy(defaultNoProxy: string, inheritedNoProxy: string): string {
+  logger.info('systemNoProxy', inheritedNoProxy)
+  if (!inheritedNoProxy) {
     return defaultNoProxy
   }
-  // 将两个 no_proxy 字符串分割成数组，去重，然后重新组合
   const noProxySet = new Set(
     [
       ...defaultNoProxy.split(',').map((item) => item.trim()),
-      ...systemNoProxy.split(',').map((item) => item.trim())
+      ...inheritedNoProxy.split(',').map((item) => item.trim())
     ].filter(Boolean)
-  ) // 过滤掉空字符串
+  )
 
   return Array.from(noProxySet).join(', ')
 }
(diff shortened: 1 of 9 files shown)


src/main/provider/providers/aiSdkProvider.ts
```diff
@@ -1422,8 +1412,7 @@ export class AiSdkProvider extends BaseLLMProvider {
           ...this.defaultHeaders,
           ...this.definition.defaultHeadersPatch
         },
-        signal,
-        ...(dispatcher ? ({ dispatcher } as Record<string, unknown>) : {})
+        signal
```
zerob13, line 1417: [P2] Preserve a finite timeout for model discovery

`createModelRequestSignal(null)` returns no signal, while the global dispatcher this PR opts into sets both `headersTimeout` and `bodyTimeout` to `0`. Removing the private `ProxyAgent` therefore changes a stalled Anthropic request from Undici's finite default to an unbounded promise; the Gemini path below has the same issue. Pass `this.getModelFetchTimeout()` through the existing signal helper (and cover a never-settling fetch) before relying on the global dispatcher.

src/main/provider/providers/voiceAIProvider.ts
```diff
@@ -416,8 +391,7 @@ export class VoiceAIProvider extends BaseLLMProvider {
   private async listVoices(): Promise<VoiceStatusResponse[]> {
     const response = await fetch(this.buildUrl('/api/v1/tts/voices'), {
       method: 'GET',
-      headers: this.getAuthHeaders(),
-      ...this.getFetchOptions()
+      headers: this.getAuthHeaders()
```
zerob13: [P2] Bound voice-list requests before opting into zero dispatcher timeouts

This fetch has no `AbortSignal`, and it is used by `check()`, startup model loading, and manual refresh. With the new global dispatcher, a server that accepts a connection but never sends headers/body can leave all three operations pending forever. Reuse the provider's existing model-fetch timeout here rather than depending on a transport timeout that this PR explicitly disables.

src/main/platform/proxy.ts
```diff
@@ -89,38 +89,54 @@ export class ProxyConfig {
...
+        this.clearProxyEnv()
       }
       return true
     } catch (error) {
       console.error('Failed to resolve proxy:', error)
+      // Never throw: a rejected resolutionPromise skips every later resolve.
+      // Reuse the last installed proxy URL so a later failure does not drop it.
+      try {
+        if (this.proxyUrl) {
+          setGlobalDispatcher(
```
zerob13: [P2] Keep the installed dispatcher on resolution failure

If `setProxy()`, `resolveProxy()`, or dispatcher construction fails, the previously installed global dispatcher has not been replaced. This branch does not reuse that dispatcher; it constructs a new agent and swaps it in. Undici's `setGlobalDispatcher()` only replaces the global reference and does not close the old agent, so repeated failures abandon connection pools. Install the no-proxy timeout dispatcher once during initialization, leave the current dispatcher untouched here, and recover the serialized queue at the `resolutionPromise` boundary; the nested retry/fallback can then be deleted.

src/main/platform/proxy.ts
```diff
@@ -129,28 +145,38 @@ export class ProxyConfig {
     delete process.env.grpc_proxy
     delete process.env.no_proxy
```
zerob13, line 139: [P2] Preserve the inherited `NO_PROXY` source across transitions

`mergeNoProxy()` reads `process.env`, `commitResolvedProxy()` overwrites it, and this helper deletes it. If the process starts with `NO_PROXY=corp.internal` and SYSTEM resolves to `DIRECT`, a later CUSTOM/SYSTEM proxy no longer includes `corp.internal`. Capture the inherited no-proxy value once (separately from the app-managed env output) and add a `DIRECT -> PROXY` transition test with a pre-set `NO_PROXY`.

src/main/provider/providers/voiceAIProvider.ts
```diff
@@ -73,16 +68,7 @@ export class VoiceAIProvider extends BaseLLMProvider {
...
-    this.proxyAgent = undefined
-    this.proxyUrl = undefined
-  }
-
-  public override updateConfig(provider: LLM_PROVIDER): void {
-    super.updateConfig(provider)
-    this.proxyAgent = undefined
-    this.proxyUrl = undefined
-  }
+  public onProxyResolved(): void {}
```
zerob13, line 71: [P3] yagni: this change turns VoiceAI's proxy hook into an empty implementation; AI SDK and Ollama already carry the same stub. Give `BaseLLMProvider.onProxyResolved()` a default no-op and delete the empty overrides, leaving only ACP and Copilot to override real behavior.

zerob13 (commented):
The dispatcher centralization is directionally smaller, but the current patch removes finite transport timeouts from request paths that do not install application-level aborts, and it adds recovery state that recreates resources instead of preserving the last good state. I also found one environment ownership bug and one now-obvious no-op abstraction.

src/main/provider/providers/voiceAIProvider.ts
zerob13: This is still present at head `8c243ae`: `BaseLLMProvider.onProxyResolved()` remains abstract, and VoiceAI, AI SDK, and Ollama still implement empty stubs. Reopening because this comment was resolved without the requested simplification.

PR #2188, opened 2026-08-18
fix(acp): allow uninstall while disabled

### LINKED ISSUE #2186: [BUG] ACP is unable to delete
### System details / 系统信息

AMD Ryzen 5 5600G with Radeon Graphics | win11 | DeepChat v1.1.0

### What's wrong? / 出了什么问题？

Codex ACP cannot be deleted. When attempting to delete it, a prompt is given stating that the request failed.Uninstallation is possible only when Codex ACP is activated.
Deepchat log:[main.log](https://github.com/user-attachments/files/31174892/main.log)

1 Install Codex ACP
2 Uninstall Codex ACP

[2026-08-18 17:57:57.669] [info]  [ACP] fetchProviderModels: found 0 agents, creating models for provider "acp"
[2026-08-18 17:57:57.670] [info]  [ACP] fetchProviderModels: returning 0 models, all with providerId="acp"
[2026-08-18 17:57:57.672] [info]  [ProviderModelHelper] setProviderModels: storing 0 models for provider "acp"
[2026-08-18 17:57:57.673] [info]  [ProviderModelHelper] setProviderModels: stored 0 models for provider "acp"
[2026-08-18 17:57:57.674] [info]  [Provider] fetchModels: fetched 0 models for provider "acp"
[2026-08-18 17:57:57.674] [info]  [ProviderModelHelper] setProviderModels: storing 0 models for provider "acp"
[2026-08-18 17:57:57.675] [info]  [ProviderModelHelper] setProviderModels: stored 0 models for provider "acp"
[2026-08-18 17:57:57.676] [info]  [ACP] init: sending MODEL_LIST_CHANGED event for provider "acp"
[2026-08-18 17:57:57.677] [info]  Provider initialized successfully: ACP
[2026-08-18 17:58:02.970] [error] Error occurred in handler for 'deepchat:route:invoke': AgentUnavailableError: Agent "codex-acp" is unavailable: invalid-config
    at Object.resolveInput (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:212948:17)
    at async Object.snapshot (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:118671:11)
    at async SessionAssignment.assessTransferSession (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:136969:19)
    at async SessionAssignment.getAgentTransferImpact (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:136678:26)
    at async file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:197489:19
    at async file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:202984:14
    at async runTrackedRouteTask (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:202960:12)
    at async dispatchDeepchatRoute (file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:202983:12)
    at async file:///E:/files1/DeepChat/resources/app.asar/out/main/index.js:202994:14
    at async Session.<anonymous> (node:electron/js2c/browser_init:2:116572)

Image
Image

zerob13:
## Summary

- use lightweight session snapshots only for direct ACP transfer assessment
- discard queued input, stop running sessions, and wait for cancellation to settle before transfer
- allow active or queued sessions to be deleted through the existing runtime cleanup path
- preserve conservative blocking when pending-input inspection itself fails
- add regression coverage for unavailable ACP, active transfer, queued input, and deletion

## Root cause

Disabling a registry ACP Agent removes it from the executable ACP configuration list. The uninstall flow queried related sessions with a full runtime snapshot, which tried to resolve that disabled Agent and failed with `AgentUnavailableError: invalid-config`.

The transfer flow also treated running sessions and queued input as hard blockers. Agent removal should instead discard queue-only input and stop active work before ownership changes. Cancellation is asynchronous, so transfer now waits until the runtime leaves `generating` (with a bounded timeout) before changing the Agent context. Session deletion already performs runtime cleanup before removing durable data.

## Behavior

```text
BEFORE                              AFTER
[Agent has related sessions]        [Agent has related sessions]
             |                                   |
     [Running / queued?]                  [Running / queued?]
             |                                   |
       [Block removal]                [Discard queue + stop run]
                                                 |
                                         [Wait until settled]
                                                 |
                                        [Move or delete sessions]
                                                 |
                                           [Remove Agent]
```

## Validation

- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck:node`
- `pnpm exec vitest run test/main/session/assignment.test.ts test/main/session/session.integration.test.ts` (156 passed)
- `pnpm run test:main` (8149 passed, 457 skipped)
- PR Check: static, test-main, test-renderer, test-native-memory, build, and pr-required passed
- Package Check: package-impact and package-required passed
- CodeRabbit passed

Local `pnpm run typecheck` is blocked by the unchanged `MarkdownRenderer.vue` readonly theme tuple error; the clean CI `static` job, including typecheck, passed on the current revision.

Closes #2186

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session transfer assessments with lightweight ACP snapshots.
  * Active sessions can now be transferred safely; ongoing generation is canceled and queued inputs are removed first.
  * Transfers are blocked when pending-input checks cannot be completed.
  * Improved compatibility handling for agent migrations and unavailable ACP sessions.

* **Tests**
  * Added coverage for active-session transfers, queued-input cleanup, unavailable sessions, and successful batch transfers.

### DIFF
--- a/src/main/session/assignment.ts
+++ b/src/main/session/assignment.ts
@@ -500,7 +500,8 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
     const { handle, facet } = this.dependencies.runtime.resolveTransferSource(
       toAppSessionId(session.id)
     )
-    const state = await handle.snapshot({ lightweight: true })
+    const state =
+      handle.kind === 'acp' ? await handle.snapshot({ lightweight: true }) : await handle.snapshot()
     const status = state?.status ?? 'idle'
     let hasMessages = true
     try {
@@ -578,10 +579,9 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
       session.projectDir
     )
     const source = this.dependencies.runtime.resolveTransferSource(toAppSessionId(sessionId))
-    const sourceState = await source.handle.snapshot({ lightweight: true })
     const previousDirectAcp = source.handle.kind === 'acp'
     const previousCompatibilityAcp =
-      source.handle.kind === 'deepchat' && sourceState?.providerId === 'acp'
+      source.handle.kind === 'deepchat' && (await source.handle.snapshot())?.providerId === 'acp'
     const { facet: transferTarget } = this.dependencies.runtime.resolveDeepChatTransferTarget(
       target.agentId
     )
--- a/src/main/session/assignment.ts
+++ b/src/main/session/assignment.ts
@@ -500,7 +500,7 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
     const { handle, facet } = this.dependencies.runtime.resolveTransferSource(
       toAppSessionId(session.id)
     )
-    const state = await handle.snapshot()
+    const state = await handle.snapshot({ lightweight: true })
     const status = state?.status ?? 'idle'
     let hasMessages = true
     try {
@@ -578,7 +578,7 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
       session.projectDir
     )
     const source = this.dependencies.runtime.resolveTransferSource(toAppSessionId(sessionId))
-    const sourceState = await source.handle.snapshot()
+    const sourceState = await source.handle.snapshot({ lightweight: true })
     const previousDirectAcp = source.handle.kind === 'acp'
     const previousCompatibilityAcp =
       source.handle.kind === 'deepchat' && sourceState?.providerId === 'acp'
--- a/src/main/session/assignment.ts
+++ b/src/main/session/assignment.ts
@@ -495,6 +495,7 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
   private async assessTransferSession(session: SessionRecord): Promise<{
     status: SessionWithState['status']
     isEmptyDraft: boolean
+    queuedInputIds: string[]
     blockReason?: AgentTransferBlockReason
   }> {
     const { handle, facet } = this.dependencies.runtime.resolveTransferSource(
@@ -513,15 +514,18 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
       )
     }
 
-    let hasPendingInput = false
+    let queuedInputIds: string[] = []
+    let pendingInputInspectionFailed = false
     try {
-      hasPendingInput = (await facet.listPendingInputs(toAppSessionId(session.id))).length > 0
+      queuedInputIds = (await facet.listPendingInputs(toAppSessionId(session.id)))
+        .filter((input) => input.mode === 'queue')
+        .map((input) => input.id)
     } catch (error) {
       console.warn(
         `[SessionAssignment] Failed to inspect pending input for session=${session.id}:`,
         error
       )
-      hasPendingInput = true
+      pendingInputInspectionFailed = true
     }
 
     const hasSubagentChildren =
@@ -532,9 +536,24 @@ export class SessionAssignment implements SessionAgentAssignmentPort, SessionAss
       }).length > 0
     const isEmptyDraft = Boolean(session.isDraft) && !hasMessages && !hasSubagentChildren
 
-    if (status === 'generating') return { status, isEmptyDraft, blockReason: 'active' }
-    if (hasPendingInput) return { status, isEmptyDraft, blockReason: 'pending-input' }
-    return { status, isEmptyDraft }
+    if (pendingInputInspectionFailed) {
+      return { status, isEmptyDraft, queuedInputIds, blockReason: 'pending-input' }
+    }
+    return { status, isEmptyDraft, queuedInputIds }
+  }
+
+  private async prepareSessionForTransfer(
+    sessionId: string,
+    status: SessionWithState['status'],
+    queuedInputIds: string[]
+  ): Promise<void> {
+    if (status !== 'generating' && queuedInputIds.length === 0) return
+
+    const { handle } = this.dependencies.runtime.resolveTransferSource(toAppSessionId(sessionId))
+    if (status === 'generating') await handle.cancel()
+    for (const itemId of queuedInputIds) {
+      await handle.pending.delete(itemId)
+    }
   }
 
   private async moveSessionToAgentInternal(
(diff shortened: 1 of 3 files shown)


zerob13:
## Code Review

### 🔴 Blocking: `setSessionAgentContext` will throw for generating sessions

The core new feature — transferring active (generating) sessions — will fail in production. `prepareSessionForTransfer` calls `handle.cancel()`, but `cancel()` does **not** synchronously transition the session status to `idle`:

- `runLifecycleCoordinator.cancel()` (line 281) calls `instance.requestGenerationAbort()`, which only aborts the `AbortController` — it does not clear `activeRun` or call `transitionStatus`.
- At line 285, `hasAsyncSettlementOwner` is `true` (because `getActiveGeneration()` still returns the active run), so `cancel()` returns early.
- The status transitions to `idle` only later, asynchronously, when the provider stream settles via `applyProcessResultStatus`.

Immediately after `prepareSessionForTransfer` returns, `moveSessionToAgentUnderLifecycleGate` calls:

```ts
await transferTarget.setSessionAgentContext(toAppSessionId(sessionId), { ... })
```

This reaches `sessionSettingsCoordinator.setAgentContext` which guards at line 212:

```ts
if (state?.status === 'generating') {
  throw new Error('Cannot move session while it is generating.')
}
```

Since the status is still `'generating'`, **this throws**. The transfer fails.

**Why tests don't catch it:** Both `assignment.test.ts` and `session.integration.test.ts` mock `setSessionAgentContext` as `vi.fn().mockResolvedValue(undefined)`, bypassing the real `sessionSettingsCoordinator` guard entirely. The `getSessionState` mock returns `'generating'` for `s-active`, but that mock is never consulted by the real guard path.

**Suggested fix:** After `handle.cancel()`, wait for the status to leave `'generating'` before proceeding (e.g. poll `handle.snapshot()` with a timeout, or add a `waitForIdle()` method). Alternatively, suppress the `generating` guard when the generation has already been aborted via `cancel()` within the same operation.

---

### 🟡 Minor: orphaned steer-mode pending inputs after transfer

`assessTransferSession` filters only `mode === 'queue'` inputs into `queuedInputIds`. This is correct per the deletion contract (`assertDeletablePendingInput` at `pendingInputs.ts:541` rejects steer items). However, after `cancel()` + transfer, a pending steer input remains in the DB.

After transfer, when the session hydrates on the target agent and the pump drains (status is `idle`, `canDrainFromStatus` returns `true`), `getNextSteerInput` returns the orphaned steer record. `claimSteerInput` then creates a new assistant message and `turnStarter.start` processes it — an unexpected turn on the target agent triggered by stale input from the old generation.

This is an edge case (steer inputs are typically claimed quickly during generation), but worth a follow-up: consider releasing/consuming stale steer inputs during `prepareSessionForTransfer`, or guarding the pump against steer inputs whose original generation was cancelled.

---

### ✅ Correct aspects

- **Lightweight ACP snapshot** (`handle.snapshot({ lightweight: true })`) correctly avoids resolving the disabled ACP agent that caused the original `AgentUnavailableError`.
- **`mode === 'queue'` filter** is the right choice — steer inputs are "sent conversation facts" and cannot be deleted by design.
- **`pendingInputInspectionFailed` blocking** is appropriately conservative.
- **Deletion path is safe** — `deleteSessionTree` calls `cleanupSessionBackends` → `sessionLifecycle.cleanup` → `cancel()` + `clearOwnedState()` + `registry.evict()` before deleting data.
- **Re-snapshot after cancel** for `previousCompatibilityAcp` is correct (minor nit: lightweight snapshot would suffice since only `providerId` is checked).

---

### Summary

The ACP uninstall root cause fix (lightweight snapshots) is correct and well-tested. However, the new active-session transfer feature has a blocking bug: `cancel()` doesn't synchronously settle the status, so `setSessionAgentContext`'s `generating` guard throws in production. The mock-based tests give false confidence. Recommend fixing the status-settlement gap before merge.

zerob13:
## Re-review: blocking issue resolved ✅

The polling loop in `prepareSessionForTransfer` correctly fixes the `setSessionAgentContext` guard issue. After `handle.cancel()`, it polls `handle.snapshot()` until the status leaves `'generating'`, with a 10s timeout that throws a clear error.

### Verification

- **Polling logic** (`assignment.ts:560-572`): ACP handles use `{ lightweight: true }` snapshots during polling, deepchat handles use full snapshots — both correct. `null` snapshot safely exits the loop via `?.status`.
- **Timeout behavior**: 10s timeout throws `Session ${sessionId} did not stop before transfer.`, which propagates to `moveAgentSessions`/`moveSessionToAgent`. The session stays on the original agent — acceptable degradation.
- **Operation order**: delete queued inputs → cancel → poll for idle → `setSessionAgentContext`. The order change (delete before cancel) is sound — queued inputs are cleaned up regardless of whether cancel settles in time.
- **`deleteAgentSessions` path unaffected**: deletion calls `deleteSessionTree` → `cleanupSessionBackends` → `sessionLifecycle.cleanup` directly, not `prepareSessionForTransfer`. No regression.
- **Test fidelity**: The integration test now simulates the real cancel→idle transition (`activeCancelRequested` flag flips `activeStatus` to `idle`) and implements the real `generating` guard in the `setSessionAgentContext` mock. This would have caught the original bug. Operation order assertions (`deletePendingInput` < `cancelGeneration` < `setSessionAgentContext`) are correct.
- **ACP test**: Updated to return `idle` after `cancel()` is called, validating the polling loop for the ACP path.
- 156 tests pass.

### Remaining (non-blocking, previously noted)

The orphaned steer-input edge case is still unaddressed — after cancel+transfer, a pending `steer` input (which cannot be deleted by design) may trigger an unexpected turn on the target agent when the pump drains. Acceptable as a follow-up given steer inputs are typically claimed quickly during generation.

### Minor nit

`previousCompatibilityAcp` re-snapshot at line 620 uses a full `snapshot()`. A lightweight snapshot would suffice since only `providerId` is checked. Not blocking.

LGTM — the blocking issue is properly fixed and tested.

PR #2193, opened 2026-08-19
feat(toolchains): move Node and uv to managed installs

### LINKED ISSUE #2153: [Feature] Move Node, Bun, and Python runtimes to optional managed installs

DeepChat should stop shipping language runtime payloads inside the application package. Node.js, Bun, and Python toolchains should be optional, independently managed installs stored under DeepChat's platform data directory.

Node.js and uv are currently prepared during packaging, copied from `./runtime` into `app.asar.unpacked/runtime`, and resolved by `RuntimeHelper` from the installed application. Bun and future Python runtime support should use the same managed lifecycle instead of adding more payloads to every release artifact.

This change is needed because:

- users currently download runtime payloads even when they never use the features that require them;
- runtime security and compatibility updates are coupled to full application releases;
- application install directories may be read-only or replaced during updates;
- one built-in-runtime switch cannot express per-runtime installation, version, repair, or removal state.

### Requested behavior

Add a Runtime Manager that treats Node.js, Bun, and Python as separate optional components:

- The default DeepChat installer contains no Node.js, Bun, Python, uv, or uvx payload.
- Each runtime shows its source, installed version, available version, disk usage, integrity state, and actions such as Install, Update, Repair, and Remove.
- Installation is explicit. A feature that needs a missing runtime presents an actionable install/select-runtime flow and can be cancelled without breaking the rest of the app.
- Managed runtimes live in a versioned directory below the platform application data directory, not below the application installation directory.
- Runtime discovery and selection are defined per runtime. An explicitly selected managed, system, or custom runtime must not be silently replaced by another source.
- Runtime releases use their own version manifest and update check, independent of the DeepChat application updater.
- Downloads use DeepChat's proxy/network configuration, verify artifact integrity before activation, and install atomically so an interrupted update cannot replace a working version.
- The previous working version remains available for rollback or repair until the new version is verified.
- Application upgrades, reinstalls, and installation-directory changes preserve managed runtimes and their selected versions.
- Missing or corrupt runtimes fail closed with remediation UI; DeepChat must never execute an unverified partial download.

All current runtime consumers must be covered, including Agent/ACP process launch, MCP stdio commands, skills, and the DeepChat CLI. The app itself must still start and support runtime-independent chat on a clean installation with no managed runtimes.

For the CLI, runtime resolution must be explicit: use the configured managed/system runtime or exit with actionable installation instructions. It must not depend on a stale path inside the app bundle.

This request is limited to Node.js, Bun, and Python toolchains. RTK, OCR, CUA, and other native assets should keep their own lifecycle unless separately migrated.

### Acceptance criteria

- [ ] Release artifacts contain no bundled Node.js, Bun, Python, uv, or uvx runtime payloads.
- [ ] A clean install launches and supports runtime-independent chat without downloading a runtime.
- [ ] Node.js, Bun, and Python can be installed, updated, repaired, and removed independently.
- [ ] Managed files are stored below DeepChat's platform data directory in versioned locations.
- [ ] App updates preserve installed runtimes and runtime updates do not require a DeepChat release.
- [ ] Runtime activation is atomic and requires a verified manifest/checksum for the exact platform and architecture.
- [ ] Interrupted downloads, concurrent install requests, corrupt artifacts, and failed activation preserve the last working runtime.
- [ ] Every runtime-dependent entry point reports which runtime is missing and offers the same remediation path.
- [ ] Explicit system/custom runtime choices remain stable and are never silently shadowed.
- [ ] Packaging checks fail if a language runtime payload is accidentally reintroduced into the application artifact.
- [ ] Cross-platform tests cover macOS, Windows, and Linux path resolution, persistence across app updates, update rollback, repair, and removal.

yyhhyyyyyy:
## Summary

Stop shipping official Node in the installer. Node and uv now go through one `ToolchainService` with an explicit persisted source. Users can install the official Node pin from Settings.

Closes #2153.

Accepted deviation: uv stays in the artifact as a bundled seed, with an optional managed override. Re-downloading that small binary is worse than keeping it, and a CVE still needs a separate managed path.

## Before / After

| | Before | After |
| --- | --- | --- |
| Installer | Ships Node/npm/npx/corepack and uv | Ships uv seed only. Official Node is an on-demand managed install |
| Resolver | Each consumer poked `RuntimeHelper` and invented its own missing-runtime path | MCP, ACP, Skill, and OCR resolve only through `ToolchainService` |
| Source | Implicit. A missing bundled tree often fell through to system PATH | Explicit persisted source: `bundled \| managed \| system \| custom \| unconfigured` |
| First run | Nothing written. PATH refresh could rememoize the source | Writes `state.json`. Later PATH may promote first-run `unconfigured` → `system` once. A user-chosen `unconfigured` stays |
| Skill `auto` | Preferred system over bundled; Python could fall back to system CPython | Uses the persisted toolchain. Python `auto` fail-closes on uv |
| CLI | Needed bundled official Node | Hosted by packaged Electron (`ELECTRON_RUN_AS_NODE=1`) |
| OCR | Handshake compared a baked bundled Node string | Official Node `>=24.18.0 <25` and ABI 137. No Electron Node fallback |
| Missing runtime | Per-consumer: silent PATH, spawn a hole, or fail closed | One aggregated banner. No N dialogs |

## Behavior

- MCP, ACP, Skill, and OCR resolve Node/uv only through `ToolchainService`.
- First-run writes `state.json`. After that, source does not walk a PATH chain. Login-shell PATH may promote a first-run `unconfigured` to `system` once per userData. A user-chosen `unconfigured` is marked `explicit` and survives restart.
- Skill `auto` uses the persisted toolchain. Python `auto` fail-closes on uv; system CPython is only for explicit `system`.
- Downloads verify sha256, resume, and activate via rename. A failed or cancelled install leaves the previous tree active.
- OCR still requires official Node `>=24.18.0 <25` and ABI 137. It does not fall back to Electron Node.
- CLI is hosted by packaged Electron (`ELECTRON_RUN_AS_NODE=1`), not official Node.
- Missing runtime on auto-connect is one aggregated banner, not N dialogs.

## Known follow-ups

- Windows still does not load login-shell PATH. nvm-windows version dirs are scanned; scoop / user-only PATH may not light up as system.
- Settings `managed` does not send a version. Fine while the catalog has one Node pin.
- `RuntimeHelper` Node/uv getters are inert; ACP Init still calls them.
- Users who cleared a source on `d1dabdb14` (no `explicit` mark) may get one more PATH promotion. New clears will not.

## Summary by CodeRabbit

- **New Features**
  - Added Toolchains settings for selecting, installing, repairing, reverting, and cancelling Node.js and uv toolchains.
  - Added progress indicators, status reporting, and missing-toolchain notifications.
  - MCP services now use configured toolchains.
  - CLI services can run through the packaged application host.

- **Bug Fixes**
  - Improved reliability during concurrent service starts and stops.
  - OCR now reports clearer runtime, version, and compatibility issues.

- **Documentation**
  - Updated runtime and MCP documentation.

- **Chores**
  - Reduced packaged application size by removing bundled Node.js.


src/main/toolchains/stateStore.ts
```diff
@@ -0,0 +1,106 @@
...
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') return null
+    throw error
+  }
+}
+
+export function quarantineCorruptState(userDataDir: string): void {
+  const filePath = stateFilePath(userDataDir)
+  try {
+    renameSync(filePath, `${filePath}.corrupt`)
```
zerob13: `quarantineCorruptState` renames `state.json` to a fixed `state.json.corrupt`. If that target already exists (i.e. `state.json` became corrupt a second time, after a previous quarantine plus a later rewrite), `renameSync` throws `EEXIST` — which is not `ENOENT`, so it propagates out of `loadPersisted()`'s catch block and turns every `getState()` / `setSource()` / `getStatus()` call into a throw for the rest of the session.

Rare, but easy to make robust: use a timestamped quarantine name (or `rmSync` the old quarantine first), and/or wrap the quarantine call in its own try/catch inside `loadPersisted`.

.github/workflows/_package-windows.yml
```diff
@@ -144,12 +147,18 @@ jobs:
...
 
+      # Blocks outbound traffic from the runner Node used by smoke-light-ocr.
+      # That Node version must remain 24.18.0 to match runtime-versions.json.
       - name: Verify packaged Light OCR offline
         shell: pwsh
         run: |
           $ErrorActionPreference = 'Stop'
           $PSNativeCommandUseErrorActionPreference = $true
-          $nodePath = (Resolve-Path "dist/$env:UNPACKED_DIRECTORY/resources/app.asar.unpacked/runtime/node/node.exe").Path
+          $nodeCommand = Get-Command node -CommandType Application | Select-Object -First 1
```
zerob13: `Get-Command node -CommandType Application | Select-Object -First 1` returns the first `node` on PATH, but nothing asserts it is the 24.18.0 installed by `actions/setup-node`. On Windows runners the App Execution Alias stub at `%LOCALAPPDATA%\Microsoft\WindowsApps\node.exe` can shadow the real binary depending on PATH order, and the runner's preinstalled Node can win if setup-node's bin isn't first. The comment says the version "must remain 24.18.0", but there is no verification — if it resolves to a different binary, the firewall rule silently blocks the wrong process and the offline-OCR check becomes meaningless (or the smoke itself runs on a different Node than the one firewalled).

Suggest asserting the version before creating the rule — e.g. run `& $nodePath --version` and fail fast unless it matches 24.18.0 — and ideally resolve the executable from the same `process.execPath` the smoke script actually uses so the firewalled binary is exactly the one under test.

src/main/toolchains/catalog.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const NODE_OFFICIAL_DIST = 'https://nodejs.org/dist/'
+const NODE_DEFAULT_MIRROR_DIST = 'https://npmmirror.com/mirrors/node/'
+
+export function defaultNodeMirrorUrl(officialUrl: string): string | undefined {
+  if (!officialUrl.startsWith(NODE_OFFICIAL_DIST)) return undefined
+  return `${NODE_DEFAULT_MIRROR_DIST}${officialUrl.slice(NODE_OFFICIAL_DIST.length)}`
+}
+
+const NODE_ARCHIVES: Record<string, { filename: string; sha256: string }> = {
```
zerob13: The `NODE_ARCHIVES`/`UV_ARCHIVES` `sha256` values are hardcoded per target, and `catalog.test.ts` only asserts that the filename/URL embed `NODE_PIN`/`UV_PIN`. For Node the filename contains the version, so a pin bump forces the filename to change and (usually) the hash alongside it — but nothing forces the `sha256` to be updated with it. For uv the artifact names don't contain the version at all, so a `UV_PIN` bump that forgets the hashes passes every test and then fails at runtime with `checksum_mismatch` on first install.

Worth tying the hash to the pin explicitly (e.g. key the archive table by pin, or add a test that fails when the pin changes without the hashes changing) so a catalog bump can't silently ship stale hashes.

src/main/mcp/index.ts
```diff
@@ -896,6 +899,32 @@ export class McpService implements McpServicePort {
     await this.mcpSettings.removeMcpServer(serverName)
   }
 
+  async retryUnstartedEnabledServers(): Promise<void> {
```
zerob13, line 902: `retryUnstartedEnabledServers` runs at MCP init and on every `toolchains.changed` event (source change, PATH refresh, install completion). It retries *all* enabled-but-unstarted servers, including ones failing for reasons unrelated to toolchains (bad URL, missing package, auth). A server that keeps failing will be respawned on every toolchain mutation — with many enabled servers that's a burst of spawns each time, and a failing server looks like a restart loop in logs.

Consider only retrying servers whose last failure was a toolchain-resolution error (the MCP failure path already carries a typed reason), or add a small per-server backoff/cooldown so unrelated failures don't get retried on every toolchain change.

zerob13 (commented):
Reviewed the toolchain service, the consumers (MCP / ACP / Skill / OCR / CLI), the downloader/extract/probe stack, and the CI changes. Overall the design is solid: one resolver, an explicit persisted source, atomic activation, and the banner aggregation are all good calls. I found a few things I think are worth addressing before merge — detailed inline.

src/main/toolchains/routes.ts
```diff
@@ -0,0 +1,125 @@
...
+      async (rawInput) => {
+        const input = toolchainsRevertRoute.input.parse(rawInput)
+        return toolchainsRevertRoute.output.parse(deps.service.revert(input.kind))
+      }
+    ],
+    [
+      toolchainsPickCustomRoute.name,
+      async (rawInput) => {
+        const input = toolchainsPickCustomRoute.input.parse(rawInput)
+        const picked = await deps.pickPath()
```
zerob13, line 91: `pickCustom` delegates to `deviceService.selectFiles({ multiple: false })`, so the UI can only ever pick a file. But `probeCustomNode`/`probeCustomUv` explicitly support a directory (toolchain root) as well, and the natural pick for e.g. an nvm root or a uv install directory is a directory, not the executable inside it. Either the picker should support directories, or the custom option should be clearly documented as file-only (in which case selecting a directory is dead UI).

src/main/mcp/index.ts
yyhhyyyyyy: Leaving this for a follow-up. Last errors are still plain strings, so we cannot tell a toolchain miss from a bad URL or auth failure. Adding backoff or typed retry here would change MCP start/stop behavior beyond this PR.

src/main/toolchains/routes.ts
yyhhyyyyyy: Leaving this as file-only for now. The current picker contract is the executable, and opening directories would need a device-dialog plus Settings copy change. Probe still accepts a typed-in root if we add that later.

zerob13:
Follow-up on the remaining items after `a9ec4191` — the state quarantine, the firewall pin assertion, and the catalog pin-keyed hashes are all verified fixed. Three things are still open:

1. **`retryUnstartedEnabledServers` still retries unconditionally** ([src/main/mcp/index.ts:902](https://github.com/ThinkInAIXYZ/deepchat/blob/a9ec4191c07b028015888b41d90ef6e444d7c397/src/main/mcp/index.ts#L902)). It retries *every* enabled-but-unstarted server on MCP init and on every `toolchains.changed` event, including servers failing for unrelated reasons (bad URL, missing package, auth). A persistently failing server gets respawned on each toolchain mutation with no cooldown. If this is a deliberate deferral, fine — but it isn't listed under "Known follow-ups", so I can't tell whether it was triaged or missed.

2. **`pickCustom` is still file-only** ([src/main/toolchains/routes.ts:91](https://github.com/ThinkInAIXYZ/deepchat/blob/a9ec4191c07b028015888b41d90ef6e444d7c397/src/main/toolchains/routes.ts#L91) → `deviceService.selectFiles` uses `properties: ['openFile']`). Meanwhile `probeCustomNode`/`probeCustomUv` explicitly support a directory root (an nvm root or a uv install dir is the natural pick). As shipped, the directory path in the probe is unreachable from the UI. Either add `openDirectory` to the picker or drop the directory branch from the probe.

3. **`spawnSync` still blocks the main process** ([src/main/toolchains/service.ts:968](https://github.com/ThinkInAIXYZ/deepchat/blob/a9ec4191c07b028015888b41d90ef6e444d7c397/src/main/toolchains/service.ts#L968), already flagged by CodeRabbit). First inspection of a slow/hung system or custom Node freezes the entire Electron main process — UI, IPC, every window — for up to 5 s, and the inspection cache is cleared on every `persist()` and `updateDetectionEnv()`. This one I'd want addressed (or at least consciously accepted) before merge, not after.

1 and 2 are fine as documented follow-ups if that's the call; 3 I'd treat as blocking.

PR #2194, opened 2026-08-20
feat(markstream): upgrade and inject stream-diffs worker pool

Simon-He95:
## Summary

- Upgrade `markstream-vue` from `2.0.0-beta.2` to `2.0.1-beta.1` (the newest release that exposes `setStreamDiffsWorkerPool`; stable `2.0.0` does not include it).
- Add `@pierre/diffs@1.2.12` as a direct dependency (deduped with `stream-diffs@0.0.2`'s transitive version) so the worker pool entry can be imported.
- Inject the shared stream-diffs highlight worker pool into markstream-vue in `markdownWorkerLifecycle.ts`, following the existing lazy KaTeX/Mermaid worker pattern:
  - Lazily create the pool via `getOrCreateWorkerPoolSingleton` with `poolSize: 4` and the `@pierre/diffs/worker/worker.js?worker` factory.
  - Initial theme `{ dark: 'vitesse-dark', light: 'vitesse-light' }` matches the renderer's `codeBlockThemes`.
  - `setStreamDiffsWorkerPool(pool)` registers it; `beforeunload`/test reset terminates the pool.
  - Failure-isolated: without a pool markstream falls back to main-thread highlighting.

## Verification

- `typecheck:web` / `typecheck:node` pass
- 83 targeted renderer tests pass (MarkdownRenderer, MarkstreamDomContract, ThinkContent, MessageBlockToolCall, markstreamLanguage)
- `lint`, `format:check`, `i18n` pass
- Production `electron-vite build` passes; diffs worker emitted as a separate asset and the pool module is only dynamically loaded on first markdown render (not in the main entry)

## Summary by CodeRabbit

- **New Features**
  - Improved streamed content rendering with enhanced syntax-diff highlighting.
  - Added background processing for code highlighting to keep the interface responsive.
  - Added resilient worker initialization with automatic recovery after temporary failures.

- **Bug Fixes**
  - Improved cleanup and reset behavior for rendering workers.

- **Chores**
  - Updated the Markstream rendering component.
  - Added development support for diff highlighting and worker lifecycle testing.


zerob13 (changes requested):
## Code Review — PR #2194

Overall the worker-pool injection is well-structured: idempotent init, failure-isolated fallback to main-thread highlighting, and correct cleanup wiring. The theme config matches `MarkdownRenderer.vue`'s `codeBlockThemes`. A few issues need addressing before merge.

---

### 🔴 Must Fix

**1. spec.md inconsistency — `stream-diffs` vs `@pierre/diffs`**

`spec.md` still says:
> 增强代码块使用其 `stream-diffs@0.0.2` peer

But the code now directly imports `@pierre/diffs/worker` and `@pierre/diffs/worker/worker.js?worker`, bypassing `stream-diffs` entirely for the worker pool. The spec should be updated to document:
- Why `@pierre/diffs` is introduced as a direct dependency (the worker pool singleton API is not exposed through `stream-diffs`)
- The relationship between `stream-diffs` (still present as markstream-vue's peer dep) and `@pierre/diffs` (direct, for worker pool lifecycle)

**2. Missing test coverage for new pool lifecycle logic**

`mainMarkdownWorkers.test.ts` only asserts `main.ts` doesn't eagerly call `ensureMarkdownWorkers`. The new code paths have zero coverage:
- `ensureStreamDiffsWorkerPool` idempotency (second call returns early when pool exists)
- Failure retry: after a failed `loadDiffsWorkerConstructor` or `getOrCreateWorkerPoolSingleton`, `diffsWorkerPool` stays `null` and the next `ensureMarkdownWorkers` call retries — this is the key design decision (comment: "transient injection failure is retried") and should be tested
- `cleanupMarkdownWorkers` calls `terminateStreamDiffsWorkerPool` and resets `diffsWorkerPool = null`
- `_resetForTesting` resets `DiffsWorkerConstructor`

At minimum, mock `@pierre/diffs/worker` and `markstream-vue` exports, then verify the retry-after-failure and cleanup paths.

**3. Singleton pool cleanup safety needs verification**

`getOrCreateWorkerPoolSingleton` from `@pierre/diffs/worker` creates a **module-level singleton**. `cleanupMarkdownWorkers` calls `terminateStreamDiffsWorkerPool()` (a `markstream-vue` API) and sets `diffsWorkerPool = null`, but the `@pierre/diffs` singleton reference may still exist internally.

If `terminateStreamDiffsWorkerPool` terminates the pool's workers but `getOrCreateWorkerPoolSingleton` returns the same (now-terminated) singleton on next init, the pool could be in a dead state. Please verify one of:
- `getOrCreateWorkerPoolSingleton` recreates the pool after termination (safe), **or**
- There's an explicit `@pierre/diffs` API to reset/destroy the singleton that should be called in `cleanupMarkdownWorkers`

---

### 🟡 Suggestions

**4. `@pierre/diffs` version pinning**

Pinned to exact `1.2.12` while `stream-diffs@0.0.2` declares `@pierre/diffs: ^1.2.12` (latest is `1.3.5`). If the pin is intentional for reproducibility, a brief comment explaining why (vs. `^1.2.12`) would help future maintainers. If not intentional, consider `^1.2.12` to pick up patch fixes.

**5. `STREAM_DIFFS_POOL_SIZE` rationale**

The comment says "Keep the highlight pool small" but doesn't explain why 4 specifically. A one-line note (e.g., "4 workers cover typical concurrent code blocks in a streaming chat without excessive memory") would make the constant self-documenting.

---

### ✅ What looks good

- Failure isolation: `ensureStreamDiffsWorkerPool` catches and logs without throwing, so markdown worker init isn't blocked
- `ensureMarkdownWorkers` callers use `.catch()` fire-and-forget, so the async pool init won't block rendering
- Theme config `{ dark: 'vitesse-dark', light: 'vitesse-light' }` correctly matches `MarkdownRenderer.vue:118`
- Code style conforms to project conventions (single quotes, no semicolons, 100 cols)
- No breaking changes: `stream-diffs@0.0.2` retained, markstream-vue peer dep satisfied, new APIs are additive

Simon-He95:
Addressed the review feedback in `6383eb76b`:

**CodeRabbit**
- `ensureStreamDiffsWorkerPool`: `setStreamDiffsWorkerPool(pool)` now runs before caching `diffsWorkerPool`; the catch path calls `terminateWorkerPoolSingleton()` so a retry builds a fresh pool.
- `ensureMarkdownWorkers` catch now tears down the already-injected pool (`terminateStreamDiffsWorkerPool()` + reset) so a failed lifecycle leaves no half-initialized state.

**zerob13**
1. `spec.md` now documents why `@pierre/diffs` is a direct dependency (worker-pool singleton API is not re-exported by `stream-diffs`) and the relationship between the two.
2. Added `test/renderer/lib/markdownWorkerLifecycle.test.ts` covering pool injection idempotency, failure retry + singleton reset, cleanup, and re-injection after `_resetForTesting` (4 tests).
3. Singleton safety verified: `@pierre/diffs` `WorkerPoolManager.terminate()` resets `initialized`/workers/caches, so the cached singleton re-initializes safely on the next mount; noted in a code comment.
4. `@pierre/diffs` range changed to `^1.2.12` to match `stream-diffs`' declared range (lock still resolves `1.2.12`, deduped).
5. Added a rationale comment for `STREAM_DIFFS_POOL_SIZE = 4`.

Verification re-run: `typecheck:web`, `lint`, `format`, 87 targeted renderer tests (4 new + 83 existing), and the production `electron-vite build` all pass.

PR #2195, opened 2026-08-20
feat(acp): support terminal authentication

### LINKED ISSUE #2144: [Feature] Support ACP terminal authentication during agent onboarding

DeepChat should support the ACP v1 terminal authentication flow for registry and custom agents. This is needed by agents such as MiniMax Code (`mcode acp`) that advertise a terminal auth method from `initialize` when the client declares `clientCapabilities.auth.terminal`.

The current capability builder already supports `enableTerminalAuth`, and the process handle preserves returned `authMethods`, but the normal ACP process initialization only enables filesystem and terminal capabilities. As a result, terminal-auth methods are not advertised to the agent and there is no first-run onboarding path outside the ACP debug surface. Pre-authenticated users can work, but a new user cannot complete login from DeepChat.

Requested behavior:

- advertise terminal auth only when the normal runtime can actually execute and present the flow;
- surface returned terminal `authMethods` in ACP agent onboarding / auth-required states;
- run the selected terminal method with the same verified launch command plus the method args, without shell interpolation;
- call ACP `authenticate` after the terminal command succeeds, then retry the blocked Session creation;
- handle cancellation, process failure, reconnect, and agents with multiple methods;
- add process-manager and user-flow tests so capability advertisement never exceeds the implemented behavior.

This is a general ACP client capability, not a MiniMax-specific built-in registration. MiniMax Code itself should become discoverable through the official ACP Registry after its public npm release.

zerob13:
## Summary

- advertise `clientCapabilities.auth.terminal` and preserve typed ACP authentication methods
- handle numeric `auth_required` results with bounded agent or terminal authentication flows
- launch terminal authentication from the immutable agent launch snapshot, then reconnect and retry without calling `authenticate`
- add onboarding/settings UI, typed IPC contracts, lifecycle integration, SDD, and focused regression coverage

## UI

BEFORE

```text
New ACP thread
  └─ session/new → auth_required → generic failure
```

AFTER

```text
New ACP thread
  └─ session/new → auth_required
      └─ authentication dialog
          ├─ agent method → authenticate → retry
          └─ terminal method → embedded PTY → reconnect → retry
              └─ ready draft
```

## Validation

- `pnpm run format`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`
- focused main ACP/session/routes tests
- focused renderer API/settings/new-thread/auth-dialog tests
- `pnpm run build`
- `acpx` interoperability probes with deterministic fake ACP and `pi-acp@0.0.33`

Closes #2144

## Summary by CodeRabbit

* **New Features**
  * Added ACP authentication with sign-in methods and interactive terminal authentication.
  * Added authentication dialogs with method selection, terminal output, status updates, cancellation, and reconnect handling.
  * New threads now identify required authentication and retry setup after successful sign-in.
  * Added localized authentication messaging across supported languages.
  * Updated the ACP agent registry with newer releases.

* **Bug Fixes**
  * Improved authentication ownership, cleanup, cancellation, timeout, and output-limit handling.

* **Documentation**
  * Added ACP terminal-authentication specifications and implementation plans.


src/main/agent/acp/runtime/acpProcessManager.ts
```diff
@@ -599,6 +642,186 @@ export class AcpProcessManager implements AgentProcessManager<AcpProcessHandle,
...
+    workdir?: string
+  ): Promise<AcpAuthChallenge> {
+    const handle = await this.getConnection(agent, workdir)
+    return this.createAuthChallenge(handle, { origin: 'settings_probe' })
+  }
+
+  async authenticateAgent(challengeId: string, methodId: string): Promise<void> {
+    const challenge = await this.claimAuthChallenge(challengeId, methodId, 'agent')
+    try {
+      await challenge.handle.connection.authenticate({ methodId })
```
yyhhyyyyyy: [P1] The ACP SDK has no timeout or AbortSignal here and does not reject pending requests when the stream closes. A hung or crashed agent leaves the auth scope locked until restart; race this with timeout, connection close, renderer destruction, and shutdown, then release exactly once.

src/main/agent/acp/runtime/acpProcessManager.ts
```diff
@@ -599,6 +642,186 @@ export class AcpProcessManager implements AgentProcessManager<AcpProcessHandle,
...
+    }
+  }
+
+  async completeTerminalAuthentication(challengeId: string): Promise<void> {
+    const challenge = this.requireStoredAuthChallenge(challengeId)
+    if (!challenge.active) {
+      throw new Error('ACP authentication challenge is not active')
+    }
+    try {
+      await this.disposeHandle(challenge.handle)
```
yyhhyyyyyy, line 817: [P1] `challenge.handle` remains a reusable warmup while terminal login runs, so another session can bind it before this line. This can kill a newly active session; reserve the handle during auth or verify it is still the same unbound warmup under the binding lock before disposal.

yyhhyyyyyy (commented):
Review findings are left inline.

src/main/agent/acp/auth/acpAuthService.ts
```diff
@@ -0,0 +1,259 @@
...
+  constructor(private readonly dependencies: AcpAuthServiceDependencies) {}
+
+  async inspect(agentId: string, workdir?: string): Promise<AcpAuthChallenge> {
+    this.ensureActive()
+    const agent = await this.resolveAgent(agentId)
+    const challenge = await this.dependencies.owner
+      .getOrCreate()
+      .processManager.inspectAuthentication(agent, workdir)
+    this.ensureActive()
+    this.rememberStatus({
```
yyhhyyyyyy, line 40: [P2] Required challenges are stored without `ownerWebContentsId`, so another renderer that obtains this ID can start first and become the terminal owner. Bind the challenge when issued and enforce ownership before lookup/start.

src/main/agent/acp/auth/acpTerminalAuthRunner.ts
```diff
@@ -0,0 +1,135 @@
...
+  shutdown(): void {
+    for (const [runId, run] of this.runs) {
+      this.terminate(runId, run)
+    }
+  }
+
+  private terminate(runId: string, run: AcpTerminalAuthRun): void {
+    run.cancelled = true
+    try {
+      run.pty.kill()
```
yyhhyyyyyy, line 141: [P2] Cancellation cleanup still depends on `onExit`. If `kill()` does not emit it, completion never settles and the auth scope remains locked; add a bounded fallback that settles and removes the run without the event.

src/renderer/src/components/acp/AcpAuthDialog.vue
```diff
@@ -0,0 +1,257 @@
...
+  const attempt = ++authenticationAttempt
+  error.value = null
+  state.value = 'running'
+  try {
+    const result = await client.start(props.challenge.id, selectedMethod.value.id)
+    if (attempt !== authenticationAttempt || !props.open) {
+      if (result.runId) cancelRun(result.runId)
+      return
+    }
+    state.value = result.state
```
yyhhyyyyyy, line 191: [P2] A terminal `stateChanged` event can arrive before this older start response, after which these assignments move the UI back to `running`. Include a monotonic revision in responses and events and ignore stale updates.

src/main/agent/acp/runtime/acpProcessManager.ts
```diff
@@ -599,6 +642,186 @@ export class AcpProcessManager implements AgentProcessManager<AcpProcessHandle,
...
+      await this.getConnection(challenge.agent, challenge.public.workdir)
```
yyhhyyyyyy, line 818: [P2] If reconnect fails after the old handle is disposed, the unconsumed challenge still points to that dead handle, so every UI retry fails as stale. Consume or replace it and expose a fresh challenge/re-inspect path.

src/main/agent/acp/auth/acpTerminalAuthRunner.ts
```diff
@@ -0,0 +1,135 @@
...
+      pty,
+      cancelled: false,
+      completion,
+      timeout: null,
+      forceKillTimeout: null
+    }
+    this.runs.set(runId, run)
+    const dataSubscription = pty.onData((data) => {
+      for (let offset = 0; offset < data.length; offset += 65_536) {
+        input.onData(runId, data.slice(offset, offset + 65_536))
```
yyhhyyyyyy, line 87: [P2] Chunking limits individual IPC payloads but not total output or rate. A noisy auth process can build an unbounded main-to-renderer queue; add per-run backpressure, queue limits, and truncation or termination on overflow.

PR #2201, opened 2026-08-21
fix(ui): format decimal values in memory diagnostics panel

Dorence:
image

Avoid directly display numbers in memory diagnostics panel

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting for archive age, decay scores, and recall latency values.
  * Added locale-aware precision for diagnostic metrics, including fractional values.
  * Ensured percentile latency and archive-candidate metrics display consistently.
  * Preserved clear display of unavailable latency values and non-finite diagnostic values.

### DIFF
--- a/src/renderer/settings/components/MemoryDiagnosticsPanel.vue
+++ b/src/renderer/settings/components/MemoryDiagnosticsPanel.vue
@@ -164,6 +164,7 @@
         </DcSectionCard>
 
         <DcSectionCard
+          data-testid="archive-candidates"
           :title="t('settings.memory.redesign.archiveCandidatesTitle')"
           :description="t('settings.memory.redesign.archiveCandidatesDescription')"
         >
@@ -195,12 +196,12 @@
               <div class="mt-1 flex flex-wrap gap-x-3 gap-y-1 text-muted-foreground">
                 <span>
                   {{ t('settings.deepchatAgents.memoryManager.health.archivePrediction.ageDays') }}:
-                  {{ formatNumber(lifecycle.forget.ageDays) }}
+                  {{ formatDays(lifecycle.forget.ageDays) }}
                 </span>
                 <span>
                   {{
                     t('settings.deepchatAgents.memoryManager.health.archivePrediction.decayScore')
-                  }}: {{ formatNumber(lifecycle.forget.decayScore) }}
+                  }}: {{ formatScore(lifecycle.forget.decayScore) }}
                 </span>
               </div>
             </li>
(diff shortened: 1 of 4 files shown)


zerob13 (changes requested):
## Review — format decimal values in memory diagnostics panel

Thanks for the fix. Using `Intl.NumberFormat` for locale-aware formatting is the right direction and consistent with the sibling `MemoryLifecyclePanel.vue`.

### 🔴 Must fix

**1. Duplicated formatters with `MemoryLifecyclePanel.vue`**

The newly added `decimalFormatter`, `dayFormatter` computeds and `formatScore` / `formatDays` functions are **byte-for-byte identical** to the ones already in `MemoryLifecyclePanel.vue` (L163–176, L257–269):

| definition | MemoryLifecyclePanel.vue | MemoryDiagnosticsPanel.vue (this PR) |
|---|---|---|
| `decimalFormatter` | L163–168 | L345–351 |
| `dayFormatter` | L170–175 | L352–358 |
| `formatScore` | L257–259 | L489–491 |
| `formatDays` | L267–269 | L493–495 |

Both components live in the same directory and operate on the same memory lifecycle/diagnostics domain. Please extract a shared composable (e.g. `useMemoryNumberFormatters`) into `settings/lib/` — `useMemoryInlineFeedback.ts` there is a good precedent — and have both panels consume it. Otherwise the two copies will silently drift as formatting requirements evolve.

### 🟡 Suggestions (nits)

**2. `formatOptionalMilliseconds` is misnamed**

The function delegates to `decimalFormatter` (a generic decimal formatter), so "Milliseconds" in the name is misleading — nothing about it is millisecond-specific. The sibling component calls its nullable variant `formatOptionalScore`. Consider a domain-agnostic name like `formatOptionalDecimal`, or align with the sibling's naming once the shared composable lands.

**3. Dropped `Number.isFinite` guard**

The previous `formatNumber` guarded against `NaN` / `Infinity` and fell back to `String(value)`. The new helpers pass straight to `Intl.NumberFormat.format`, so `Infinity` now renders as `∞` instead of `Infinity`. Realistic data is unlikely to hit this, but when extracting the shared composable, consider restoring a finite-check guard for robustness.

### ✅ Looks good

- `Intl.NumberFormat` over `toFixed` — correct, locale-aware, consistent with `MemoryLifecyclePanel` / `OcrSettings` / `DashboardSettings`.
- `computed`-cached formatter instances that rebuild on `locale` change — correct.
- Test coverage for float truncation (`685.716`, `0.123`, `12.3`) and the `—` fallback — solid, no over-testing.
- Scope is tight and focused; no breaking changes.

Requested changes center on the duplication — once the shared composable is extracted and both panels consume it, this is good to go.

Issue #2174, opened 2026-08-17
[Feature] Add the function of controlling the frame rate of the client side


jianpuwuwei:
### What do you need? / 你需要什么？

The frame rate will be particularly high when the program window is displayed on the front end.
<img width="405" height="42" alt="Image" src="https://github.com/user-attachments/assets/d215f518-f236-4602-84d8-ce1ce6ca6f8f" />
When there is no client window on the desktop:
<img width="410" height="48" alt="Image" src="https://github.com/user-attachments/assets/0f052171-87de-400e-86bc-df3a0eb8993f" />
I think it would be feasible to add a function to set a frame rate limit in the settings, such as limiting it to 60fps or 30fps. Generally, a frame rate of over 100fps is not necessary.

jianpuwuwei:
The client merely needs to display a loading indicator circle and may not require such a high frame rate.

zerob13:
Thanks for reporting this. Reducing unnecessary rendering and power usage is a valid concern, especially on high-refresh-rate displays.

However, the screenshots alone do not confirm that DeepChat is responsible for the reported GPU usage—the GPU load remains nearly identical when the DeepChat window is not visible. Also, Electron’s `webContents.setFrameRate()` only applies to offscreen rendering, so it cannot reliably limit the frame rate of a normal application window.

Before deciding whether a new setting is appropriate, could you provide:

- DeepChat version
- Windows version and GPU model
- Display refresh rate
- Whether Content Protection is enabled
- DeepChat’s per-process GPU usage in Windows Task Manager while idle, waiting for a response, streaming a response, and minimized

We will treat this as a performance investigation first. If a continuous loading animation or another rendering path is causing measurable resource usage, we would prefer to optimize that specific behavior rather than add a frame-rate setting that may not work as expected.

jianpuwuwei:
DeepChat v1.1.0
Win11 25H2 OS26200.8875 | NVIDIA GeForce GTX 1660 SUPER   Drive32.0.16.1088
Screen refresh rate179.76 Hz
Content Protection is not enabled

Idle
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/789a4b63-34f1-45b6-bd87-0153c77e3ff8" />
Waiting for a response
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/8f0085b1-3b7a-4b8d-ac6b-656d7ca42824" />
Streaming a response
[...]

PR #2202, opened 2026-08-24
perf(markdown): gentle prefix batching + fix(retry): keep user message on retry

zhangmo8:
## Summary

> **Note:** this PR now contains two independently reviewable changes — the original **markdown prefix batching** perf work, plus a **retry fix** (keep the user message on retry).

### Part 1 — perf(markdown): gentle batch for prefix advance

link #2174

Follow-up to #2200: the committed-prefix segment now advances in ~6 KB chunks, and mounting a whole chunk at once queued a burst of GPU rasterization (profiles showed 0.2–1 s GPU-process stalls right after each prefix advance, with a visible draw gap).

This renders the **prefix segment with gentle, budget-capped batching** (smaller batch, a frame delay, and a tight per-frame budget) instead of the big static batch, so the mount and raster work is spread across frames rather than stalling.

- Streaming tail behavior and the whole-render path are unchanged.
- Node virtualization on the prefix is kept.

### Part 2 — fix(retry): keep user message on retry

Reworks retry so the retried user prompt survives instead of being deleted and recreated as a fresh record:

- **Assistant retry**: truncates from the retried assistant message onward; the preceding user prompt stays in place and is reused via the pre-stream anchor (`retryFromOrderSeq` / `sourceMessageId`). The ACP path intentionally keeps the old inclusive-truncation behavior.
- **User retry**: keeps the user prompt itself and truncates from the next message — the message no longer disappears and reappears in the UI.
- **Compaction budget**: when retry reuses a user prompt already present in the history, `prepareForNextUserTurn` no longer projects it a second time, avoiding an inflated context estimate that could prematurely trigger compaction.
- **Renderer**: optimistically truncates stale rows before the retry IPC and restores them via a reload when a retry is blocked.

### Verification note

Behavior/unit tests pass (including harness retry, session, renderer, and compaction suites). The performance benefit of Part 1 should be confirmed with a streaming profile; the change is low-risk either way.

Refs #2200

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retry behavior for assistant and user messages, preserving the correct conversation context.
  - Prevented retried prompts from appearing twice in requests or conversation history.
  - Restored message views correctly when a retry is blocked or rejected.
  - Improved transcript truncation so messages after a retry point are removed consistently.
  - Optimized streaming-prefix rendering with smoother, smaller incremental updates.

### DIFF
--- a/src/main/agent/deepchat/runtime/turnCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/turnCoordinator.ts
@@ -800,6 +800,13 @@ export class TurnCoordinator {
             state.providerId,
             state.modelId
           )
+        // A retried user prompt that is anchored and kept in the transcript is
+        // already part of the history, so compaction must not project it a second
+        // time (avoids inflating the context estimate).
+        const anchorMessageId = instance.getPreStreamTranscriptAnchorId()
+        const newUserContentInHistory = Boolean(
+          anchorMessageId && historyRecords.some((record) => record.id === anchorMessageId)
+        )
         return await this.runPreStreamStep(
           {
             sessionId,
@@ -822,6 +829,7 @@ export class TurnCoordinator {
               preserveEmptyInterleavedReasoning:
                 interleavedReasoning.preserveEmptyReasoningContent === true,
               newUserContent: content,
+              newUserContentInHistory,
               ...(pendingContextPressure ? { forceContextPressure: true } : {}),
               historyRecords,
               signal: preStreamAbortSignal
--- a/src/main/agent/deepchat/runtime/turnCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/turnCoordinator.ts
@@ -148,6 +148,7 @@ export interface TurnStartContext {
   maxProviderRounds?: number
   preserveResolvedRepresentations?: boolean
   beforeHistoryPreparation?: () => void
+  preStreamAnchorMessageId?: string
 }
 
 export interface TurnExecutionContext extends TurnStartContext {
@@ -495,7 +496,9 @@ export class TurnCoordinator {
       if (!instance) throw new Error(`Session ${sessionId} not found`)
       instance.clearPreStreamTranscriptAnchor()
       const reservedTranscriptAnchor =
-        reservedSteerAssistantMessageId ?? linkedSteerMessageIds.at(-1)
+        context?.preStreamAnchorMessageId ??
+        reservedSteerAssistantMessageId ??
+        linkedSteerMessageIds.at(-1)
       if (reservedTranscriptAnchor) {
         instance.setPreStreamTranscriptAnchorId(reservedTranscriptAnchor)
       }
--- a/src/main/agent/deepchat/runtime/turnCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/turnCoordinator.ts
@@ -1003,6 +1003,14 @@ export class TurnCoordinator {
         >
       ) => {
         const contextBuildStartedAt = Date.now()
+        // A retried user prompt that is anchored and kept in the transcript is
+        // already carried by historyRecords; exclude it from the history turns so
+        // the newUserMessage (which holds the live turn and memory injection)
+        // does not duplicate it.
+        const anchorMessageId = instance.getPreStreamTranscriptAnchorId()
+        const newUserContentInHistory = Boolean(
+          anchorMessageId && historyRecords.some((record) => record.id === anchorMessageId)
+        )
         const contextBuild = buildTapeChatView({
           sessionId,
           newUserContent: content,
@@ -1020,7 +1028,9 @@ export class TurnCoordinator {
             preserveInterleavedReasoning: interleavedReasoning.preserveReasoningContent,
             preserveEmptyInterleavedReasoning:
               interleavedReasoning.preserveEmptyReasoningContent === true,
-            providerReplayProjector
+            providerReplayProjector,
+            newUserContentInHistory,
+            newUserContentMessageId: newUserContentInHistory ? anchorMessageId : undefined
           }
         })
         logSlowPreStreamStep(sessionId, 'context-build', contextBuildStartedAt)
--- a/src/main/session/transcriptMutations.ts
+++ b/src/main/session/transcriptMutations.ts
@@ -43,7 +43,13 @@ export class SessionTranscriptMutations {
   async prepareRetryMessage(
     sessionId: string,
     messageId: string
-  ): Promise<{ content: SendMessageInput; projectDir: string | null; sourceOrderSeq: number }> {
+  ): Promise<{
+    content: SendMessageInput
+    projectDir: string | null
+    sourceOrderSeq: number
+    retryFromOrderSeq: number
+    sourceMessageId: string | null
+  }> {
     const target = this.requireMessage(sessionId, messageId)
     const sourceUserMessage =
       target.role === 'user'
@@ -62,14 +68,25 @@ export class SessionTranscriptMutations {
       throw new Error('Cannot retry an empty user message.')
     }
 
-    return { content, projectDir, sourceOrderSeq: sourceUserMessage.orderSeq }
+    return {
+      content,
+      projectDir,
+      sourceOrderSeq: sourceUserMessage.orderSeq,
+      // Truncate from the retried message (not the source user message) so the
+      // user prompt survives and only the assistant reply is regenerated. When
+      // the retried message IS the user prompt, keep it too and truncate from
+      // the next order sequence, anchoring it so the same record is re-sent
+      // instead of a fresh duplicate.
+      retryFromOrderSeq: target.role === 'user' ? target.orderSeq + 1 : target.orderSeq,
+      sourceMessageId: sourceUserMessage.id
+    }
   }
 
-  commitRetryMessage(sessionId: string, sourceOrderSeq: number): void {
+  commitRetryMessage(sessionId: string, orderSeq: number): void {
     this.dependencies.runInTransaction(() => {
-      this.dependencies.transcript.deleteFromOrderSeq(sessionId, sourceOrderSeq)
+      this.dependencies.transcript.deleteFromOrderSeq(sessionId, orderSeq)
     })
-    this.dependencies.runtime.invalidateTranscriptFrom(sessionId, sourceOrderSeq)
+    this.dependencies.runtime.invalidateTranscriptFrom(sessionId, orderSeq)
   }
 
   async deleteMessage(sessionId: string, messageId: string): Promise<void> {
--- a/src/main/session/transcriptMutations.ts
+++ b/src/main/session/transcriptMutations.ts
@@ -68,6 +68,14 @@ export class SessionTranscriptMutations {
       throw new Error('Cannot retry an empty user message.')
     }
 
+    // The retried user prompt is kept in the transcript and reused as the
+    // pre-stream anchor. A failed/error row (e.g. a failed Steer prompt) must be
+    // restored to 'sent' so context history filtering keeps it; otherwise the
+    // prompt silently drops out of future turns.
+    if (sourceUserMessage.status !== 'sent') {
+      this.dependencies.transcript.updateMessageStatus(sourceUserMessage.id, 'sent')
+    }
+
     return {
       content,
       projectDir,
--- a/src/renderer/src/components/markdown/MarkdownRenderer.vue
+++ b/src/renderer/src/components/markdown/MarkdownRenderer.vue
@@ -422,8 +422,11 @@ const renderSegments = computed<RenderSegment[]>(() => {
       smoothStreaming: false,
       typewriter: false,
       nodeVirtual: 'auto',
-      maxLiveNodes: STATIC_MAX_LIVE_NODES,
-      liveNodeBuffer: STATIC_LIVE_NODE_BUFFER,
+      // Incremental batching only takes effect when virtual live-node limiting is
+      // off (markstream-vue gates batching on `maxLiveNodes <= 0`), so the prefix
+      // must not pin live nodes or the gentle raster spreading never happens.
+      maxLiveNodes: 0,
+      liveNodeBuffer: 0,
       initialBatch: PREFIX_INITIAL_RENDER_BATCH_SIZE,
       batchSize: PREFIX_RENDER_BATCH_SIZE,
       batchDelay: PREFIX_RENDER_BATCH_DELAY_MS,
--- a/src/renderer/src/components/markdown/MarkdownRenderer.vue
+++ b/src/renderer/src/components/markdown/MarkdownRenderer.vue
@@ -244,6 +244,15 @@ const STATIC_RENDER_BATCH_IDLE_TIMEOUT_MS = 16
 const STATIC_PARSE_COALESCE_MS = 0
 const STATIC_MAX_LIVE_NODES = 260
 const STATIC_LIVE_NODE_BUFFER = 80
+// The committed prefix advances in ~6 KB chunks. Mounting a whole chunk at once
+// queued a burst of GPU rasterization (0.2-1 s stalls in profiles), so the
+// prefix renders with gentle, budget-capped batching that spreads the mount and
+// raster work across frames.
+const PREFIX_INITIAL_RENDER_BATCH_SIZE = 12
+const PREFIX_RENDER_BATCH_SIZE = 8
+const PREFIX_RENDER_BATCH_DELAY_MS = 12
+const PREFIX_RENDER_BATCH_BUDGET_MS = 4
+const PREFIX_RENDER_BATCH_IDLE_TIMEOUT_MS = 40
 
 const shouldVirtualizeNodes = computed(() => props.virtualizeNodes && !isStreaming.value)
 const resolvedNodeVirtual = computed(() =>
@@ -415,11 +424,11 @@ const renderSegments = computed<RenderSegment[]>(() => {
       nodeVirtual: 'auto',
       maxLiveNodes: STATIC_MAX_LIVE_NODES,
       liveNodeBuffer: STATIC_LIVE_NODE_BUFFER,
-      initialBatch: STATIC_INITIAL_RENDER_BATCH_SIZE,
-      batchSize: STATIC_RENDER_BATCH_SIZE,
-      batchDelay: STATIC_RENDER_BATCH_DELAY_MS,
-      batchBudget: STATIC_RENDER_BATCH_BUDGET_MS,
-      batchIdle: STATIC_RENDER_BATCH_IDLE_TIMEOUT_MS,
+      initialBatch: PREFIX_INITIAL_RENDER_BATCH_SIZE,
+      batchSize: PREFIX_RENDER_BATCH_SIZE,
+      batchDelay: PREFIX_RENDER_BATCH_DELAY_MS,
+      batchBudget: PREFIX_RENDER_BATCH_BUDGET_MS,
+      batchIdle: PREFIX_RENDER_BATCH_IDLE_TIMEOUT_MS,
       parseCoalesce: STATIC_PARSE_COALESCE_MS,
       customId: `${customRendererId.value}::prefix`
     },
--- a/src/renderer/src/features/chat-page/composables/useMessageActions.ts
+++ b/src/renderer/src/features/chat-page/composables/useMessageActions.ts
@@ -84,6 +84,19 @@ export function useMessageActions(options: UseMessageActionsOptions) {
     try {
       activeRetrySessionIds.add(sessionId)
       options.messageStore.clearStreamingState()
+      // The main process truncates from the retried message onward (the user
+      // prompt survives) and re-streams with fresh ids. Mirror that in the UI
+      // BEFORE the IPC: the main starts streaming before the retry IPC resolves,
+      // so an after-await truncation would leave the stale rows visible under
+      // the new stream. When the retried message IS the user prompt, keep it in
+      // place and truncate from the next order sequence instead. The blocked/
+      // failure paths below restore via a reload.
+      const target = options.messageStore.messageCache.get(messageId)
+      if (target) {
+        const fromOrderSeq = target.role === 'user' ? target.orderSeq + 1 : target.orderSeq
+        options.messageStore.truncateMessagesFromOrderSeq(sessionId, fromOrderSeq)
+      }
+
       const result = attachmentFallbackPolicy
         ? await options.sessionClient.retryMessage(sessionId, messageId, {
             attachmentFallbackPolicy
@@ -94,6 +107,9 @@ export function useMessageActions(options: UseMessageActionsOptions) {
           blockedRetryAttempt.value = { sessionId, messageId }
           retryAttachmentPreparationSummary.value =
             result.attachmentPreparation ?? fallbackPreparationSummary()
+          // A blocked retry did not delete anything; bring the optimistically
+          // removed rows back.
+          await options.loadMessagesForSession(sessionId)
         }
         return
       }
--- a/src/renderer/src/features/chat-page/composables/useMessageActions.ts
+++ b/src/renderer/src/features/chat-page/composables/useMessageActions.ts
@@ -41,7 +41,7 @@ type UseMessageActionsOptions = {
   chatClient: ChatClientLike
   beginPlanTurn: (sessionId: string) => void
   clearPlanSnapshotForDeletedMessage: (sessionId: string, messageId: string) => void
-  loadMessagesForSession: (sessionId: string) => Promise<unknown>
+  loadMessagesForSession: (sessionId: string, count?: number) => Promise<unknown>
   applyRestoredSessionSummary: (session: unknown) => void
   currentRestoreRequestId: () => number
   canWriteSessionView: (sessionId: string, requestId: number) => boolean
@@ -92,6 +92,9 @@ export function useMessageActions(options: UseMessageActionsOptions) {
       // place and truncate from the next order sequence instead. The blocked/
       // failure paths below restore via a reload.
       const target = options.messageStore.messageCache.get(messageId)
+      // Optimistic truncation shortens messageIds; remember the pre-truncation
+      // window so a blocked retry can restore the full loaded view.
+      const priorMessageCount = options.messageStore.messageIds.length
       if (target) {
         const fromOrderSeq = target.role === 'user' ? target.orderSeq + 1 : target.orderSeq
         options.messageStore.truncateMessagesFromOrderSeq(sessionId, fromOrderSeq)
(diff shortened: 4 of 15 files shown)


zerob13 (approved):
## Code Review — PR #2202

### Change Summary
Introduces dedicated `PREFIX_*` batch constants for the streaming prefix segment, replacing the reuse of `STATIC_*` constants. Smaller batches (12/8 vs 96/80) with frame delays (12ms vs 0ms) spread GPU rasterization across frames, addressing the 0.2–1s stalls observed during prefix advances.

### Review Criteria

**1. Over-engineering — ✅ None**
Minimal constants-only change: 5 new constants replacing 5 reused ones. No new abstractions, no unnecessary complexity. The `PREFIX_` naming follows the established `STREAM_*`/`STATIC_*` pattern.

**2. Style & Design Consistency — ✅ Consistent**
- Naming convention matches existing `STREAM_*`/`STATIC_*` pattern
- Comment explains the rationale clearly (6KB chunks, GPU raster stalls, frame spreading)
- Values are architecturally coherent: sit between `STREAM_*` (aggressive, live tail) and `STATIC_*` (fast, non-streaming), which matches the prefix being committed-but-still-performance-sensitive content
- Oxfmt-compliant (no semicolons, single quotes, <100 cols)

**3. Destructive Changes — ✅ None**
Purely pacing constants. Tail segment config (`STREAM_*`) and whole-render path (`STATIC_*`) are unchanged. No API, behavior, or data flow changes.

**4. Test Coverage — ✅ Appropriate**
No new tests needed for a constants-only change. Existing 26 tests pass (verified), covering the prefix/tail split behavior, fence-aware splitting, and stream restart. The batch pacing values are internal implementation details not observable through component tests.

**5. Code Logic — ✅ Sound**
- `initialBatch: 12` (was 96): 8x reduction is significant, but virtualization (`nodeVirtual: 'auto', maxLiveNodes: 260`) handles the viewport. During prefix advances, only visible nodes need immediate render; the rest trickle in at 8 nodes/12ms.
- `batchBudget: 4ms`: ~24% of a 60fps frame budget for prefix rendering — reasonable.
- `batchIdle: 40ms` (was 16): less aggressive idle rendering, intentional to reduce pressure during streaming.
- `parseCoalesce: 0` (unchanged): immediate re-parse on content change, rendering is spread by batching. No issue.

### Minor Note
The PR body mentions "0.2–1 s stalls in profiles" (implying profiling was done to identify the problem) but also states "The performance benefit should be confirmed with a streaming profile." A before/after streaming profile would be welcome to validate the specific tuning values, but this is low-risk either way — the change is easily revertible and the worst case is slightly slower prefix rendering.

### Verdict
**Approved.** Clean, well-scoped, low-risk performance optimization. No issues against any review criteria. A streaming profile to validate the tuning would be a nice follow-up but is not blocking.

zhangmo8:
Don’t rush the merge, there’s an optimistic update to retry along with it

zerob13 (changes requested):
## Code Review — PR #2202 (re-review: incremental commit `b43cad7c`)

> Re-review of the new commit `fix(retry): keep user message mounted on retry` (on top of the already-approved `49ce2fdc` prefix-batch change). Reviewed head: `b43cad7c`, incremental diff: +185 / -19, 11 files.

### Change Summary
Reworks retry truncation so the user prompt survives a retry instead of being deleted and recreated as a fresh record. Introduces `retryFromOrderSeq` (truncate-from point distinct from the source user message) and `sourceMessageId` (pre-stream anchor) in `prepareRetryMessage`. The DeepChat path now truncates from the retried assistant message onward and re-anchors the surviving user record; the ACP path is intentionally kept on the old inclusive-truncation behavior with a clear comment. The renderer optimistically mirrors the truncation before the IPC to avoid stale rows under the new stream.

---

### Review Criteria

**1. Over-engineering — ✅ None**
Two new return fields + one new context field, each with a clear purpose. No gratuitous abstraction. The `truncateMessagesFromOrderSeq` store method is single-purpose. The ACP/DeepChat split is justified by genuinely different re-creation semantics.

**2. Style & Design Consistency — ⚠️ Format violations (see blocking issue)**
- Logic and naming are consistent with the codebase (`retryFromOrderSeq`, `preStreamAnchorMessageId` read clearly).
- Comments explain the non-obvious ACP divergence and the optimistic-truncation rationale well.
- **But**: the conditional spread is written multi-line in two files where Oxfmt requires single-line (fits within 100 cols). `pnpm oxfmt --check` fails on both.

**3. Destructive Changes — ✅ None (intentional behavior fix)**
Retry semantics change (user message id now stable instead of recreated), but this is the intended fix. Memory-cursor rewind point shifts accordingly (4→5, tested). ACP path explicitly preserves old behavior. No accidental breakage.

**4. Test Coverage — ✅ Appropriate**
3 new tests (user-message retry truncation in `transcriptMutations`, UI optimistic truncation for both assistant- and user-retry in `useMessageActions`), plus correct updates to existing harness/integration/turn tests. Not excessive.

**5. Code Logic — ✅ Sound**
- `retryFromOrderSeq = target.role === "user" ? target.orderSeq + 1 : target.orderSeq` — correct: keep the prompt when retrying it directly, otherwise drop only the assistant reply onward.
- `preStreamAnchorMessageId` is highest-priority in the anchor resolution but only set during retry, so non-retry flows are unaffected.
- Blocked-retry path correctly reloads to restore optimistically removed rows.
- Minor: the `fromOrderSeq` expression is duplicated between renderer (`useMessageActions`) and main (`transcriptMutations`). Acceptable since the renderer value is an optimistic preview corrected by the post-stream reload, but worth noting if either side changes independently.

---

### 🔴 Blocking — Oxfmt format violations (CI `format:check` will fail)

**`src/main/session/turn.ts`** (lines ~375-377) and **`test/main/agent/deepchat/harness/deepChatAgentHarness.test.ts`** (lines ~1283-1285):

```diff
-        ...(prepared.sourceMessageId
-          ? { preStreamAnchorMessageId: prepared.sourceMessageId }
-          : {}),
+        ...(prepared.sourceMessageId ? { preStreamAnchorMessageId: prepared.sourceMessageId } : {}),
```

The single-line form is 92 columns (< 100), so Oxfmt collapses it. Verified: `pnpm oxfmt --check` reports both files. Fix: run `pnpm oxfmt` on these two files.

### 💬 Notes (non-blocking)

1. **Scope mismatch**: This commit (`fix(retry)`) is unrelated to the PR title (`perf(markdown): gentle batch for prefix advance`). The prefix-batch change was already approved at `49ce2fdc`. Bundling an unrelated retry fix into the same branch makes the PR harder to review and independently revert. Consider a separate PR next time, or at minimum note the added scope in the PR description.

2. **Incomplete mock** — `test/main/session/assignment.test.ts:661`: the `prepareRetryMessage` mock returns `{ content, projectDir }` without the new `retryFromOrderSeq` / `sourceMessageId` fields. Tests pass today because that test only asserts call-occurrence, but the mock no longer matches the real return shape. Add the fields for consistency.

### Verdict
**Changes requested** — solely for the Oxfmt violations (trivial 2-line fix). The retry logic itself is sound, well-tested, and correctly handles the ACP/DeepChat divergence. Once formatted, this is good to go.

zhangmo8:
Addressed the review feedback:

- **Oxfmt violations (blocking)** — ran `pnpm format`; the conditional spread in `src/main/session/turn.ts` and `test/main/agent/deepchat/harness/deepChatAgentHarness.test.ts` is now single-line (92 cols < 100). `format:check` passes.
- **Note 2 (mock shape)** — `test/main/session/assignment.test.ts` `prepareRetryMessage` mock now returns the full `{ content, projectDir, sourceOrderSeq, retryFromOrderSeq, sourceMessageId }` shape.
- **Note 1 (scope)** — PR title and description updated to reflect both parts: the original `perf(markdown)` prefix batching plus the `fix(retry): keep user message on retry` change (user-prompt-preserving retry + compaction budget double-count fix).

zerob13 (approved):
## Code Review — PR #2202 (re-review: commits `25888ee6` → `b5a208bb`)

> Incremental review of 3 new commits since last review (`b43cad7c`).

### Summary

Previous CHANGES_REQUESTED issues are resolved and a new compaction double-count bug is fixed. All 390 tests pass.

### Changes Reviewed

**`25888ee6` — fix(compaction): avoid retry budget double count**

- `compactionService.ts`: adds optional `newUserContentInHistory` param. When true (retry with anchored user prompt), `projectedMessages` is empty — avoids counting the user prompt in both `historyRecords` and `projectedMessages`, which inflated the context estimate and could trigger premature compaction. ✅ Correct and well-scoped.
- `turnCoordinator.ts`: derives `newUserContentInHistory` by checking whether the pre-stream anchor message ID exists in `historyRecords`. Guard is tight (`Boolean(anchorMessageId && historyRecords.some(...))`), non-retry paths unaffected. ✅

**`c3abf50a` — style: collapse retry anchor spread for oxfmt**

- `turn.ts` + `deepChatAgentHarness.test.ts`: collapses the multi-line conditional spread `...(prepared.sourceMessageId ? { ... } : {})` to single-line. This was the 🔴 blocking Oxfmt violation from last review. ✅ Fixed.

**`b5a208bb` — test(session): align prepareRetryMessage mock shape**

- `assignment.test.ts`: adds missing `sourceOrderSeq`, `retryFromOrderSeq`, `sourceMessageId` fields to the `prepareRetryMessage` mock return value, aligning it with the real interface shape introduced in `b43cad7c`. ✅ Necessary fix.
- `deepChatAgentHarness.test.ts`: adjusts the retry-under-compaction-pressure assertion — no longer checks `generateText` was called; instead checks `run.messages` contains the compaction checkpoint (`Persisted Rolling Summary`). The comment explains the rationale: retry keeps the user prompt in-transcript, so pre-stream compaction already fits the budget without further runtime compaction. ✅ Reasonable.

### Checklist

| Criterion | Result |
|---|---|
| Over-engineering | No — minimal targeted fix |
| Consistency with DeepChat patterns | Yes — follows existing optional-param + conditional-spread conventions |
| Breaking changes | None — `newUserContentInHistory` is optional, default behavior unchanged |
| Excessive test additions | No — only adjusts existing assertions, no new test bloat |
| Logic correctness | Verified — 390 tests pass, scope and guards are sound |

### Verdict

✅ **APPROVED** — previous blocking issues resolved, new compaction fix is correct and well-tested.

src/main/agent/deepchat/runtime/turnCoordinator.ts
```diff
@@ -819,6 +829,7 @@ export class TurnCoordinator {
               preserveEmptyInterleavedReasoning:
                 interleavedReasoning.preserveEmptyReasoningContent === true,
               newUserContent: content,
+              newUserContentInHistory,
```
yyhhyyyyyy, line 832: **P1:** `newUserContentInHistory` prevents double-counting only during compaction. The final context build still appends `newUserContent` while `historyRecords` already contains the anchored prompt, so a normal retry sends the same user message twice.

src/main/session/transcriptMutations.ts
```diff
@@ -62,14 +68,25 @@ export class SessionTranscriptMutations {
...
+    return {
+      content,
+      projectDir,
+      sourceOrderSeq: sourceUserMessage.orderSeq,
+      // Truncate from the retried message (not the source user message) so the
+      // user prompt survives and only the assistant reply is regenerated. When
+      // the retried message IS the user prompt, keep it too and truncate from
+      // the next order sequence, anchoring it so the same record is re-sent
+      // instead of a fresh duplicate.
+      retryFromOrderSeq: target.role === 'user' ? target.orderSeq + 1 : target.orderSeq,
```
yyhhyyyyyy, line 88: **P2:** If this is a failed Steer user row (`status: 'error'`), preserving and reusing it never restores `status: 'sent'`. Future context filtering then drops the retried prompt.

src/renderer/src/features/chat-page/composables/useMessageActions.ts
```diff
@@ -94,6 +107,9 @@ export function useMessageActions(options: UseMessageActionsOptions) {
           blockedRetryAttempt.value = { sessionId, messageId }
           retryAttachmentPreparationSummary.value =
             result.attachmentPreparation ?? fallbackPreparationSummary()
+          // A blocked retry did not delete anything; bring the optimistically
+          // removed rows back.
+          await options.loadMessagesForSession(sessionId)
```
yyhhyyyyyy: **P2:** This reload derives its count from `messageIds.length`, but optimistic truncation already shortened that array. With more than 100 loaded messages, a blocked retry restores only the latest 100 and loses the older loaded window.

src/renderer/src/components/markdown/MarkdownRenderer.vue
```diff
@@ -415,11 +424,11 @@ const renderSegments = computed<RenderSegment[]>(() => {
       nodeVirtual: 'auto',
       maxLiveNodes: STATIC_MAX_LIVE_NODES,
       liveNodeBuffer: STATIC_LIVE_NODE_BUFFER,
-      initialBatch: STATIC_INITIAL_RENDER_BATCH_SIZE,
-      batchSize: STATIC_RENDER_BATCH_SIZE,
-      batchDelay: STATIC_RENDER_BATCH_DELAY_MS,
-      batchBudget: STATIC_RENDER_BATCH_BUDGET_MS,
-      batchIdle: STATIC_RENDER_BATCH_IDLE_TIMEOUT_MS,
+      initialBatch: PREFIX_INITIAL_RENDER_BATCH_SIZE,
```
yyhhyyyyyy, line 430: **P3:** These batch settings are inactive while `maxLiveNodes` is 260. In `markstream-vue@2.0.1-beta.1`, incremental batching requires `maxLiveNodes <= 0`, so the intended raster spreading does not occur.

src/renderer/src/stores/ui/message.ts
```diff
@@ -863,6 +863,30 @@ export const useMessageStore = defineStore('message', () => {
...
+      return Boolean(record && record.orderSeq >= fromOrderSeq)
+    })
+    if (removedIds.length === 0) return
+
+    markLiveMessageViewMutation(sessionId)
+    for (const id of removedIds) {
+      messageCache.value.delete(id)
+      parsedMessageCache.delete(id)
+    }
+    messageIds.value = messageIds.value.filter((id) => !removedIds.includes(id))
```
yyhhyyyyyy: **P3:** `removedIds.includes(id)` makes truncation O(n²). This store can retain thousands of loaded messages, so retrying an old turn can block the renderer; use a `Set` for membership.

yyhhyyyyyy (commented):
Inline findings, ordered by severity.

zerob13 (changes requested):
## Code Review — PR #2202 (4th review: re-review after @yyhhyyyyyy findings)

> Re-review triggered by @yyhhyyyyyy's inline findings. Previous review was APPROVED at `b5a208bb` — that approval is **withdrawn**. All findings below have been independently verified against the code at `b5a208bb`.

---

### 🔴 P1 (Blocking) — Retry sends the user message twice

**File:** `src/main/agent/deepchat/runtime/turnCoordinator.ts` / `contextBuilder.ts`

**Verified:** In the retry path, `prepareRetryMessage` anchors the source user message so it survives `commitRetryMessage` (truncation starts at `retryFromOrderSeq`, which is `> userMessage.orderSeq`). The user message therefore remains in `historyRecords` with `status: 'sent'`.

When `buildCacheAwareContextWithMetadata` runs:

1. `historyRecords` (containing the anchored user message) → `buildHistoryTurns` → `flattenTurns(selectedTurns)` — **includes the user message**
2. `newUserContent: content` (extracted from the same source user message via `extractUserMessageInput`) → `createUserChatMessage` → `newUserMessage` — **also includes the user message**

Final `messages` array:
```
[...leadingMessages, ...flattenTurns(selectedTurns), newUserMessage]
```

Both `flattenTurns(selectedTurns)` and `newUserMessage` carry the same user content. The LLM receives the prompt **twice**.

`newUserContentInHistory: true` only prevents double-counting in **compaction** (`projectedMessages` is emptied). It does **not** prevent the final context build from appending `newUserContent` on top of a `historyRecords` array that already contains the anchored prompt.

**Fix direction:** When `newUserContentInHistory` is true, `buildCacheAwareContextWithMetadata` (or the caller) must skip creating/appending `newUserMessage` — the history already carries it. Alternatively, exclude the anchored user record from `historyRecords` before passing to the context builder (but then ensure compaction still sees it for budget estimation).

---

### ⚠️ P2 — Steer error retry: user message status never restored

**File:** `src/main/session/transcriptMutations.ts:80`

**Verified:** `isContextHistoryRecord` (`contextBuilder.ts:378`) includes user records only when `status === 'sent'`. After a Steer error retry, the anchored user row retains `status: 'error'` — `prepareRetryMessage` never transitions it back to `'sent'`. Consequently:

- The retried prompt is absent from `historyRecords` (filtered out by `isContextHistoryRecord`), so P1 does not fire in this specific path.
- But any **future** non-retry turn's context build also drops this user message from history, silently losing the prompt from the conversation context.

**Fix direction:** After anchoring, restore the user record's status to `'sent'` (or adjust `isContextHistoryRecord` to include retried user rows).

---

### ⚠️ P2 — Blocked retry reload may lose older loaded messages

**File:** `src/renderer/src/features/chat-page/composables/useMessageActions.ts:112`

**Plausible (not fully verified):** When a retry is blocked (`accepted === false`), `loadMessagesForSession` is called to restore optimistically truncated rows. If this loader has a default limit (e.g. 100) and optimistic truncation already shortened `messageIds`, the reload may only fetch the latest N messages, permanently losing the older loaded window from the renderer cache. Please confirm the load limit and whether it accounts for the pre-truncation count.

---

### ⚠️ P3 — Prefix batch settings may be inactive with `maxLiveNodes: 260`

**File:** `src/renderer/src/components/markdown/MarkdownRenderer.vue:427`

**Needs author confirmation:** @yyhhyyyyyy notes that in `markstream-vue@2.0.1-beta.1`, incremental batching requires `maxLiveNodes <= 0`. The prefix segment uses `maxLiveNodes: STATIC_MAX_LIVE_NODES` (260), which may cause the `initialBatch`/`batchSize`/`batchDelay` settings to be silently ignored — defeating the PR's core performance goal. Please verify against the markstream-vue 2.0.1-beta.1 source whether batch settings take effect when `maxLiveNodes > 0`.

---

### ⚠️ P3 — `removedIds.includes(id)` is O(n²)

**File:** `src/renderer/src/stores/ui/message.ts:886`

**Verified:**
```ts
messageIds.value = messageIds.value.filter((id) => !removedIds.includes(id))
```
`Array.includes` is O(n) inside `Array.filter` (also O(n)), making this O(n²). With thousands of loaded messages, retrying an old turn can block the renderer. Use a `Set` for O(1) membership:
```ts
const removedSet = new Set(removedIds)
messageIds.value = messageIds.value.filter((id) => !removedSet.has(id))
```

---

### Review Checklist

| Criterion | Result |
|---|---|
| Over-engineering | No — the retry anchor approach is sound in concept, but the implementation has a gap between compaction and context build |
| Consistency with DeepChat patterns | Mostly — naming and structure follow conventions; the O(n²) filter is inconsistent with perf-conscious patterns elsewhere |
| Breaking changes | ⚠️ Yes — P1 silently changes retry behavior (duplicate user message sent to LLM). P2 silently loses context in Steer error paths |
| Excessive test additions | No — test adjustments are appropriate |
| Logic correctness | ❌ P1 confirmed: duplicate user message. P2 confirmed: status not restored. P3 confirmed: O(n²) |

### Verdict

**Changes requested.** The previous APPROVED is withdrawn. P1 is a blocking bug that sends the user prompt twice on every normal retry. P2 has silent context-loss consequences. These must be addressed before merge.

zhangmo8:
Addressed the findings from @yyhhyyyyyy / @zerob13's 4th review:

**🔴 P1 — retry sends the user message twice**
-  / : when retry reuses an anchored user prompt that is already in history (), the context build now excludes that record from the history turns (), so it appears only once as the live  (which carries memory injection). Compaction budget behavior is unchanged.
- Added a harness regression test (`does not duplicate a retried user prompt in the assembled request`).

**⚠️ P2 — Steer error retry status never restored**
- `transcriptMutations.ts`: `prepareRetryMessage` now restores the anchored user row's status to `'sent'` when it isn't already, so context history filtering keeps the retried prompt. Covered by an updated steer test.

**⚠️ P2 — Blocked retry reload loses older messages**
- `useMessageActions.ts`: records the pre-truncation message count and passes it to the reload on a blocked retry, restoring the full visible window instead of deriving the count from the already-shortened `messageIds`.

**P3 — `truncateMessagesFromOrderSeq` O(n²)**
- `message.ts`: membership now uses a `Set`.

**P3 — Prefix batching inactive while `maxLiveNodes` is 260**
- Verified in markstream-vue@2.0.1-beta.1 that incremental batching is gated on `maxLiveNodes <= 0`. `MarkdownRenderer.vue` prefix segment now sets `maxLiveNodes: 0` (tail already does this) so the gentle PREFIX_* batching actually applies.

All main + renderer suites pass (harness 370, session, renderer), typecheck/lint/format green.

zhangmo8:
Addressed the findings from @yyhhyyyyyy / @zerob13's 4th review:

**P1 — retry sends the user message twice**
- `contextBuilder.ts` / `turnCoordinator.ts`: when retry reuses an anchored user prompt already present in history (`newUserContentInHistory`), the context build now excludes that record from the history turns (`newUserContentMessageId`), so it appears only once as the live `newUserMessage` (which carries memory injection). Compaction budget behavior is unchanged.
- Added a harness regression test: `does not duplicate a retried user prompt in the assembled request`.

**P2 — Steer error retry status never restored**
- `transcriptMutations.ts`: `prepareRetryMessage` now restores the anchored user row's status to `sent` when it isn't already, so context history filtering keeps the retried prompt. Covered by an updated steer test.

**P2 — Blocked retry reload loses older messages**
- `useMessageActions.ts`: records the pre-truncation message count and passes it to the reload on a blocked retry, restoring the full visible window instead of deriving the count from the already-shortened `messageIds`.

**P3 — truncateMessagesFromOrderSeq O(n²)**
- `message.ts`: membership now uses a `Set`.

**P3 — Prefix batching inactive while maxLiveNodes is 260**
- Verified in markstream-vue@2.0.1-beta.1 that incremental batching is gated on `maxLiveNodes <= 0`. `MarkdownRenderer.vue` prefix segment now sets `maxLiveNodes: 0` (tail already does this) so the gentle `PREFIX_*` batching actually applies.

All main + renderer suites pass (harness 370, session, renderer), typecheck/lint/format green.

zerob13 (approved):
## Code Review — PR #2202 (5th review: incremental commits `b5a208bb` → `2994e721`)

> Incremental review of 3 new commits since the 4th review (`b5a208bb`, CHANGES_REQUESTED). All 5 findings from the 4th review have been addressed.

### Changes Reviewed

- `54565ba6` — fix(retry): address review findings on retry context
- `017ebf08` — fix(markdown): enable prefix gentle batching
- `2994e721` — fix(retry): avoid duplicate pinned prompt on retry

---

### Finding Resolution

#### 🔴 P1 (was Blocking) — Retry sends the user message twice → ✅ Resolved

**Fix:** `turnCoordinator.ts` now derives `newUserContentInHistory` + `newUserContentMessageId` from the pre-stream anchor. `contextBuilder.ts` excludes the anchored user record from `historyRecords` (so it no longer appears in `flattenTurns(selectedTurns)`), leaving it only in `newUserMessage`.

Verified: final `messages = [...leadingMessages, ...flattenTurns(selectedTurns), newUserMessage]` — anchored user appears exactly once (in `newUserMessage`). Compaction budget remains consistent: `newUserContentInHistory=true` empties `projectedMessages`, and the anchored user's tokens are counted once via `newUserMessage` in `fixedTokens`.

Non-retry path unaffected: `anchorMessageId` is `undefined` → `newUserContentMessageId` is `undefined` → `.filter(id !== undefined)` is a no-op → behavior identical to before.

Regression tests added: `does not duplicate a retried user prompt` + `does not duplicate a pinned first user prompt on retry`. ✅

#### ⚠️ P2 — Steer error retry: user status never restored → ✅ Resolved

**Fix:** `transcriptMutations.ts` restores `sourceUserMessage.status` to `'sent'` when it's not already, before returning from `prepareRetryMessage`. This ensures `isContextHistoryRecord` includes the anchored user in future non-retry context builds.

Test assertion added in `transcriptMutations.test.ts`. ✅

#### ⚠️ P2 — Blocked retry reload may lose older loaded messages → ✅ Resolved

**Fix:** `useMessageActions.ts` captures `priorMessageCount = messageStore.messageIds.length` before optimistic truncation, then passes it to `loadMessagesForSession(sessionId, priorMessageCount)` on the blocked path. `loadMessages` correctly uses `desiredCountOverride` to restore the pre-truncation window. ✅

#### ⚠️ P3 — Prefix batch settings inactive with `maxLiveNodes: 260` → ✅ Resolved

**Fix:** `MarkdownRenderer.vue` prefix segment now uses `maxLiveNodes: 0, liveNodeBuffer: 0`, allowing markstream-vue's incremental batching to take effect (gated on `maxLiveNodes <= 0`). The `initialBatch: 12 / batchSize: 8 / batchDelay: 12ms` settings now actually spread GPU rasterization across frames.

**Note (non-blocking):** `maxLiveNodes: 0` disables virtual live-node limiting for the prefix segment, so all prefix DOM nodes stay mounted. This is the necessary trade-off given markstream-vue's gating, and prefix content is committed/static (bounded by compaction), so the risk is acceptable. Worth monitoring if very long prefixes cause memory pressure.

#### ⚠️ P3 — `removedIds.includes(id)` is O(n²) → ✅ Resolved

**Fix:** `message.ts` `truncateMessagesFromOrderSeq` now builds a `Set` for O(1) membership. ✅

---

### Pinned-first-user edge case (commit `2994e721`)

When the retried prompt is itself the pinned first user, `emittedPinnedFirstUser` is set to `null` (dropped from `leadingMessages`), and the record is excluded from `historyRecords`. The prompt survives only in `newUserMessage`. The `buildCacheAwareOverflowError` still reports `includesPinnedFirstUser: true` using the original `pinnedFirstUser` — correct, since the pinned content is still conceptually present (just delivered via the live turn). ✅

### Checklist

| Criterion | Result |
|---|---|
| Over-engineering | No — each fix is targeted, no new abstractions |
| Consistency with DeepChat patterns | Yes — naming, comments, and structure follow conventions |
| Breaking changes | None — all changes are bug fixes; non-retry paths verified unaffected |
| Excessive test additions | No — 2 regression tests for P1, 1 assertion for P2, mock alignment only |
| Logic correctness | Verified — all 5 findings resolved, budget estimation consistent, edge cases handled |

### Verdict

✅ **Approved.** All blocking issues from the 4th review are correctly resolved. The P1 fix is sound: anchored user excluded from history turns, present only in `newUserMessage`, with compaction budget staying consistent. The pinned-first-user edge case (commit `2994e721`) is correctly handled via `emittedPinnedFirstUser`. Tests are appropriate and CI is green.

PR #2207, opened 2026-08-25
fix(startup): unblock interactive startup tasks

zhangmo8:
## Summary

`StartupWorkloadCoordinator` caps io concurrency at 2 and never preempts a running task. Long-running background io tasks (MCP / remote-runtime init) used to occupy every lane, so the interactive `main.bootstrap` task queued for ~15s even though its body only needs ~5ms. Since the renderer's first paint is gated on bootstrap, this delayed the whole startup.

Interactive-phase tasks are now admitted even when a resource lane is at capacity, while still counting toward `runningCounts` — so `isIdle()`, cancellation, and dedupe semantics stay intact. Background/deferred tasks keep the existing `cpu=1`/`io=2` caps.

## Verification

- Measured in a real run: `main.bootstrap` queued for ~15s before the fix, 0ms after (its body is ~5ms).
- Regression test added: interactive io work starts while background lanes are fully occupied; confirmed it fails on the old code and passes on the fix.
- `pnpm typecheck:node` / `pnpm typecheck:web` pass
- CI: `static`, `test-main`, `test-renderer`, `test-native-memory`, `build` pass

### DIFF
--- a/src/main/app/startupWorkloadCoordinator/index.ts
+++ b/src/main/app/startupWorkloadCoordinator/index.ts
@@ -292,7 +292,14 @@ export class StartupWorkloadCoordinator {
     })
 
     for (const task of sortedPending) {
-      if (this.runningCounts[task.resource] >= MAX_CONCURRENCY[task.resource]) {
+      // Interactive work is short and latency-critical (startup bootstrap, first session
+      // page). The concurrency caps protect the event loop / disk from heavy background
+      // work; when long-running background io tasks (MCP / remote runtime init) hold every
+      // lane, an interactive task would otherwise starve and delay the first paint. Admit
+      // interactive tasks even at capacity — they still count toward runningCounts, so
+      // isIdle()/cancel/dedupe semantics stay intact.
+      const atCapacity = this.runningCounts[task.resource] >= MAX_CONCURRENCY[task.resource]
+      if (atCapacity && task.phase !== 'interactive') {
         continue
       }
 
--- a/test/main/app/startupWorkloadCoordinator.test.ts
+++ b/test/main/app/startupWorkloadCoordinator.test.ts
@@ -278,4 +278,68 @@ describe('StartupWorkloadCoordinator', () => {
 
     expect(onFailure).not.toHaveBeenCalled()
   })
+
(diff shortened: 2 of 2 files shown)


zerob13 (approved):
## Review — APPROVED ✅

Reviewed head: `2369f1931` (1 commit, +72/-1, 2 files)

### Change Summary

Modifies `StartupWorkloadCoordinator.pickNextTask()` to let `interactive` phase tasks bypass the resource concurrency cap (`cpu=1`, `io=2`). Previously, when background io tasks (MCP init, remote runtime init) occupied both io lanes, interactive tasks (bootstrap, first session page) would starve and delay first paint.

### Assessment

| Criterion | Result |
|---|---|
| Over-engineering | No — 2-line logic change, no unnecessary abstraction |
| Style / pattern consistency | Yes — matches existing code style (Oxfmt, single quotes, no semicolons) |
| Destructive changes | No — soft behavioral change, `runningCounts` still tracks interactive tasks, `isIdle()/cancel/dedupe` semantics intact |
| Excessive tests | No — 1 targeted regression test with proper deferred gates and 1000ms fail-fast timeout |
| Logic correctness | Verified — interactive tasks still increment/decrement `runningCounts`; `isIdle()` check unchanged; all 7 tests pass; typecheck clean |

### Detailed Analysis

**Interactive task assumption holds.** Checked all 3 interactive task callers:
- `main.bootstrap:route` — one-shot route handler (session lookup, agents, ACP, workspace)
- `main.session.firstPage:route` — one-shot session list fetch
- `settings.providers.summary:route` — one-shot provider summary fetch

All are short, latency-critical route handlers the renderer blocks on for first paint. Bypassing the cap is the right call — they are genuinely short and few.

**No unbounded concurrency risk in practice.** Interactive tasks are rare (3 call sites, all route handlers) and complete quickly. The `runningCounts` decrement in the `finally` block ensures the count returns to normal after each interactive task completes.

**Test quality.** The new test fills both io lanes with gated background tasks, schedules an interactive task, and verifies it starts within 1000ms while background tasks are still blocked. Uses the existing `createDeferred` helper and follows the same pattern as other tests in the file.

### Verdict

Clean, minimal, well-documented fix for a real startup latency problem. No changes requested.

PR #2213, opened 2026-08-26
fix(chat): release stuck pending permission approvals (#2210)

### LINKED ISSUE #2210: [BUG] 长时间不回复权限审批后，审批页面卡死无法关闭
### System details / 系统信息

英特尔x7 358H，win11家庭版，V1.1.0

### What's wrong? / 出了什么问题？

1.选择deepchat agent正常给大模型下任务
2.大模型给出权限审批页面
3.超过5分钟不回复，模型继续工作，但是权限审批页面就一直卡在那，新审批会覆盖在上面，但是当允许/拒绝之后会依旧显示旧的未处理的审批页面，图片里是卡住的审批页面实例，卡住后，无论点击什么按钮都无法关掉审批页面

Image

Image

zhangmo8:
## Problem

Closes #2210: 权限审批页面在长时间未回复后卡死无法关闭。

When a pending tool permission / question interaction is no longer backed by a
running generation (the run ended, the session was replaced mid-response, or
the subagent session vanished), the main process cannot resolve its action
block. The renderer derives the approval UI from the persisted block
(`status === 'pending'` + `needsUserAction !== false`), so the approval stays
shown as the first pending interaction forever. No button closes it, newer
approvals queued behind it are never reached, and the orphaned block keeps
blocking new turns.

## Fix

Two layers:

**Renderer** (`useToolInteraction`): after responding to a pending interaction
we reload the session. If that interaction is **still pending after the reload**
(the main process did not resolve it) — or the respond was rejected — it is a
stale interaction: drop it from the pending list so the UI advances to the next
approval, and ask the main process to dismiss it durably.

**Main process** (new `chat.dismissToolInteraction` route): marks the orphaned
`tool_call_permission` / `question_request` block as resolved (`denied` /
`success`, `needsUserAction: false`) and persists it in the transcript with
`interactionResolution: 'cancelled'`. No tool executes and no run resumes. This
removes the lingering block so it no longer reappears after a session reload or
blocks new turns.

Normal resolutions, `handledInline` (skill-draft view), and
`waitingForUserMessage` (question_other follow-up) are left untouched.

## BEFORE

```
 pending approval (stale, run gone)  ← always shown
 └─ Allow / Deny  → nothing happens, page stuck
 (newer approvals queued but never reached)
 new message      → blocked: "resolve pending interactions first"
```

## AFTER

```
 pending approval (stale, run gone)  → respond → dropped + dismissed durably
 next pending approval now shown
 new message      → allowed (orphaned block resolved in transcript)
```

## Tests

- `test/renderer/.../useToolInteraction.test.ts`
  - stale interaction dropped + dismissed after a non-resolving respond
  - rejected respond refreshes + drops the unresolvable interaction
  - resolved interaction is kept (no dismiss)
- `test/main/.../deepChatAgentHarness.test.ts` (`dismissToolInteraction`)
  - resolves a stale pending permission / question block in the transcript
  - stamps `interactionResolution: 'cancelled'`
  - returns false for a non-pending interaction without writing

Verified: format / lint / i18n / typecheck (node + web) / full main harness
(375) / session+manager (702) / full renderer suite (266 files, 2355 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool interactions that require a user response.
  * Automatically refreshes conversations after failed or rejected tool responses.
  * Permanently dismisses stale permission and question prompts after refresh.
  * Prevents outdated prompts from blocking conversations or resuming runs unexpectedly.
  * Preserves successfully resolved interactions and unrelated pending prompts.
  * Ensures dismissed prompts are recorded correctly, including when the runtime is unavailable.

### DIFF
--- a/src/main/agent/deepchat/harness/deepChatAgentHarness.ts
+++ b/src/main/agent/deepchat/harness/deepChatAgentHarness.ts
@@ -228,6 +228,13 @@ export class DeepChatAgentHarness
     )
   }
 
+  readonly dismissToolInteraction = (
+    sessionId: string,
+    messageId: string,
+    toolCallId: string
+  ): Promise<boolean> =>
+    this.services.interactionCoordinator.dismiss(sessionId, messageId, toolCallId)
+
   async setPermissionMode(sessionId: string, mode: PermissionMode): Promise<void> {
     await this.services.sessionSettings.setPermissionMode(sessionId, mode)
   }
@@ -275,11 +282,10 @@ export class DeepChatAgentHarness
 
   async cancelGeneration(sessionId: string): Promise<void> {
     const instance = this.services.runtime.getHydratedScope(toAppSessionId(sessionId))?.instance
-    if (instance) {
+    if (instance)
       try {
         this.services.toolSurfaceDiagnostics.cancelPending(instance)
       } catch {}
-    }
     await this.services.runLifecycle.cancel(sessionId)
   }
 
@@ -288,9 +294,7 @@ export class DeepChatAgentHarness
   }
 
   async cancelGenerationByEventId(sessionId: string, eventId: string): Promise<boolean> {
-    if (this.services.runLifecycle.getActiveGeneration(sessionId)?.eventId !== eventId) {
-      return false
-    }
+    if (this.services.runLifecycle.getActiveGeneration(sessionId)?.eventId !== eventId) return false
     await this.cancelGeneration(sessionId)
     return true
   }
--- a/src/main/agent/deepchat/runtime/interactionCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/interactionCoordinator.ts
@@ -747,6 +747,51 @@ export class InteractionCoordinator {
     }
   }
 
+  /**
+   * Closes a stale pending interaction (permission/question) whose backing run
+   * can no longer resolve it. The block is marked denied/resolved in the
+   * transcript so the approval does not linger across session reloads or block
+   * new turns. No tool executes and no run resumes.
+   */
+  async dismiss(sessionId: string, messageId: string, toolCallId: string): Promise<boolean> {
+    const message = await this.ports.messageStore.getMessage(messageId)
+    if (!message || message.role !== 'assistant' || message.sessionId !== sessionId) {
+      return false
+    }
+
+    const blocks = parseAssistantBlocks(message.content)
+    const entry = collectPendingInteractionEntries(messageId, blocks).find(
+      ({ interaction }) => interaction.toolCallId === toolCallId
+    )
+    if (!entry) {
+      return false
+    }
+
+    const actionBlock = blocks[entry.blockIndex]
+    if (actionBlock.action_type === 'tool_call_permission') {
+      const permissionType = parsePermissionPayload(actionBlock)?.permissionType ?? 'write'
+      markPermissionResolved(actionBlock, false, permissionType)
+      updateToolCallResponse(blocks, toolCallId, 'User denied the request.', true)
+    } else if (actionBlock.action_type === 'question_request') {
+      markQuestionResolved(actionBlock, '')
+    } else {
+      return false
+    }
+
+    const metadata = stampInteractionResolution(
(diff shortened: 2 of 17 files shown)


zerob13 (approved):
## Review: fix(chat): release stuck pending permission approvals (#2210)

**APPROVED** ✅

Reviewed merge-base diff (`origin/dev...HEAD`), 2 files (+136/-9).

### Problem
When a tool interaction (permission approval / question request) becomes stale — backing run is gone, session replaced mid-response, or subagent session vanished — the approval UI gets stuck on an interaction that can never close, blocking access to newer pending approvals.

### Core Logic Assessment

**`staleInteractionKeys` Set + filtering** — Clean, minimal solution:
- Immutable Set replacement (`new Set(prev).add(key)`) correctly triggers Vue reactivity
- `interactionKey` composite (`sessionId:messageId:toolCallId`) is unique per tool call — no collision risk
- Filtering applied at the end of `pendingInteractions` computed with an early-return fast path when the set is empty

**`onToolInteractionRespond` refactor** — Well-structured:
- Extracted `isCurrentView`, `refreshAfterResponse`, `isStillPending` helpers improve readability without over-abstracting
- Success path: reload → if still pending after reload, `markStale` (backing run is gone)
- Error path: reload → if still pending, `markStale` (rejected respond = strong staleness signal)
- `handledInline` / `waitingForUserMessage` early-exit before stale check — correct, the interaction is being actively handled
- `finally` block always releases the lock

**`waitingForUserMessage` type addition** — Correct:
- Field already exists in `ToolInteractionResultSchema` (`src/shared/contracts/common.ts:394`) and route output schema
- Main process `chatService.respondToolInteraction` already returns it (`chatService.ts:279`)
- Renderer `ChatClientLike` type was just incomplete — this PR fixes the gap

**Race condition handling** — Sound:
- `isCurrentView()` guards prevent stale-marking when the user has switched sessions
- If session changes during refresh, `refreshAfterResponse` returns `false` and the stale check is skipped

### Observations (non-blocking)

1. **`staleInteractionKeys` never clears**: The set grows with each stale interaction but is bounded by tool-call count per session and keys are short strings. Not a real concern for typical usage.

2. **Behavioral change in error path**: Previously, a rejected `respondToolInteraction` would not trigger a reload. Now it reloads to check if the interaction is still pending. This is intentional and correct — a rejected respond is itself a staleness signal. The existing test was appropriately updated to reflect this.

### Test Coverage

- Updated test: rejection path now verifies stale-dropping + reload behavior
- New test: responded interaction stays pending → dropped, next pending surfaces
- New test: responded interaction resolved by main process → not marked stale

3 test cases (+81 lines) — proportionate for the behavioral change, no excessive coverage.

### Verdict

No over-engineering, no unexplained breaking changes, no excessive tests. Clean bugfix that follows deepchat conventions (Vue reactivity, composable patterns, structural typing). Code passes typecheck, lint, and all tests.

> Note: GitHub prevents self-approval; if blocked, verdict stands as comment.

zhangmo8:
Addressed the CodeRabbit review feedback in commit `037d54df8`:

**Round 1 (renderer)**
- ✅ `ChatClientLike.respondToolInteraction` now uses the shared `ToolInteractionResult` type instead of the inline return shape.
- ✅ Wrapped the long `it(...)` test declaration to ≤100 columns (renamed to keep the single-line declaration within the limit, matching Oxfmt's canonical form).

**Round 2 (main process)**
- ✅ `InteractionCoordinator.dismiss` for a dismissed `question_request` now also resolves the paired `tool_call` block via `updateToolCallResponse`.
- ✅ When the dismissed message has no remaining pending interactions, it is finalized (`status: 'sent'`, `interactionResolution: 'cancelled'`) instead of staying `pending` after reload.
- ✅ Dismissal now only removes the dismissed entry from the instance queue, preserving pending interactions from other messages (with a regression test).
- ✅ ACP `dismiss` now persists a durable denied block when no live ACP runtime can resolve the request (with a regression test).

All checks green: format / lint (0 warnings) / i18n / typecheck (node+web) / main harness+session+manager (1078) / ACP backend (18) / full renderer suite (266 files, 2355 tests).

PR #2222, opened 2026-08-31
fix(agent): honor stop and code timeouts

### LINKED ISSUE #2221: [BUG] 模型工作到一半突然停止工作，无法强制停止或自行继续
### System details / 系统信息

v1.1.1-beta.5

### What's wrong? / 出了什么问题？

模型工作到一半突然停止工作，卡在那一步半小时，无法强制停止或自行继续，也没有给出报错之类的玩意
以下是具体函数参数与截图
函数参数：{"code":"const expression = `new Promise((resolve) => {\n  const wait = (ms) => new Promise((done) => setTimeout(done, ms));\n  (async () => {\n    await wait(650);\n    const chart = document.querySelector('#chart');\n    const snapshot = (label) => ({\n      label,\n      ready: document.readyState,\n      theme: document.documentElement.dataset.theme,\n      title: document.querySelector('#title')?.textContent,\n      activeLength: document.querySelector('[data-input-length].on')?.dataset.inputLength,\n      activeMetric: document.querySelector('[data-metric].on')?.dataset.metric,\n      chartChildren: chart?.childElementCount,\n      paths: chart?.querySelectorAll('.path').length,\n      points: chart?.querySelectorAll('.hit').length,\n      switching: chart?.classList.contains('chart-switching'),\n      entering: chart?.classList.contains('chart-entering'),\n      busy: chart?.getAttribute('aria-busy') || null\n    });\n    const initial = snapshot('initial');\n    document.querySelector('[data-input-length=\"512\"]').click();\n    const switching = snapshot('immediate-after-length-click');\n    await wait(80);\n    const entering = snapshot('after-animation-frame');\n    await wait(980);\n    const settled = snapshot('after-animation-settled');\n    document.querySelector('[data-metric=\"input\"]').click();\n    const metricSwitching = snapshot('immediate-after-metric-click');\n    await wait(1200);\n    const metricSettled = snapshot('metric-settled');\n    resolve({initial,switching,entering,settled,metricSwitching,metricSettled,externalCss:Array.from(document.querySelectorAll('link[rel=\"stylesheet\"]')).length,externalJs:Array.from(document.scripts).filter((script) => script.src).length,styleTags:document.querySelectorAll('style').length,inlineScripts:document.querySelectorAll('script:not([src])').length});\n  })();\n})`;\nconst result = await tools.cdp_send({method:'Runtime.evaluate',params:{expression,awaitPromise:true,returnByValue:true}});\nreturn result;","description":"验证单文件图表动画和资源内联"}
Image

zerob13:
Provider streams and browser CDP calls can remain pending after cancellation, preventing Stop from settling the local run. The fixed five-minute Code Mode deadline also prevents legitimate longer operations from completing.

This change makes provider iterator waits, browser readiness and CDP waits, and utility-process startup observe cancellation. Provider cleanup cannot block cancellation, abandoned utility processes are reclaimed, and late results cannot dispatch more work, change the session store, or report successful browser activity.

- Add optional `run_code.timeout_ms`, defaulting to `300000` ms, and the equivalent Code Mode `exec` pragma.
- Document in the schema, tool descriptions, and SDK prompts that routine calls omit the parameter; the model supplies a larger value only when it expects an operation to exceed five minutes.
- Validate timeout values before dispatch. The limit applies to one cell, does not reset on yield or continuation, and does not extend individual subtool timeouts.
- Keep explicit cancellation independent of the selected duration, without adding an idle timeout or a whole-run deadline.

Cancellation ends local execution and waiting; it does not roll back browser or external operations already dispatched. No UI layout or dependency changes.

Validation passed:

- `pnpm run format`, `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, and `pnpm run typecheck`.
- 407 main-process tests across 13 relevant suites, plus all three ChatPage stop-request tests.
- Regression coverage includes a ten-minute cell timeout, silent provider/CDP cancellation, startup races, and late completion handling. Timing cases use fake clocks and browser integration cases use mocked Electron boundaries.

Closes #2221

## Summary by CodeRabbit

- **New Features**
  - Added optional execution timeouts for Code Mode, configurable through `run_code` and `exec`.
  - Timeouts default to five minutes and support validated custom values up to approximately 24.8 days.
  - Added clear guidance describing timeout behavior, limits, and cancellation precedence.

- **Bug Fixes**
  - Improved cancellation for stalled code execution, browser actions, and provider responses.
  - Prevented late results from continuing after cancellation or triggering unintended actions.
  - Ensured browser operations stop promptly when canceled.

### DIFF
--- a/docs/features/experimental-tool-modes/spec.md
+++ b/docs/features/experimental-tool-modes/spec.md
@@ -40,8 +40,8 @@ Agent Mode 是兼容基线。工具 registry、disabled-tool 配置、权限、�
 Code Mode 保留当前已启用工具的能力，但不把这些工具逐个发送给模型：
 
 - `openai-codex` 路由使用 Codex 风格的 raw JavaScript `exec` 和续跑 `wait`；
-- 其他 function-tool 路由只发送 `run_code`，参数固定为
-  `{ code: string, description: string }`；
+- Function-tool routes expose `run_code` with
+  `{ code: string, description: string, timeout_ms?: number }`;
 - 两种入口都进入同一个 `RunCodeRuntimeManager`，不存在 `transport` 或第二套 runtime；
 - function-tool 路由把当前目录生成成 TypeScript SDK，Codex 路由把嵌套声明写入 `exec`
   描述；
@@ -188,10 +188,17 @@ IPC allowlist、V8 内存限制、heartbeat 和强制回收的组合。
 | READY | 5 秒 |
 | heartbeat 丢失 | 约 3.5 秒后终止 |
 | VM 同步执行 slice | 2 秒 |
-| cell 总执行时间 | 5 分钟 |
+| cell 总执行时间 | Default 5 minutes; optional `timeout_ms` override |
 | yielded / permission cell lease | 60 秒 |
 | RSS hard ceiling | 512 MiB |
 | STOP grace | 500 ms 后 `kill()` |
+
+`timeout_ms` is a positive integer in milliseconds, up to `2147483647`. Function-tool routes pass
+it as an optional `run_code` argument; Codex routes use the first-line `exec` pragma. Models omit
+it for routine work and request a larger value only when the operation is expected to exceed five
+minutes. The deadline covers the cell and its awaited subtools, is not reset by `wait` or permission
+continuation, and does not extend a subtool's own timeout. User cancellation remains independent
+of the selected duration. Invalid timeout values fail before dispatch.
 
 成功、异常、取消、超时、进程退出、会话清理和应用退出都进入同一个幂等 cleanup：取消嵌套
 调用，清理 timer/listener/active map，发送 `STOP`，超时后强制 `kill()`。失败 cell 不自动
--- a/docs/issues/agent-cancellation-code-timeout/spec.md
+++ b/docs/issues/agent-cancellation-code-timeout/spec.md
@@ -1,0 +1,92 @@
+# Agent Cancellation and Code Cell Timeouts
+
+## Context
+
+[Issue #2221](https://github.com/ThinkInAIXYZ/deepchat/issues/2221) reports a Code Mode run that
+stopped progressing and could not be stopped by the user. Its partial trace records a provider
+429 followed by a scheduled retry; it does not identify the complete incident sequence.
+
+Independent reproduction identifies two cancellation boundaries that require protection:
+
+- A provider iterator can remain pending after the Run signal is aborted. Waiting for its next
+  event or asynchronous cleanup must not prevent attempt settlement.
+- An unresolved browser Promise can leave a YoBrowser CDP command and its activity waiting
+  indefinitely unless both observe the tool cancellation signal.
+
+Code cells already have a five-minute execution deadline. That default is useful, but a longer
+operation must be able to request a longer cell lifetime without weakening user cancellation.
+
+## Contract
+
+### User Cancellation
+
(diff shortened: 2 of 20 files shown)


zerob13 (commented):
## Review verdict: APPROVED (no blocking issues)

Reviewed head `70fb36c4aa1599282db8638d5b7496696f7ae3d1` (merge-base diff vs `dev`, 20 files, +874/−124). Full review performed on a local checkout of the PR branch.

### Scope

Two related improvements bundled: (1) agent stop/cancellation signal now propagates into YoBrowser page operations (navigate / CDP send), code-mode host spawn/readiness, and provider stream consumption (#2221); (2) `run_code`/`exec` gain an optional `timeout_ms` (default 5 min, cap 2^31−1) replacing the hard-coded cell deadline.

### Verification

- `vitest run` on the 8 touched suites: **284/284 passed** (BrowserTab, YoBrowserPresenter/ToolHandler, runCodeRuntimeManager, toolModeTools, toolService, contextCoordinator, agentToolManagerYoBrowser).
- `pnpm lint`, `oxfmt --check` clean. `typecheck` failures are all in renderer `markdownWorkerLifecycle.ts` / `markstream-vue` — pre-existing on current `dev`, untouched by this PR.
- Spec updated (`docs/features/experimental-tool-modes/spec.md`) consistently with the implementation.

### Correctness notes (non-blocking)

1. **Aborted navigation leaves the tab in `Loading`** — `BrowserTab.navigateUntilDomReady` throws on abort after `beginMainFrameNavigation()` has already flipped state (`BrowserTab.ts:78-96`); the underlying `loadURL` is not cancelled and `markNavigationError` is not reached. Tab stays `Loading`/`awaitingMainFrameInteractive` until the real navigation event lands. Display-only drift, matches the "no side effects after abort" tests, but worth a follow-up if a cancelled load can stay stale for long.
2. **Non-aborted generator close can still hang** — `contextCoordinator.ts` `observeProviderAttempt` guards the aborted case with fire-and-forget `closing.catch()`, but if the observation body itself throws while `stream.next()` is pending and the signal is *not* aborted, the `finally` awaits `return()` unconditionally; a provider generator that never settles its pending `next()` hangs this path (the code comment acknowledges return() can wait forever, but the guard only covers abort).
3. **RESULT-after-`stopping` is now unconditionally swallowed** — the new `handleCellMessage` early-return (`runCodeRuntimeManager.ts:773`) drops a settled RESULT for a cell that was stopped, even one without `pendingHostSettlement`. Dev already ignored messages for terminal cells and abort rejection handles the waiter, so this is safe today; just noting the settlement-buffer drain order is what makes it so.
4. **Legacy `exec` without pragma is tightened from ∞ to 5 min** — previously pragma-less `exec` cells used the fixed 5-minute deadline too, so behavior is unchanged; the new default/cap only *loosens* it. Spec'd as intentional.

### Style/consistency

- `timeout_ms` validation reuses `readBoundedCodeModeNumber` and the shared protocol constants; prompt/SDK/spec descriptions stay in sync (covered by the new contract test).
- `AgentBrowserToolPort` signature extension is additive; `YoBrowserToolHandler` is its only implementation (wired in `composition.ts`), and the non-agent fallback path intentionally stays signal-less.
- Tests target real cancellation races (late RESULT discard, listener leak, post-abort dispatch), not implementation details. No over-engineering or excessive test surface.
- No breaking changes: all new parameters optional, additive ports, spec-documented.

Nice piece of work on the spawn-abandonment cleanup (`spawnReadyHost` killing the abandoned host on late abort) — that's exactly the fiddly part usually missed.

PR #2227, opened 2026-09-02
fix(orchestration): inherit parent tool mode in delegated sessions

zhangmo8:
## Summary

Delegated subagent sessions did not inherit the parent session's tool mode override. The spawn snapshot carried permission mode, model, and generation settings, but the child's tool mode was resolved from the model catalog default. As a result, a child of an explicitly agent-mode parent could silently run in code mode when the model catalog defaults to `code` (e.g. DeepSeek), since the child uses the same model as the parent and therefore the same catalog default.

## Changes

- Expose `toolModeOverride` on `ConversationSessionInfo` so the parent's resolved override is visible to orchestration
- Carry `toolModeOverride` through `CreateSubagentSessionInput`, the subagent assignment policy, and subagent session creation
- Assignment policy semantics: the parent's override is passed through for DeepChat children and forced to `null` for ACP children (tool mode does not apply to ACP sessions)
- Document in the `deepchat_subagents` tool description that a spawned child inherits the parent's permission mode and tool mode, so its tool calls run under the same trust boundary as approved at spawn time
- Extend the delegation snapshot regression test: the child receives the parent's override as of spawn time, and later parent mutations do not leak into the already-created child
- Update test fixtures for the new required field

## Testing

- `pnpm typecheck` (node + web): pass
- `pnpm format` / `pnpm lint`: pass
- `test/main/session/assignment.test.ts` (21 tests) and `test/main/orchestration/liveDelegationSafety.test.ts` (7 tests): pass
- `test/main/orchestration/liveDelegationService.test.ts` is gated on the native SQLite module in this environment and self-skips; the policy-layer behavior was additionally verified directly (parent override passthrough, missing override normalizes to `null`, ACP children force `null`)

## Notes

- Cross-agent children inherit the parent's tool mode intent as-is. Unlike permission mode, tool mode has no strictness ordering (`agent` / `code` / `minimal` are trade-offs, not a severity scale), so the target agent config does not intersect it.
- Deliberately out of scope: whether cell-internal subagent orchestration should be allowed under the `proactive` policy (the direct-tool carve-out keeps delegation out of the code-mode sandbox). That is a separate product decision.

## Summary by CodeRabbit

* **New Features**
  * Child sessions now inherit the parent session’s tool mode and permission settings, keeping delegated tool use within the approved trust boundary.
  * Tool mode overrides are preserved when creating and managing child sessions.
  * Delegated sessions consistently report their effective tool mode, including when no override is configured.

* **Tests**
  * Added coverage confirming tool mode settings remain consistent for child sessions, even when the parent session changes later.

### DIFF
--- a/src/main/app/composition.ts
+++ b/src/main/app/composition.ts
@@ -1595,6 +1595,7 @@ export async function createMainProcessControl(dependencies: {
           projectDir: session.projectDir ?? null,
           permissionMode,
           orchestrationPolicy: session.orchestrationPolicy,
+          toolModeOverride: session.toolModeOverride ?? null,
           generationSettings,
           disabledAgentTools,
           activeSkills,
--- a/src/main/orchestration/liveDelegationSafety.ts
+++ b/src/main/orchestration/liveDelegationSafety.ts
@@ -1,4 +1,5 @@
 import type { PermissionMode, SessionGenerationSettings } from '@shared/types/agent-interface'
+import type { ToolModeOverride } from '@shared/toolMode'
 import type { ConversationSessionInfo } from '@/tool/runtimePorts'
 import type {
   SessionAgentAssignmentPort,
@@ -10,6 +11,7 @@ export interface LiveDelegationTurnExecutionSnapshot {
   providerId: string
   modelId: string
   generationSettings: SessionGenerationSettings | null
+  toolModeOverride: ToolModeOverride
 }
 
 export interface PrepareLiveDelegationTurnInput {
--- a/src/main/orchestration/liveDelegationService.ts
+++ b/src/main/orchestration/liveDelegationService.ts
@@ -1058,7 +1058,7 @@ export class LiveDelegationService {
           providerId: executionSnapshot.providerId,
           modelId: executionSnapshot.modelId,
           permissionMode: safety.parent.permissionMode,
-          toolModeOverride: safety.parent.toolModeOverride,
+          toolModeOverride: executionSnapshot.toolModeOverride,
           generationSettings: executionSnapshot.generationSettings ?? undefined,
           disabledAgentTools: safety.parent.disabledAgentTools,
           activeSkills: safety.parent.activeSkills,
@@ -2033,7 +2033,8 @@ function createTurnExecutionSnapshot(
     modelId: session.modelId,
     generationSettings: session.generationSettings
       ? structuredClone(session.generationSettings)
-      : null
+      : null,
+    toolModeOverride: session.toolModeOverride
   }
 }
 
--- a/src/main/orchestration/liveDelegationService.ts
+++ b/src/main/orchestration/liveDelegationService.ts
@@ -1058,6 +1058,7 @@ export class LiveDelegationService {
           providerId: executionSnapshot.providerId,
           modelId: executionSnapshot.modelId,
           permissionMode: safety.parent.permissionMode,
+          toolModeOverride: safety.parent.toolModeOverride,
           generationSettings: executionSnapshot.generationSettings ?? undefined,
           disabledAgentTools: safety.parent.disabledAgentTools,
           activeSkills: safety.parent.activeSkills,
--- a/src/main/session/assignmentPolicy.ts
+++ b/src/main/session/assignmentPolicy.ts
@@ -1,4 +1,5 @@
 import { resolveAcpAgentAlias } from '@shared/utils/acpAgentAlias'
+import { normalizeToolModeOverride } from '@shared/toolMode'
 import type {
   DeepChatAgentConfig,
   PermissionMode,
@@ -123,7 +124,8 @@ export class SessionAssignmentPolicy implements SessionAssignmentPolicyPort {
         permissionMode: resolveAssignmentPermissionMode(input.permissionMode),
         generationSettings: { systemPrompt: '' },
         disabledAgentTools: [],
-        activeSkills: []
+        activeSkills: [],
+        toolModeOverride: null
       }
     }
 
@@ -132,6 +134,7 @@ export class SessionAssignmentPolicy implements SessionAssignmentPolicyPort {
     const parentAgentId = input.parentAgentId?.trim() || null
     const isCrossAgent = Boolean(parentAgentId && parentAgentId !== descriptor.id)
     const targetAgentId = input.targetAgentId?.trim() ? descriptor.id : null
+    const toolModeOverride = normalizeToolModeOverride(input.toolModeOverride)
 
     if (!isCrossAgent) {
       return {
@@ -142,7 +145,8 @@ export class SessionAssignmentPolicy implements SessionAssignmentPolicyPort {
         permissionMode: resolveAssignmentPermissionMode(input.permissionMode),
         generationSettings: input.generationSettings,
         disabledAgentTools: normalizeDisabledAgentTools(input.disabledAgentTools),
-        activeSkills: normalizeActiveSkills(input.activeSkills)
+        activeSkills: normalizeActiveSkills(input.activeSkills),
+        toolModeOverride
       }
     }
 
@@ -172,7 +176,8 @@ export class SessionAssignmentPolicy implements SessionAssignmentPolicyPort {
         { disabledAgentTools: input.disabledAgentTools },
         { disabledAgentTools: agentConfig?.disabledAgentTools }
       ).disabledAgentTools,
-      activeSkills: normalizeActiveSkills(input.activeSkills)
+      activeSkills: normalizeActiveSkills(input.activeSkills),
+      toolModeOverride
     }
   }
 
--- a/src/main/session/contracts.ts
+++ b/src/main/session/contracts.ts
@@ -422,6 +422,7 @@ export interface SubagentAssignmentInput {
   generationSettings?: Partial<SessionGenerationSettings>
   disabledAgentTools?: string[]
   activeSkills?: string[]
+  toolModeOverride?: ToolModeOverride
 }
 
 export interface ResolvedSubagentAssignment {
@@ -433,6 +434,7 @@ export interface ResolvedSubagentAssignment {
   generationSettings?: Partial<SessionGenerationSettings>
   disabledAgentTools: string[]
   activeSkills: string[]
+  toolModeOverride: ToolModeOverride
 }
 
 export interface ResolvedTransferTarget {
@@ -584,6 +586,7 @@ export interface SessionLifecycleSubagentInput {
   generationSettings?: Partial<SessionGenerationSettings>
   disabledAgentTools?: string[]
   activeSkills?: string[]
+  toolModeOverride?: ToolModeOverride
   liveDelegationContext?: LiveDelegationSubagentContext
 }
 
--- a/src/main/session/lifecycle.ts
+++ b/src/main/session/lifecycle.ts
@@ -311,7 +311,8 @@ export class SessionLifecycle implements SessionLifecyclePort {
       permissionMode: input.permissionMode,
       generationSettings: input.generationSettings,
       disabledAgentTools: input.disabledAgentTools,
-      activeSkills: input.activeSkills
+      activeSkills: input.activeSkills,
+      toolModeOverride: input.toolModeOverride
     })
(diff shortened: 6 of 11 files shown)


zerob13 (changes requested):
## Code Review — Request Changes

The overall shape is right: minimal field plumbing, correct ACP `null`-out, well-justified cross-agent passthrough. One blocking issue, verified by actually running the suite that self-skipped in your environment.

### 🔴 Blocking: spawn-time freeze is not implemented — the modified regression test fails when run

The PR states the child receives the parent's override **as of spawn time**. The implementation does the opposite:

- `src/main/orchestration/liveDelegationService.ts:1061` reads `toolModeOverride: safety.parent.toolModeOverride`
- `safety` is resolved at **turn execution time** (`executeTurn` → `resolveCurrentSafety` → `requireCapableParent` → live `resolveConversationSessionInfo`), so the child receives the parent's **current** override at execution time, not the spawn-time value.

Running the suite this PR modifies:

```
pnpm vitest run test/main/orchestration/liveDelegationService.test.ts

Test Files  1 failed (1)
      Tests  1 failed | 57 passed (58)
```

```
- toolModeOverride: "code"   // expected: spawn-time value
+ toolModeOverride: "agent"  // received: post-scheduling live value
```

`test/main/orchestration/liveDelegationService.test.ts:1073` documents the intended semantics; the implementation doesn't deliver it. The PR body notes this suite self-skips in your environment (native SQLite gate) — but this is exactly the file whose expectations this PR changed, so the modified assertion has never been executed anywhere.

**Suggested fix** (matches the frozen `providerId`/`modelId`/`generationSettings` pattern):

1. Add `toolModeOverride` to `LiveDelegationTurnExecutionSnapshot` (`src/main/orchestration/liveDelegationSafety.ts`)
2. Populate it in `createTurnExecutionSnapshot` (`liveDelegationService.ts:2028`)
3. Use `executionSnapshot.toolModeOverride` in `acquireAndBindChild` instead of `safety.parent.toolModeOverride`

The `follow_up` path already threads `executionSnapshot` through `synchronizeChildSafety` → `prepareTurn`, so this stays coherent. (Same as the CodeRabbit comment on line 1061 — agreed.)

Alternatively, if live-read semantics were intended, both the test expectation and the PR description need to change — but that contradicts the documented intent, so the snapshot approach is preferred.

### ✅ What looks good

- Minimal, focused diff (+28/−5); follows the existing `permissionMode` plumbing pattern.
- ACP children forced to `null`; cross-agent passthrough without intersection is correct and well-justified (tool modes are trade-offs, not a severity scale).
- Data chain verified: `lifecycle.ts` → `appSessionService` persists normalized `tool_mode_override` → read back for the parent snapshot in `composition.ts`.
- No breaking changes: `typecheck:node` passes with the new required `ConversationSessionInfo.toolModeOverride` field; test additions are proportionate, not excessive.

### Verification (this review)

- `pnpm vitest run test/main/orchestration/liveDelegationService.test.ts` → **1 failed** / 57 passed
- `pnpm vitest run test/main/session/assignment.test.ts test/main/orchestration/liveDelegationSafety.test.ts` → 28 passed
- `pnpm typecheck:node` → pass (renderer `typecheck:web` failures are pre-existing on dev, untouched by this PR)

zhangmo8:
The first CI run failed in `test-main`: `assignmentPolicy.test.ts` asserts the full resolved assignment with `toEqual`, so the new required `toolModeOverride` field on `ResolvedSubagentAssignment` broke three assertions (8 keys vs 7). Updated those expectations (`null` for ACP and cross-agent children without an explicit override) and extended the self-target test to assert parent-override passthrough (`'code'` in → `'code'` out). All 53 tests in the three session suites now pass locally.

zerob13 (changes requested):
## Code Review — Request Changes (re-review)

**Delta since last review** (`681a4410` → `6b3cf060`): only `test/main/session/assignmentPolicy.test.ts` expectations updated for the new required `ResolvedSubagentAssignment.toolModeOverride` field. Production code is unchanged. That fixes the CI failure, and `assignmentPolicy.test.ts` + `liveDelegationSafety.test.ts` now pass — but the blocking P1 from the previous round is unaddressed.

### P1 — spawn-time freeze of `toolModeOverride` still not implemented

`acquireAndBindChild` reads `safety.parent.toolModeOverride` ([liveDelegationService.ts:1061](https://github.com/ThinkInAIXYZ/deepchat/blob/6b3cf0604a1cbc110e3aff71fee69b50b1c2318a/src/main/orchestration/liveDelegationService.ts#L1061)), but `safety` is resolved at turn **execution** time (`executeTurn` → `resolveCurrentSafety` at :1006 → live `resolveConversationSessionInfo`). The child therefore inherits whatever the parent's tool mode is *when the turn executes*, not at spawn/scheduling time — contradicting the PR's stated behavior and its own regression test.

Verified by running the PR's own suites on this head:

```
pnpm vitest run test/main/orchestration/liveDelegationService.test.ts \
  test/main/orchestration/liveDelegationSafety.test.ts \
  test/main/session/assignmentPolicy.test.ts
# → 1 failed / 75 passed
```

The failure is `freezes generation settings while applying live permission and workspace safety` (liveDelegationService.test.ts:1068): expected `toolModeOverride: "code"` (value at scheduling time), received `"agent"` (post-scheduling live value). Note the suite is SQLite-gated (`describeIfSqlite`), so it self-skips in environments without native SQLite — please run it locally with the gate enabled before pushing.

**Fix direction** (unchanged, agrees with CodeRabbit's inline comment):

- Add `toolModeOverride` to `LiveDelegationTurnExecutionSnapshot` ([liveDelegationSafety.ts:9](https://github.com/ThinkInAIXYZ/deepchat/blob/6b3cf0604a1cbc110e3aff71fee69b50b1c2318a/src/main/orchestration/liveDelegationSafety.ts#L9))
- Populate it in `createTurnExecutionSnapshot` ([liveDelegationService.ts:2028](https://github.com/ThinkInAIXYZ/deepchat/blob/6b3cf0604a1cbc110e3aff71fee69b50b1c2318a/src/main/orchestration/liveDelegationService.ts#L2028))
- Use `executionSnapshot.toolModeOverride` at :1061 instead of `safety.parent.toolModeOverride`
- The `follow_up` path already threads `executionSnapshot` through `synchronizeChildSafety` → `prepareTurn`, so this lands cleanly.

Everything else from the previous round still holds: minimal plumbing, ACP children forced to `null`, cross-agent passthrough without intersection is justified, and the data chain (`lifecycle` → `composition` parent snapshot) is verified. Once the snapshot freeze is implemented and the regression test passes, happy to approve.

zhangmo8:
Both reviews pointed at the same root cause — agreed, and fixed in f164f3aa5 exactly as suggested:

- `toolModeOverride` added to `LiveDelegationTurnExecutionSnapshot`
- Populated in `createTurnExecutionSnapshot` alongside `providerId` / `modelId` / `generationSettings`
- `acquireAndBindChild` now reads `executionSnapshot.toolModeOverride`, so the child is created from the spawn-time frozen value; later parent mutations cannot leak in

Verification this time went further than the first round: I made the native SQLite gate runnable locally (rebuilt the `better-sqlite3-multiple-ciphers` prebuild for Node 24, restored the original Node 26 binary afterwards). The previously self-skipped suite now executes: **58/58 passed**, including the spawn-freeze assertion that failed before the fix (1 failed / 57 passed in @zerob13's run).

One deliberate non-change: the snapshot already threads through `synchronizeChildSafety` → `prepareTurn` on `follow_up`, but I did not extend `applyTurnExecutionSnapshot` to re-apply the tool mode to an existing child. Re-application would reset a drifted child mode and felt like a behavior/product call rather than part of this inheritance fix; happy to add it if maintainers want drift-reset semantics for tool mode too.

zhangmo8:
@zerob13 your re-review (11:17Z) landed 13 minutes before the fix commit and targets `6b3cf060`; the P1 is addressed by f164f3aa5, pushed at 11:30Z:

- `toolModeOverride` added to `LiveDelegationTurnExecutionSnapshot`, populated in `createTurnExecutionSnapshot`, and `acquireAndBindChild` now reads the frozen `executionSnapshot.toolModeOverride` — exactly the three steps from your review and the CodeRabbit comment.

Verification (this time with the suite actually executing): I made the native SQLite gate runnable locally (Node 24 prebuild for `better-sqlite3-multiple-ciphers`, original Node 26 binary restored afterwards). `liveDelegationService.test.ts` — the file containing the failing assertion from your run — now passes **58/58**, including `freezes generation settings while applying live permission and workspace safety`. CI on f164f3aa5 is green (test-main, build, static, test-renderer, test-native-memory).

The one deliberate non-change from my earlier comment stands: `applyTurnExecutionSnapshot` does not re-apply tool mode to existing children on `follow_up` — happy to add drift-reset if you consider it in scope.

zerob13 (approved):
## Code Review — Approve (re-review #2)

**Delta since last review** (`6b3cf060` → `f164f3aa`): `fix(orchestration): freeze tool mode in spawn snapshot` — exactly the fix direction from the previous round, +5/−2 across `liveDelegationSafety.ts` / `liveDelegationService.ts`. The blocking P1 is resolved.

### ✅ P1 fixed — spawn-time freeze now implemented

- `LiveDelegationTurnExecutionSnapshot` now carries `toolModeOverride` ([liveDelegationSafety.ts:14](https://github.com/ThinkInAIXYZ/deepchat/blob/f164f3aa540dd274253655623be7d67a8b8720b0/src/main/orchestration/liveDelegationSafety.ts#L14)), populated in `createTurnExecutionSnapshot` from the session resolved at scheduling time ([liveDelegationService.ts:2037](https://github.com/ThinkInAIXYZ/deepchat/blob/f164f3aa540dd274253655623be7d67a8b8720b0/src/main/orchestration/liveDelegationService.ts#L2037)).
- The snapshot is created **before** `scheduleTurn` on every path — spawn (`:403` → `:422`), follow-up (`:518`), retry (`:1644`) — so the value is captured at turn scheduling, not execution.
- `acquireAndBindChild` now reads `executionSnapshot.toolModeOverride` ([liveDelegationService.ts:1061](https://github.com/ThinkInAIXYZ/deepchat/blob/f164f3aa540dd274253655623be7d67a8b8720b0/src/main/orchestration/liveDelegationService.ts#L1061)) instead of the live `safety.parent.toolModeOverride`, matching the frozen `providerId`/`modelId`/`generationSettings` pattern. `permissionMode` / `disabledAgentTools` / `activeSkills` intentionally stay live — consistent with the regression test's stated contract ("freezes generation settings while applying live permission and workspace safety").

The previously failing regression test (mutates the parent's tool mode between scheduling and execution and expects the frozen `'code'`) now passes.

### Verification (this review)

- `pnpm vitest run test/main/orchestration/liveDelegationService.test.ts test/main/orchestration/liveDelegationSafety.test.ts test/main/session/assignmentPolicy.test.ts` → **76 passed (3 files)**, including the SQLite-gated freeze test that self-skips in your environment
- CI on `f164f3aa`: build / static / test-main / test-native-memory / test-renderer / package-impact all pass

### 🟡 Non-blocking observations

1. **Follow-up restore asymmetry**: `restoreExecutionSnapshot` ([liveDelegationSafety.ts:126](https://github.com/ThinkInAIXYZ/deepchat/blob/f164f3aa540dd274253655623be7d67a8b8720b0/src/main/orchestration/liveDelegationSafety.ts#L126)) restores `providerId`/`modelId`/`generationSettings` for `follow_up` turns but not `toolModeOverride`. Today this is self-consistent because the follow-up snapshot is taken **from the child** (`:518`), so the field equals the child's scheduling-time value — but if the child's tool mode is mutated between scheduling and execution, provider/model/generation are protected while tool mode is not. If you want the freeze to be complete, `applyTurnExecutionSnapshot` would need to accept `toolModeOverride`; fine to leave for a follow-up given `permissionMode` is intentionally live in the same window.
2. The tool description now says the child inherits "permission mode and tool mode" — permission mode is live-at-execution while tool mode is frozen-at-spawn. Minor nuance, no change needed.

No over-engineering, no breaking changes (additive optional inputs; the `?? null` fallback in `composition.ts` covers legacy persisted sessions), test changes are proportionate. LGTM.

PR #2093, opened 2026-08-06
feat(provider): add DeepSeek native web search

yyhhyyyyyy:
## Summary

- Add provider-native Web Search for DeepSeek V4 Flash through the official Responses API.
- Route only search and compatible replay requests through Responses; ordinary requests retain the configured Chat Completions transport and endpoint.
- Add a per-turn search toggle with visible search activity and clickable source links.
- Support stateless multi-turn replay of DeepSeek search results.
- Upgrade all existing AI SDK packages to their latest stable versions.

## Implementation

- Reuse the existing OpenAI Responses transport without adding a direct `@ai-sdk/deepseek` dependency.
- Isolate DeepSeek protocol handling in a request-scoped replay adapter.
- Preserve raw `web_search_call` payloads for replay and normalized search blocks for UI and export.
- Force `store: false` and reject unresolved replay markers before sending requests.
- Keep malformed persisted replay out of both history and in-flight tool rounds while retaining fail-closed wire validation for accepted markers.
- Store versioned replay envelopes in existing assistant block metadata without database migrations.
- Strip opaque replay payloads from streaming snapshots and paginated client reads while keeping durable storage intact.
- Recover resume search intent from the closest persisted user record, including turns containing multiple assistant records.
- Include replay payloads in context budgeting and trim replay-bearing turns atomically when required.
- Send `search: true` only when explicitly enabled, preserving the existing payload shape otherwise.
- Keep legacy MCP search blocks separate from provider-native search activity.
- Open result links through the existing safe external-link path.

## Validation

- [x] Format, i18n, lint, node typecheck, and web typecheck.
- [x] Full main suite: 6301 passed, 349 skipped.
- [x] Focused routing, replay, session projection, dispatch, and resume suite: 446 passed.
- [x] Local two-round conformance verifies raw `web_search_call` replay, `store: false`, no `item_reference`, and no continuation fields.
- [x] Real-key native-search canary confirms completed `search` and `open_page` items.
- [ ] **Blocking merge gate:** run a real-key second independent user turn and confirm DeepSeek accepts the replayed `web_search_call` while retaining the first turn's searched context.

## UI

Before:

```text
Composer:  [+] [Mic] [Send]
Response:  Assistant text only
```

After:

```text
Composer:  [+] [Web Search] [Mic] [Send]
Response:  Search activity
           Sources
           - Clickable source title
           - Clickable source title
```

## Summary by CodeRabbit

* **New Features**
  * Added native web search for supported DeepSeek V4 Flash models.
[...]

PR #2228, opened 2026-09-02
fix(provider): restore DeepSeek Responses replay

yyhhyyyyyy:
## Summary

- migrate the official DeepSeek V4 Flash search/replay route from `@ai-sdk/openai` to the stateless `@ai-sdk/open-responses` transport
- preserve plaintext reasoning replay, native `web_search_call` history, reasoning effort, and incremental MCP tool input
- keep ordinary DeepSeek requests on Chat Completions and keep DeepSeek-specific wire rules inside the existing fail-closed adapter

## Root cause

[PR #2093](https://github.com/ThinkInAIXYZ/deepchat/pull/2093) routed DeepSeek native search and compatible replay through `@ai-sdk/openai`. That provider models OpenAI's server-stored or encrypted continuation semantics: it drops id-less plaintext reasoning, does not consume DeepSeek's `reasoning_text` stream, and gates reasoning effort using recognized OpenAI model names.

As a result, a tool-bearing follow-up could omit the reasoning text that DeepSeek requires clients to replay and fail with a `reasoning_text` 400. Configured `low`/`max` effort could also be silently absent from the request.

## Implementation

- add a dedicated `deepseek-open-responses` provider kind backed by `createOpenResponses({ name: 'deepseek' })`
- activate it only for the exact official DeepSeek V4 Flash search/replay boundary
- inject the bare `{ type: 'web_search' }` tool only when search is enabled
- persist complete provider search items in the existing opaque replay envelope
- serialize replay markers as private `function_call` items, replace them with the original `web_search_call` exactly once, and reject malformed or residual replay state before network I/O
- delegate id-less plaintext reasoning and `low`/`high`/`max` effort serialization to Open Responses without synthesizing empty reasoning items
- project raw function-call start/delta events so MCP argument cards remain incremental, while the provider's canonical event closes the call without duplication
- update the maintained feature spec and protocol regression coverage
- include the provider/ACP registry snapshots refreshed by the required normal build

## Compatibility

- ordinary V4 Flash turns without search or compatible replay retain the configured Chat Completions transport
- custom DeepSeek endpoints, relays, other providers, and other model IDs are unchanged
- existing opaque search envelopes remain replayable; incompatible provider/model histories keep their visible text and omit provider protocol state
- historical turns that never persisted reasoning text cannot be reconstructed locally
- no database schema or persisted provider configuration migration is required

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run architecture:renderer-baseline:check`
- `pnpm run icons:check`
- `pnpm run typecheck`
- `pnpm run test:main` — 633 files, 8,379 passed / 461 skipped
- `NODE_OPTIONS=--max-old-space-size=6144 pnpm run test:renderer` — 267 files, 2,361 passed
- `pnpm run build`
- Linux DuckDB VSS install/smoke and native SQLite smoke
- portable memory — 54 files, 888 passed
- native Tape — 9 files, 230 passed
- encrypted OCR storage — 2 files, 12 passed
- native memory — 349 passed / 2 skipped
- memory retrieval eval — 7 passed
- memory performance — 10 passed
- real DeepSeek key: native search/open-page, `summary: []`, id-less reasoning replay, multiple interleaved reasoning/search items, same-turn MCP loop, provider function item-ID replay, `low`/`max` effort, and search-disabled replay all completed without the original 400

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek Open Responses support for native web search.
  * Added plaintext reasoning streaming and configurable reasoning effort levels.
  * Improved replay of web-search and tool interactions across conversation turns.
  * Added request tracing and normalized search-source events.

* **Bug Fixes**
  * Prevented duplicate streamed tool-call events and arguments.
  * Rejected unsupported replay markers and reserved-tool name conflicts.
  * Improved handling of empty reasoning content and stateless requests.

* **Documentation**
  * Updated DeepSeek web-search implementation plans, specifications, and validation checklists.


zerob13 (approved):
## Code Review — Approve

**Reviewed head**: `1da1fc17e` (8 commits, base `dev` @ `d42106e54`)

**Scope**: DeepSeek Responses migration from `@ai-sdk/openai` to `@ai-sdk/open-responses` (official Vercel package), plus generated catalog refresh.

**Verified locally** (on the PR branch):
- `typecheck:node` ✅ · `oxlint` ✅ · `oxfmt --check` ✅
- 145/145 tests across the 5 changed test files ✅
- Read `@ai-sdk/open-responses@2.0.38` dist source to confirm the design assumptions below against actual package behavior.

**Findings** — the design holds up:

1. **Migration rationale confirmed.** The package genuinely parses plaintext `reasoning_text` and resolves `reasoningEffort` from `providerOptions['deepseek']` (follows `createOpenResponses({ name })`), matching `providerOptionsMapper`'s `open_responses` case. The `@ai-sdk/openai` limitations documented in the spec are real, not post-hoc justification.

2. **Tool-call dedupe is sound.** The provider buffers `function_call_arguments.delta` internally and emits only an atomic final `tool-call` (no `tool-input-delta` parts), so the adapter's raw projection (`output_item.added` → `tool_call_start`, deltas → `tool_call_chunk`, `item_id`↔`call_id` correlation) is the only incremental-argument source, and the `rawProjectedToolCalls` guard stops the canonical event from double-emitting start/chunk. Missing/malformed raw events fall back to the atomic path. Verified against the package's stream parser.

3. **Stateless-contract simplification is safe.** `@ai-sdk/open-responses` never sends `store` / `previous_response_id` / `conversation` / `truncation`, so dropping the old continuation-field surgery is correct; and the request-local `providerPatch` (`baseUrl: https://api.deepseek.com`) guarantees the strict `/responses` pathname assertion holds even when the persisted baseUrl carries `/v1`.

4. **No breaking changes.** The `providerKind` rename and the replay-marker tool-name rename are both request-local / regenerated at prompt-build time; persisted history only stores the versioned `provider_replay` envelope, so old conversations stay replayable (`omitEmptyReasoning` additionally strips empty `reasoning_content` from old records). Removed type exports had no external consumers.

5. **Trace move is an improvement.** With the adapter active, request traces now capture the final wire body (after replay substitution + `web_search` injection) from `wrapFetch`, and the `if (!runtime.providerAdapter)` guard prevents double-tracing.

6. **Tests: appropriate, not excessive.** 41 adapter tests + 19 stream tests map to distinct spec behaviors (wire validation, replay persistence, dedupe, real-serializer integration). Regression protection for persistence + protocol-boundary code; nothing to trim.

Minor notes (non-blocking):
- `applyNativeWebSearchTool`'s search-disabled refusal is effectively unreachable (nothing else can inject `web_search`), but it is consistent with the adapter's fail-closed style — fine to keep.
- The lockfile now carries two `@ai-sdk/provider` (4.0.5/4.0.10) and `provider-utils` (5.0.22/5.0.36) instances; plan.md already records this as the compatibility boundary until the next coordinated SDK update. 👍

CI green on all 15 checks. Nice work.

PR #2229, opened 2026-09-03
feat(workspace): add file open app functionality and context menu options

Soya-xy:
CleanShot 2026-09-03 at 16 44 01@2x

## Summary by CodeRabbit

- **New Features**
  - Open workspace files with compatible installed editors, IDEs, or terminal applications.
  - Choose an application from a context menu or use the system default.
  - Available applications and icons are detected for macOS, Windows, and Linux.
  - Preferred applications are remembered for future file-opening actions.

- **Bug Fixes**
  - Added clearer error notifications when opening files or revealing locations fails.
  - Automatically fall back to the system default when a preferred application is unavailable.

- **Localization**
  - Added translated labels and error messages across supported languages.


zerob13 (changes requested):
## Review Summary

Solid feature structure overall: the route/contract/service/client layers follow the existing workspace patterns exactly (`defineRouteContract` + zod schemas, route catalog registration, `isPathAllowed` guards mirroring `openFile`/`revealFileInFolder`), `appId` is validated against the detected installed list so the renderer can't turn this into an arbitrary command launcher, and i18n keys are complete across all 20 locales. The curated-registry approach with a single JXA batch probe on macOS is a good design — not over-engineered.

However, the terminal half of this feature is functionally broken on Windows and Linux, and Windows app detection misses the most common per-user installs. Requesting changes for the items below.

### 🔴 P1 — Terminal launching is broken on Windows and Linux

`openFileWithApp` assumes every app accepts the target path as a positional argument ([fileOpenApps.ts:356-372](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/main/workspace/fileOpenApps.ts#L356-L372)):

- **Windows**: `spawn(wt.exe, [dir])` — `wt.exe` treats positional arguments as the command line to run, so this opens a tab that tries to *execute* the directory path. The correct invocation is `wt.exe -d <dir>`.
- **Linux**: `gio launch <desktop-entry> <dir>` passes the directory as a positional file argument. The `Exec=` lines of `org.gnome.Terminal.desktop` and `org.kde.konsole.desktop` have no `%f`/`%U` field codes, so the argument is silently ignored and the terminal opens in the home directory. Correct flags are `gnome-terminal --working-directory=<dir>` and `konsole --workdir <dir>`; only kitty and a few others actually cd into a positional directory.

The registry (`WorkspaceFileOpenAppDefinition`) stores detection identifiers (bundleIds/executables/desktopIds) but has no per-app launch-args concept, which terminals require on all three platforms. Suggest extending the definition with per-platform launch args (e.g. an `openArgs(path)` builder per app), or at minimum special-casing terminals per platform.

### 🟠 P2 — Windows detection misses per-user installs (the default VS Code/Cursor install mode)

`resolveWindowsExecutable` only queries `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths` ([fileOpenApps.ts:144](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/main/workspace/fileOpenApps.ts#L144)). Per-user installs register their `App Paths` under `HKCU` (per-user apps can't write HKLM), and the `where <name>.exe` fallback misses them because the CLI shims on PATH are `.cmd` files (`code.cmd`), not `.exe`. On a stock Windows machine with the default User Installer of VS Code, the most common editor won't appear in the picker at all. Please query `HKCU` as well (or first), and consider probing well-known per-user install locations directly.

### 🟠 P2 — Silent fallback to system default masks launch failures

`WorkspaceService.openFileWithApp` catches every launch error, logs it, and silently falls back to `openFile`, after which the route returns `{ opened: true }` ([index.ts:821-834](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/main/workspace/index.ts#L821-L834)). The user selects "Open in iTerm2", the launch fails, and the system-default app opens instead with zero feedback — combined with P1, every Windows terminal click would "succeed" while opening the wrong thing. Please propagate the failure through the route (reject, or return a status like `{ opened, fallbackUsed }`) so the renderer can surface an error toast.

### 🟡 P3 — Primary button semantics changed (needs explicit callout)

Previously the header button always opened the file with the system default. Now the primary button opens the remembered app, or auto-picks the first editor the OS registers for that file type ([WorkspaceViewer.vue:376-385](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/renderer/src/components/sidepanel/WorkspaceViewer.vue#L376-L385)). Two behaviors worth confirming as intentional: (a) files better handled outside the registry (images, PDFs) now get an *editor* auto-picked on the primary click; (b) "System default" is only reachable via the dropdown and is never persisted as the preferred choice, so users preferring the old behavior must open the dropdown every time. If intentional, please document it in the PR description.

### 🟡 P3 — Minor

- `detectWindowsInstalledApps` probes ~25 executables serially, each with up to 2 subprocess spawns and 8s timeouts ([fileOpenApps.ts:167-189](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/main/workspace/fileOpenApps.ts#L167-L189)); the first `listFileOpenApps` call can take seconds on a cold start. `Promise.all` over definitions would parallelize this trivially.
- `installedAppsPromise` is cached for the process lifetime, so apps installed while DeepChat is running never appear until restart. Consider a short TTL or re-probe on menu open.
- JSDoc nit: `@param appId Platform launch identifier of the chosen application` in `WorkspaceServicePort` ([workspace.ts:218](https://github.com/ThinkInAIXYZ/deepchat/blob/615f3200c571f5758d87c4955c397cb6934c8ffd/src/shared/types/workspace.ts#L218)) — it's a registry id, not a platform identifier.

### ✅ What looks good

- Layered architecture matches existing workspace routes end-to-end (contract → catalog → route → service → client); schema and route definitions follow the house style.
- Security posture is right: `isPathAllowed` guards both new methods exactly like `openFile`/`revealFileInFolder`, and `openFileWithApp` only launches ids from the locally detected installed list — no arbitrary command surface exposed to the renderer.
- macOS implementation is clean: one JXA subprocess probes all bundle ids and renders real app icons; Launch Services handler ordering cached per extension with proper failure cleanup.
- i18n complete: 4 new keys across all 20 locales with `{app}` params, `i18n.d.ts` regenerated; `DcButton` tooltip/`size="icon"` usage is valid and accessible.
- Watch race-guard (`openFilePath.value === filePath`) prevents stale async results when files are switched quickly.
- No new tests — reasonable for platform-heavy code; nothing over-tested.

zhangmo8 (changes requested):
## Review Summary

The feature is well-structured overall: contract/route/service/client layers follow the existing workspace patterns end-to-end, `appId` is validated against the locally detected installed list (so the renderer can't turn this into an arbitrary command launcher), the macOS implementation works correctly, and i18n keys are complete across all 20 locales. However, the terminal half of the feature is functionally broken on Windows and Linux, and Windows detection misses the most common per-user installs. Requesting changes.

### 🔴 P1 — Terminal launching is broken on Windows and Linux

`openFileWithApp` assumes every app accepts the target path as a positional argument ([fileOpenApps.ts:356-372](src/main/workspace/fileOpenApps.ts#L356-L372)):

- **Windows**: `spawn(wt.exe, [dir])` — `wt.exe` treats positional arguments as the command line to run in the new tab, so this tries to *execute* the directory path. The correct invocation is `wt.exe -d <dir>`.
- **Linux**: `gio launch <desktop-entry> <dir>` passes the directory as a positional file argument. Neither `org.gnome.Terminal.desktop` nor `org.kde.konsole.desktop` has `%f`/`%U` field codes in `Exec=`, so the argument is silently ignored and the terminal opens in the home directory. Correct flags are `gnome-terminal --working-directory=<dir>` and `konsole --workdir <dir>`; only a few terminals (e.g. kitty) actually cd into a positional directory.

The registry (`WorkspaceFileOpenAppDefinition`) stores detection identifiers but has no per-app launch-args concept, which terminals need on all three platforms. Suggest extending the definition with per-platform launch args (e.g. an `openArgs(path)` builder), or at minimum special-casing terminals per platform.

### 🟠 P2 — Windows detection misses per-user installs (the default VS Code/Cursor install mode)

`resolveWindowsExecutable` only queries `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths` ([fileOpenApps.ts:144](src/main/workspace/fileOpenApps.ts#L144)). Per-user installs register under `HKCU` (per-user apps can't write HKLM), and the `where <name>.exe` fallback misses them because the CLI shims on PATH are `.cmd` files (`code.cmd`), not `.exe`. On a stock Windows machine with the default User Installer of VS Code, the most common editor won't appear in the picker at all. Please query `HKCU` as well (or first), and consider probing well-known per-user install locations directly.

### 🟠 P2 — Silent fallback to system default masks launch failures

`WorkspaceService.openFileWithApp` catches every launch error, logs it, and silently falls back to `openFile`, after which the route returns `{ opened: true }` ([index.ts:821-834](src/main/workspace/index.ts#L821-L834)). The user selects "Open in iTerm2", the launch fails, and the system-default app opens instead with zero feedback — combined with P1, every broken Windows terminal click would "succeed" while opening the wrong thing. Please propagate the failure through the route (reject, or return a status like `{ opened, fallbackUsed }`) so the renderer can surface an error.

### 🟡 P3 — Primary button semantics changed (needs explicit callout)

Previously the header button always opened the file with the system default. Now the primary button opens the remembered app, or auto-picks the first editor the OS registers for that file type ([WorkspaceViewer.vue:376-385](src/renderer/src/components/sidepanel/WorkspaceViewer.vue#L376-L385)). Two behaviors worth confirming as intentional: (a) files better handled outside the registry (images, PDFs) now get an *editor* auto-picked on the primary click; (b) "System default" is only reachable via the dropdown and is never persisted as the preferred choice, so users preferring the old behavior must open the dropdown every time. If intentional, please document it in the PR description.

### 🟡 P3 — Minor

- `detectWindowsInstalledApps` probes ~25 executables serially, each with up to 2 subprocess spawns and 8s timeouts ([fileOpenApps.ts:167-189](src/main/workspace/fileOpenApps.ts#L167-L189)); the first `listFileOpenApps` call can take seconds on a cold start. `Promise.all` over definitions would parallelize this trivially.
- `installedAppsPromise` is cached for the process lifetime, so apps installed while DeepChat is running never appear until restart. Consider a short TTL or a re-probe when the menu opens.
- JSDoc nit: `@param appId Platform launch identifier of the chosen application` ([workspace.ts:218](src/shared/types/workspace.ts#L218)) — it's a registry id, not a platform identifier.

### ✅ What looks good

- Layered architecture matches existing workspace routes end-to-end (contract → catalog → route → service → client); schemas and route definitions follow the house style.
- Security posture is right: `isPathAllowed` guards both new methods exactly like `openFile`/`revealFileInFolder`, and `openFileWithApp` only launches ids from the locally detected installed list — no arbitrary command surface exposed to the renderer. Terminals receive the containing directory rather than the file, which avoids terminals trying to execute the file.
- macOS implementation is clean and verified working: one JXA subprocess probes all bundle ids and renders real app icons; Launch Services handler ordering is cached per extension with proper failure cleanup.
- i18n complete: 4 new keys across all 20 locales with `{app}` params; `i18n.d.ts` regenerated; `DcButton` tooltip/`size="icon"` usage is valid and accessible.
- The watch race-guard (`openFilePath.value === filePath`) prevents stale async results when files are switched quickly.
- No new tests — reasonable for platform-heavy code; nothing over-tested.

zhangmo8 (commented):
**Follow-up review after the fix commits (716e41c, 0336a157).** The previously raised blockers are addressed: HKCU is now queried before HKLM, terminals use per-platform launch arguments (`wt -d`, `--working-directory`, `--workdir`), launch failures now reject instead of silently falling back, detection is parallelized, and the installed-apps cache has a 60s TTL. Remaining new findings:

### 🟠 P2 — `REG_EXPAND_SZ` values from `App Paths` are never expanded

[`readRegistryDefault`](src/main/workspace/fileOpenApps.ts#L129-L137) deliberately matches both `REG_SZ` and `REG_EXPAND_SZ` (`REG_(?:EXPAND_)?SZ`), but returns the raw value. A `REG_EXPAND_SZ` default such as `%ProgramFiles%\Microsoft VS Code\Code.exe` is passed to `fs.existsSync` ([fileOpenApps.ts:147](src/main/workspace/fileOpenApps.ts#L147)) unexpanded, which always fails. The `where` fallback can't save it either: per-user installs put a `code.cmd` shim on PATH (filtered out by the `.exe` check) while the real `code.exe` lives only under App Paths. Net effect: on a machine where the App Paths default is stored as `REG_EXPAND_SZ`, the HKCU fix from 716e41c has no effect and the editor is never detected. Please expand environment variables (e.g. `%VAR%` → value) before `fs.existsSync`.

### 🟡 P3 — `handlerCache` never invalidates

The installed-apps cache got a TTL, but the macOS handler ordering cache ([fileOpenApps.ts:30](src/main/workspace/fileOpenApps.ts#L30), [fileOpenApps.ts:273-300](src/main/workspace/fileOpenApps.ts#L273-L300)) lives for the process lifetime. If the user changes the default app for an extension (or a new handler registers) while DeepChat is running, the "Open with" ordering stays stale until restart. A short TTL or cache-busting on app activation would keep it consistent with `installedAppsCache`.

### 🟡 P3 — Linux editors still rely on `gio launch` passing the file positionally

The terminal half was fixed with per-app overrides, but editors without a `launch.linux` override still go through `gio launch <desktop-entry> <file>` ([fileOpenApps.ts:361-362](src/main/workspace/fileOpenApps.ts#L361-L362)). `gio launch` only forwards the positional argument to apps whose `Exec=` line contains `%f`/`%F`/`%U`. VS Code's `code.desktop` has `%F`, but JetBrains Toolbox-generated desktop entries typically don't include field codes, so the file is silently ignored and the IDE opens empty. Consider the same per-app launch-args treatment for editors, or at least verify `Exec=` contains a field code before relying on the positional arg.

### 🟡 P3 — Windows console apps get a flash console window

`nvim.exe` (and `emacs.exe`, the console entry point of the official Windows build) are console-subsystem executables. `spawn(command, args, { detached: true, stdio: 'ignore' })` ([fileOpenApps.ts:367](src/main/workspace/fileOpenApps.ts#L367)) gives them a brand-new console window that flashes on screen, and with `stdio: 'ignore'` no interactive terminal is attached, so the editor can fail or render uselessly. Consider `windowsHide: true` plus a real terminal host (e.g. `wt`/`conhost`), or excluding console-only editors on win32.

### ✅ Confirmed good

- Layered contract → route → service → client wiring follows the existing workspace patterns exactly; zod schemas are tight.
- `appId` is validated against the locally detected list — no arbitrary command surface.
- i18n keys complete across all 20 locales with `{app}` interpolation; `i18n.d.ts` regenerated.
- macOS probe is efficient (single JXA subprocess, cached per-extension handler ordering) and the race guard in `WorkspaceViewer.vue` (`openFilePath.value === filePath`) is correct.

zerob13 (changes requested):
## Follow-up review after `716e41c` / `0336a157`

Re-reviewed the delta against my earlier review at `615f320c`. All previously raised findings are resolved, verified at HEAD `0336a157`:

- ✅ **Terminal launch args** — per-platform `launch` overrides with `{path}` templates now cover `wt -d`, `--working-directory`, `--workdir`, wezterm `start --cwd`, kitty `--directory`, alacritty/ghostty; arguments go through `buildLaunchArgs` and spawn as an argv array, so there is no injection surface.
- ✅ **HKCU before HKLM**, and the `where` fallback now only accepts a real `.exe` that exists on disk, ignoring `.cmd` shims.
- ✅ **Launch failures reject** instead of silently falling back to the system default; the renderer surfaces an `openFailed` error toast and the key is present in all 20 locales with `i18n.d.ts` regenerated.
- ✅ **Windows probing is parallelized**, the installed-apps cache has a 60s TTL with re-probe on failure, the primary button no longer auto-picks an editor, and the system default is now rememberable via the `#system-default` sentinel.
- ✅ **`isRegisteredHandler` removal is the right API narrowing** — no stale references remain in `src/` or `test/`.

Remaining items — one should-fix:

### 🟠 P2 — `REG_EXPAND_SZ` values from `App Paths` are never expanded

`readRegistryDefault` deliberately matches `REG_(?:EXPAND_)?SZ` but returns the raw value ([fileOpenApps.ts:134](src/main/workspace/fileOpenApps.ts#L134)), and the caller passes it straight to `fs.existsSync` ([fileOpenApps.ts:147](src/main/workspace/fileOpenApps.ts#L147)). An App Paths default stored as `%ProgramFiles%\Microsoft VS Code\Code.exe` always fails the existence check, and the `where` fallback cannot save it (per-user installs put `code.cmd` on PATH, which the `.exe` filter drops). On such machines the HKCU fix from `716e41c` has no effect and the editor is never detected. Either stop matching `REG_EXPAND_SZ` (treat as a miss) or — strictly better — expand `%VAR%` against `process.env` before the existence check. One line either way.

### 🟡 P3 — remaining polish (agreeing with the earlier follow-up)

- `handlerCache` ([fileOpenApps.ts:30](src/main/workspace/fileOpenApps.ts#L30)) still lives for the process lifetime; changing the default app for an extension while DeepChat is running keeps the editor ordering stale until restart. Impact is now ordering-only since `isRegisteredHandler` left the wire, but a short TTL would keep it consistent with `installedAppsCache`.
- Linux editors without a `launch.linux` override still rely on `gio launch <entry> <file>` forwarding the positional argument ([fileOpenApps.ts:360-363](src/main/workspace/fileOpenApps.ts#L360-L363)); JetBrains Toolbox desktop entries typically lack `%f`/`%F` field codes, so the IDE opens without the file. The same per-app override treatment the terminals got would fix it.
- Console-subsystem editors (`nvim.exe`, `emacs.exe`) get a fresh flashing console under `spawn(..., { detached: true, stdio: 'ignore' })` ([fileOpenApps.ts:366-373](src/main/workspace/fileOpenApps.ts#L366-L373)), and `stdio: 'ignore'` leaves interactive TUIs unusable anyway — consider launching them via a terminal host or excluding them on win32.

### 🟡 P3 — new: Linux terminal overrides use bare command names

`command = override.command ?? target.launchTarget` ([fileOpenApps.ts:355](src/main/workspace/fileOpenApps.ts#L355)) means a terminal installed via flatpak (wezterm/kitty are commonly flatpak-packaged) is no longer reachable through `gio launch`'s desktop-entry resolution — those installs regress from "opens in the wrong directory" to "fails to open" (now visible via the toast, which is the right failure mode, but still a regression). Falling back to `gio launch` when the bare command is not on PATH would preserve both paths.

Minor: the `chore` commit trimming stale comments also drops some still-informative JSDoc (e.g. field semantics on the registry types); harmless, though it could have been folded into the fix commit.

Requesting changes for the `REG_EXPAND_SZ` expansion only — everything else is non-blocking polish.

zerob13 (approved):
## Follow-up review after `bd00b9dd` (round 3)

Re-reviewed the delta `0336a157..bd00b9dd` (the `openInApp` module rewrite). Everything from my round-2 review is resolved, verified at HEAD `bd00b9dd`:

- ✅ **P2 `REG_EXPAND_SZ` expansion** — `readRegistryDefault` now anchors on the type column and expands `%VAR%` case-insensitively against `process.env` before the existence check ([detectors.ts:95-102](src/main/workspace/openInApp/detectors.ts#L95-L102), [detectors.ts:118-126](src/main/workspace/openInApp/detectors.ts#L118-L126)), with `%ProgramFiles(x86)%` handled and unknown vars left literal so the probe falls through gracefully. The `(Default)`-name matching (broken on non-English Windows) is gone as a bonus.
- ✅ **handlerCache staleness** — 60s TTL with freshness checks and failed-probe eviction ([index.ts:13](src/main/workspace/openInApp/index.ts#L13), [index.ts:80-83](src/main/workspace/openInApp/index.ts#L80-L83)); ordering refreshes without a restart.
- ✅ **`gio launch` dropping the file** — `.desktop` entries without `%f/%F/%u/%U` field codes are no longer listed at all ([detectors.ts:243-246](src/main/workspace/openInApp/detectors.ts#L243-L246), [linuxDesktopEntries.ts:48-65](src/main/workspace/openInApp/linuxDesktopEntries.ts#L48-L65), `%%` escape handled), so the silent "app opens without the file" mode is gone.
- ✅ **Console-subsystem editors** — Neovim is GUI-front-end only on win32 (with a comment explaining why), Emacs uses `runemacs.exe`; every remaining win32 executable is GUI-subsystem, so the detached `stdio: 'ignore'` spawn no longer flashes consoles.
- ✅ **Bare-name terminal spawns** — Linux detection only accepts validated `command -v` absolute paths ([detectors.ts:228-235](src/main/workspace/openInApp/detectors.ts#L228-L235), `SAFE_BINARY_REGEX`), and all launches go through argv arrays / `gio launch`; the flatpak "fails to open" regression is gone (flatpak-only terminals are simply not listed, with the reasoning documented in the registry).

The new delta itself: no blocking or should-fix findings. Security spot-checks came back clean — every spawn uses an argv array, never a shell string; the only `sh -c` is gated on registry-controlled names matching `SAFE_BINARY_REGEX`; `openFileWithApp` re-validates `appId` against the installed list; icon inputs are Launch-Services-resolved bundle paths, not renderer input; `sips` output goes to a private `mkdtemp` dir and is unlinked; JXA stdout is capped at 8 MB. IPC payload unchanged, and there is no new user-visible copy, so no i18n updates are needed.

Non-blocking nits — fold into a follow-up if you touch this again:

1. **Dangling test reference** — [fileOpenApps.ts:335](src/shared/workspace/fileOpenApps.ts#L335) says "see the test that asserts this", but nothing in `test/` references `WORKSPACE_FILE_OPEN_APPS` or `buildLaunchArgs`. Either add the ~15-line durable test (every non-Hyper terminal declares working-directory args) or drop the sentence.
2. **Icon extraction sits on the click path** — macOS detection unconditionally runs `extractMacOSIcons` (~2 subprocesses per detected app) at the end of every `detectInstalledApps()` call, and `openFileWithApp` calls `detectInstalledApps()` too, so after each TTL expiry the launch click pays ~2N subprocesses before `open -a` runs. Cache icons by bundle path and/or skip extraction when only launch targets are needed.
3. **Zed win32 detection dropped** — the old registry declared `executables: ['zed.exe']`; the rewrite keeps only bundleIds/binary/desktopIds ([fileOpenApps.ts:138-142](src/shared/workspace/fileOpenApps.ts#L138-L142)). Zed has shipped on Windows since 2025 — if the drop is intentional (e.g. the exe name was wrong), a word in the commit message would help; otherwise restore it.
4. **Format** — [fileOpenApps.ts:191](src/shared/workspace/fileOpenApps.ts#L191) is 103 columns; oxfmt would rewrap it.
5. **Commit message** — 80-char subject, missing the `(workspace)` scope its sibling commits use, and it bundles the module rewrite, registry schema change and the round-2 fixes in one commit — that forced full re-verification of previously reviewed code. Splitting refactor from fixes would make delta review much cheaper next time.
6. Minor: unused optional `platform` parameters on `detectInstalledApps`/`launchApp` (no caller passes them, no tests exist); the "never cache an empty result" comment understates the win32 re-probe cost on an editors-less machine — a short negative-cache TTL would help.

Approving — the round-2 blocking finding and all four P3s are addressed, and the nits above don't gate.

zhangmo8 (commented):
## Review Summary

I reviewed the full diff at HEAD `bd00b9d`. Overall the feature is well structured: the contract/route/service/client layers follow the existing workspace patterns, `appId` is validated against the locally detected installed list, macOS icon extraction is nicely batched, and i18n keys are complete. Three previous review rounds (CodeRabbit, zerob13, zhangmo8) already covered the terminal launch args, HKCU/HKLM detection order, `REG_EXPAND_SZ` expansion, and the Windows compile errors — I verified those are fixed at this HEAD and won't repeat them. Below are findings I believe are **not** covered by existing comments.

### 🟡 P2 — Referenced test does not exist in this PR

[`src/shared/workspace/fileOpenApps.ts:1930`](src/shared/workspace/fileOpenApps.ts#L1930) contains:

> Hyper is the one terminal here that documents a positional directory (`hyper <dir>`), so it needs no working-directory flag. Every other terminal must declare one; see the test that asserts this.

But this PR adds no test file — `git diff --stat` shows 35 changed files, none under `test/`, and a search for `buildLaunchArgs` / `fileOpenApps` / `openInApp` in `test/` finds nothing. The registry invariants ("every terminal except Hyper declares a working-directory strategy", "an app is offered only when both detect and launch exist") are exactly the kind of contract that will silently regress with the next registry edit. Please add a small unit test (e.g. for `buildLaunchArgs` and `appsForPlatform`) and update or remove the comment.

### 🟡 P2 — Launch success is resolved on `spawn`, not on successful exit

[`src/main/workspace/openInApp/launchers.ts:36-43`](src/main/workspace/openInApp/launchers.ts#L36-L43):

```ts
await new Promise<void>((resolve, reject) => {
  const child = spawn(command, args, { detached: true, stdio: 'ignore' })
  child.once('error', reject)
  child.once('spawn', () => {
    child.unref()
    resolve()
  })
})
```

`spawn` succeeds as soon as the process starts; a child that exits immediately with a non-zero code (e.g. `gio launch` fails to resolve the desktop entry, or the target binary errors out) still resolves the promise, and `WorkspaceViewer.vue` shows no error. The `macOpenA` branch (which `execFileAsync`s `open`) does report failures, so the Linux/Windows terminal+editor path is the gap. Previous rounds covered "failures now reject instead of silently falling back" for *probe* errors, but not post-spawn exit status. Consider listening for `exit`/`close` with the non-zero check after a short grace period (or at least `child.once('error')` + checking `exit` within a bounded window) so real launch failures surface through the existing `runOnOpenFile` error toast.

### 🟢 P3 — Preferred app silently degrades to system default

[`src/renderer/src/components/sidepanel/WorkspaceViewer.vue:383-389`](src/renderer/src/components/sidepanel/WorkspaceViewer.vue#L383-L389): when the persisted `preferredAppId` (from `localStorage`) no longer matches any detected app — app uninstalled, detection failed, or the stored sentinel `#system-default` — `preferredApp` is `null` and the main button silently falls back to `workspaceClient.openFile` with **no visual cue or notification**. The user who explicitly chose "Open in VS Code" last week gets the system default tomorrow without knowing why. Also, storing `#system-default` as the value of a key named `preferredAppId` is a type lie that will confuse future readers. Suggest: either keep showing the fallback but surface a one-time hint when the preferred app is unavailable, or only persist real app IDs and treat the sentinel as "clear preference".

---

Overall: solid work, and the multiple review rounds have converged well. The three points above are the remaining gaps I found; none block merge in my view, but the missing test is worth fixing in this PR since the comment promises it.

zhangmo8 (commented):
## Review Summary

Re-reviewed the full diff at HEAD `bd00b9dd` (35 files, +1515/−23). The design has converged well through the earlier rounds — contract/route/service/client layering matches the house patterns, `appId` is validated against the locally detected installed list, the macOS probe is batched into one JXA subprocess, and i18n is complete across all 20 locales. I verified the previously discussed findings (terminal launch args, HKCU/HKLM order, `REG_EXPAND_SZ` expansion, cache TTLs, `gio launch` field codes, console-subsystem editors) are addressed at this HEAD and won't repeat them. Below are items I believe are **not** covered by existing comments.

### 🔴 P1 — CI is red at the current head; the PR cannot merge

`gh pr checks` on `bd00b9dd`:

- `build` ❌ — TS errors in `src/main/workspace/openInApp/detectors.ts` at lines 124 and 191. CodeRabbit already posted inline comments with the exact fixes; I won't duplicate the details here, but note they are **still unaddressed at this HEAD**.
- `static` ❌ — `pnpm run format:check` fails on two files (see P2).
- `pr-required` ❌ — aggregates the above.

This matters because both post-failure reviews contradict the CI state: zerob13's APPROVED (Sep 3 17:31) and zhangmo8's last review (Sep 3 20:04, which states the compile errors were "verified fixed at this HEAD") were submitted after this run failed (16:22), yet `git show` of the head file still has `match` un-narrowed at line 124 and the unannotated map callback at line 191. Please re-run CI after the fixes and verify all three jobs before merge.

### 🟠 P2 — Formatting check fails on two files, one of them not mentioned before

`oxfmt --check` (repo config: single quotes, no semicolons, 100 cols) reports:

- `src/main/workspace/openInApp/detectors.ts` — the `detectDarwin` signature (lines 55–57) can collapse to one line, and the ternary at lines 124–126 exceeds 100 columns. Not flagged in any prior review.
- `src/shared/workspace/fileOpenApps.ts:191` — the rustrover `desktopIds` array is over 100 columns (already noted by zerob13; skipping details).

### 🟡 P3 — Linux apps detected via binary never get an icon

In `detectLinuxApp`, the binary branch returns `{ definition, launchTarget: binaryPath }` with no `iconDataUrl` — icons are only produced on the desktop-entry fallback branch. macOS (`extractMacOSIcons`) and Windows (`app.getFileIcon`) both attach real icons, so the Linux picker is the odd one out: for the most common installs (VS Code, Zed, terminals resolved via `command -v`) the dropdown renders the generic placeholder. Consider resolving an icon for the binary case too (e.g. look up a matching desktop entry's `Icon=` or extract from the binary), or document why binary-detected apps intentionally show no icon.

### ✅ Summary

Solid feature; the architecture and security posture are right and the earlier findings are genuinely addressed. The blocker is mechanical: the PR is unmergeable as-is because build, static, and pr-required are all failing at the current head. Apply the two TS fixes (per CodeRabbit's suggestions), run `oxfmt`, and re-verify CI before merging.

zerob13 (changes requested):
## Follow-up review after `fa8a961e6` (round 4)

Re-reviewed the delta `bd00b9dd4..fa8a961e6`. Every finding from the previous rounds is addressed in the code — verified at HEAD:

- ✅ **TS compile errors** — `match` is narrowed via a `registryType` guard ([detectors.ts:116-121](src/main/workspace/openInApp/detectors.ts#L116-L121)), the `detectWin32` map callback is annotated `Promise<DetectedApp | null>`, `detectDarwin` collapsed to one line. The `build` job is green again.
- ✅ **Missing registry test** — `test/main/shared/fileOpenApps.test.ts` now asserts the invariants the comment promised (detect/launch platform parity, working-directory strategy on every terminal except Hyper, `buildLaunchArgs` substitution, platform mapping), and the registry comment points at the real file.
- ✅ **Launch success semantics** — `desktopEntry` launches now go through `execFileAsync('gio', …)` with a 10s timeout and wait for exit ([launchers.ts:47-53](src/main/workspace/openInApp/launchers.ts#L47-L53)), so a missing entry or portal error rejects instead of resolving on spawn. The `exec` strategy keeps spawn-as-handoff with an honest rationale (a detached GUI process may outlive the app session; later exit codes are application runtime failures, not launcher failures) — I agree with that split; it mirrors `macOpenA`.
- ✅ **Preferred app degradation** — the stored id is kept (a missing probe is not proof of uninstall), "system default" clears the preference instead of persisting the `#system-default` sentinel, the legacy sentinel is migrated away on read, an app is remembered only after a successful launch, and an info toast fires when the stored app is not detected. `preferredAppUnavailable` is present in all 20 locales with `i18n.d.ts` regenerated; `notifyRenderer({ kind: 'info', … })` matches the `NotificationRequest` contract.
- ✅ **Linux binary icons** — `readLinuxBinaryIcon` prefers the desktop-entry icon and falls back to `readFileIconDataUrl(binaryPath)`; the win32 path reuses the same helper.
- ✅ **`openFile` swallowing `shell.openPath` errors** — now rejects with the error message ([index.ts:798-807](src/main/workspace/index.ts#L798-L807)), covered by a new service test.

### 🔴 P1 — CI is red at this head: the new Linux icon test never exercises `readLinuxBinaryIcon`

`gh pr checks` at `fa8a961e6`: `test-main` ❌ and `static` ❌ (`build` and `test-renderer` are green).

**test-main** — `test/main/workspace/openInApp/linuxBinaryIcon.test.ts` fails 2 of 4 tests (reproduced locally). The mock resolves the `command -v` probe from the wrong argument position:

```ts
const commandArgs = args[1] as string[] | undefined
const command = commandArgs?.[2] ?? ''   // wrong index
```

The real call is `execFileAsync('/bin/sh', ['-c', 'command -v "code"'], …)` ([detectors.ts:226](src/main/workspace/openInApp/detectors.ts#L226)), so the command string is `args[1][1]`, not `args[1][2]`. The mock therefore always answers "not found", `resolveLinuxBinary` returns `null`, and the binary branch is skipped. The `desktopIds` fallback doesn't save it either: `desktopEntryAcceptsFiles: vi.fn()` returns `undefined` (falsy), so that branch `continue`s too — **vscode is never detected at all**. Consequences:

- Tests 1–2 assert on `vscode?.iconDataUrl` of an app that was never found → fail with `undefined`.
- Tests 3–4 ("omits the icon when getFileIcon returns empty/fails") pass vacuously (`undefined === undefined`).
- Net: all four tests miss the code path they were written for; `readLinuxBinaryIcon` has zero effective coverage.

One-character fix: `commandArgs?.[1]`. After that, tests 3–4 will actually assert something.

**static** — `oxfmt --check` flags exactly one file in the repo: `test/renderer/components/WorkspaceViewer.test.ts`. Run `oxfmt` on it and re-push.

Both are mechanical, but the PR cannot merge while `static` and `test-main` are red.

### 🟡 P3 — Repeated unavailability toasts (optional)

`handleOpenFile` fires the `preferredAppUnavailable` info toast on every primary click while the stored app stays undetected. The earlier suggestion was a one-time hint. Re-notifying is defensible given "probe miss ≠ uninstalled", so this is optional polish — e.g. notify once per session per stored id.

### ✅ Everything else

The preference state machine (store → remember-on-success → clear-on-system-default → migrate sentinel) is coherent and each branch has a renderer test; the ~540 added test lines map one-to-one onto what the previous rounds asked for (registry invariants, launcher semantics, preference behavior, openFile rejection) with no padding; no over-engineering — every hunk answers a confirmed finding. `WorkspaceService.openFile` now rejecting on `shell.openPath` errors is the correct behavior change (the renderer already surfaces `openFailed`) — worth one line in the PR description.

Requesting changes for the CI failures only.

zerob13 (approved):
## Round-5 delta review (fa8a961e6 → 01a6daf13)

Verified all round-4 findings are fixed at 01a6daf13:

✅ **P1 — CI red (test-main + static)** — fixed
- `linuxBinaryIcon.test.ts`: mock index corrected `commandArgs?.[2]` → `commandArgs?.[1]`, matching the real `execFileAsync('/bin/sh', ['-c', 'command -v "code"'])` call shape. Locally 4/4 passed, and no longer vacuous — `vscode` is actually detected now, so the `iconDataUrl` assertions and the `getFileIconMock).not.toHaveBeenCalled()` negative assertion are both effective.
- `workspaceService.test.ts`: the openFile-reject test now creates the file first (`fs.writeFileSync(filePath, '')`). Necessary fix, confirmed by temporarily reverting locally: without it the test fails (promise resolves instead of rejecting) because `normalizePathForAccess` falls back to `path.resolve` when the file doesn't exist, skipping the `/tmp` → `/private/tmp` symlink resolution and mismatching the registered workspace prefix → path treated as blocked. With the file present: 17/17 pass through the real reject path.
- `WorkspaceViewer.test.ts`: oxfmt single-line normalization; `oxfmt --check .` is green repo-wide (2903 files), 15/15 tests pass.

🟡 **P3 (optional) — repeated `preferredAppUnavailable` toast on every click while the stored app is undetected** — not addressed; the re-notify stance accepted in round-4 stands.

Local verification at 01a6daf13: linuxBinaryIcon 4/4, workspaceService 17/17, WorkspaceViewer 15/15, `oxfmt --check .` clean. CI checks were still in progress at review time; all three local reproductions of the previous CI failures are green.

zerob13 (approved):
## Round-6 delta review (01a6daf13 → 2e943d2cb)

Single commit, `src/renderer/src/components/markdown/useMarkdownLinkNavigation.ts` only (+6/−1): wrap the final `workspaceClient.openFile(resolution.path)` in try/catch, `console.warn` + `return false` on rejection. **Approving** — verified below.

✅ **Fix is necessary, not defensive noise** — the main-side handler really rejects: `shell.openPath` returns an error string (missing file, no registered handler, permission) and the service throws it (`src/main/workspace/index.ts:794-806`). On the renderer side, two of the three `navigateLink` call sites are fire-and-forget (`void navigateLink(...)` in `MarkdownRenderer.vue` and `MessageBlockSearch.vue`) and the third awaits inside an async click handler without a catch (`LinkNode.vue:42-44`), so any `openFile` rejection was escaping as an unhandled promise rejection before this commit.

✅ **Consistent with the file's own conventions** — the new catch block mirrors `openExternal` exactly (try/catch, `[markdown-links]`-prefixed `console.warn` with path + error, `return false`), and matches the existing resolution-failure path at the top of `openLocalFile`. No new pattern introduced.

✅ **Contract preserved** — `false` already means "not handled" across every `navigateLink` branch, and no caller branches on the return value, so there's no behavioral regression; failures now degrade to a warn log exactly like external-link failures do.

✅ **Adjacent path checked** — the in-session branch (`sidepanelStore.selectFile`) is a synchronous store action, so it has no same-class unhandled-rejection hazard.

✅ **Checks run at 2e943d2cb** — `oxfmt --check` clean on the touched file; related suites green (`MarkdownRenderer.test.ts` 26/26, `MessageBlockSearch.test.ts` 6/6, 32 total); repo typecheck exit 0. No new tests required for this delta: the composable has no existing test file and this is a single error-path guard, so demanding a bespoke harness here would be implementation-coupled coverage, not regression protection.

No findings. Round-4/5 items remain resolved at this head.

PR #2230, opened 2026-09-03
perf(tape): faster turns, atomic message facts

yyhhyyyyyy:
## Summary

Every provider request in a long session read the whole Tape three times, re-appended the same tool facts every round, and let the transcript and its Tape facts commit separately. This is the "stop the bleeding" slice of the Tape review: no schema, fact name, provenance key or contract change. Tape semantics are equivalent by construction.

Measured on a 10k-entry encrypted session (66.8 MB database), three whole-session reads per turn:

| | first turn | steady state |
|---|---|---|
| `dev` | 1043 ms | 961 ms |
| this branch | 162 ms | 15.5 ms |

## What changed

- **Page cache 16 → 64 MiB.** SQLCipher decrypts every page it loads; with the default cache the per-turn reads of a large session evicted each other and the same pages were decrypted again every turn (640 ms → 16 ms for one whole-session read). A ceiling, not an allocation.
- **Hot-path reads filter in SQL.** The reconciler and ViewManifest source maps read the entire session with `SELECT *` and discarded everything except messages, tool facts, anchors and `message/retracted` events, about 60% of the bytes in a 1000-turn session. `getEffectiveViewInputRows` applies the same predicate the effective view uses; a real-SQLite test asserts both select identical rows.
- **Transcript writes and Tape facts commit together.** `createUserMessage`, `finalizeAssistantMessage`, `setMessageError` and `updateMessageContent` autocommitted each statement and appended the Tape fact last, so a failed append left a message the Tape never saw; for user edits the replacement fact was lost for good. They now run in one transaction, as `tape-system.md` promises. Callers holding a transaction nest it as a savepoint; no Execution Journal write is reachable inside.
- **Tool facts appended once per run.** `afterRoundPersisted` replayed every settled tool block after every round (O(n²) appends, each neutralized by two idempotency lookups). The run now skips keys it already appended; a skipped input is exactly an idempotent hit, and a failed append is retried. The terminal replay in finalize is kept: it is the only live writer for ACP, permission-denied and error paths.
- **Backfill skipped when nothing changed.** `ensureSessionTapeReady` remembers what the last completed backfill saw (message count, `max(updated_at)`, Tape incarnation, Tape head). Every transcript writer stamps `updated_at` or changes the count, so writes that bypass Tape still trigger a full backfill; the snapshot is also distrusted across same-millisecond writes, clock regression, a swapped database file (sync restore, maintenance reopen) and host transactions. Mainly helps steer/resume and idle-time IPC.

## Compatibility

- Existing Tapes read identically; `getBySessionExcludingContext` stays for readers that need events.
- Behavior change: a Tape append failure inside finalize now rolls the assistant row back to `pending` instead of committing without its fact. The only non-IO trigger is a tool fact whose name collides with a reserved namespace (`execution/...`); MCP names are validated to `^[a-zA-Z0-9_-]+$`, and that collision already breaks backfill on every later turn today. Follow-up: isolate tool fact names from reserved namespaces.
- Deferred, pre-existing: metadata-only assistant updates never reach Tape; the structural items from the review (double-write of message bodies, O(N²) manifest `excluded` lists, per-request effective view fold).

## Validation

- `test/main`: all 634 files, 0 new failures. Three pre-existing failures reproduce on untouched `dev`: `sessionDataMigrations.sqlite` (2), `deepchatPendingInputsTable` (4, Electron ABI only), `buildCli` (4, environment)
- Real SQLite under the Electron ABI: `test/main/tape`, `test/main/session`, ACP adapter, process and compaction suites, 792 passed
- Every fix ships with a regression test that fails on the parent commit (transcript atomicity, tool fact retry, backfill invalidation triggers, SQL filter vs. predicate, unreadable bootstrap anchor). The two new real-SQLite suites are added to the CI native allowlist; `postinstall` builds the module for the Electron ABI, so they would otherwise be skipped silently.

Commits are independent and individually revertible; please merge with a merge commit rather than squash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate tool facts from being saved across processing rounds while allowing failed saves to retry.
  * Improved transcript and Tape update reliability by making related changes atomic.
  * Corrected Tape view reconstruction to use only entries that affect the effective conversation state.
  * Prevented corrupted Tape metadata from causing read failures.
  * Reduced unnecessary Tape reconciliation work when session data is unchanged.

* **Performance**
  * Increased SQLite’s cache allocation to improve encrypted session data access.

* **Tests**
  * Added coverage for atomic updates, reconciliation skipping, effective Tape views, retries, and corrupted metadata.

### DIFF
--- a/.github/workflows/prcheck.yml
+++ b/.github/workflows/prcheck.yml
@@ -176,6 +176,8 @@ jobs:
           test/main/session/data/tapeLineage.test.ts
           test/main/session/data/tapeRecall.test.ts
           test/main/session/data/tapeViewReplay.test.ts
+          test/main/session/data/tables/deepchatTapeEntriesTable.test.ts
+          test/main/session/data/transcriptAtomicity.test.ts
 
       - name: Validate encrypted OCR artifact storage
         env:
--- a/src/main/agent/deepchat/runtime/process.ts
+++ b/src/main/agent/deepchat/runtime/process.ts
@@ -42,7 +42,10 @@ import {
 } from '@/agent/deepchat/loop/loopRun'
 import { emitDeepChatLoopNotification } from '@/agent/deepchat/loop/notificationObserver'
 import type { OutputSink } from '@/agent/deepchat/loop/ports'
-import { buildTapeToolFactInputs } from '@/tape/application/factPersistence'
+import {
+  buildTapeToolFactInputs,
+  buildTapeToolFactProvenanceKey
+} from '@/tape/application/factPersistence'
 import {
   ExecutionJournalCorruptionError,
   ExecutionJournalError,
@@ -1000,6 +1003,9 @@ export async function processStream(params: ProcessParams): Promise<ProcessResul
   logger.info(`[ProcessStream] start session=${io.sessionId} message=${io.messageId}`)
   let eventCount = 0
   let pendingToolSurfaceCandidateRelease: PendingToolSurfaceCandidateRelease | null = null
+  // Every round replays all settled tool blocks of the message; facts appended by an earlier
+  // round of this Run are skipped here instead of round-tripping the store's idempotency check.
+  const appendedToolFactKeys = new Set<string>()
   const commits: DeepChatLoopCommitCallbacks<StreamState, ProcessResult, ProcessResult> = {
     updateOutput: ({ outcome }) => {
       if (outcome === 'halted' || firstProviderRoundReady || state.blocks.length === 0) {
@@ -1028,7 +1034,10 @@ export async function processStream(params: ProcessParams): Promise<ProcessResul
       const toolResultBySourceId = new Map<string, TapeToolResultFactReference>()
       try {
         for (const input of buildTapeToolFactInputs(record)) {
+          const factKey = buildTapeToolFactProvenanceKey(input)
+          if (factKey && appendedToolFactKeys.has(factKey)) continue
           const receipt = await params.io.tapeToolFactWriter.appendToolFact(input)
+          if (factKey) appendedToolFactKeys.add(factKey)
           if (input.provenance.source === 'tool_result' && receipt.toolResult) {
             toolResultBySourceId.set(input.provenance.sourceId, receipt.toolResult)
           }
--- a/src/main/data/connectionConfig.ts
+++ b/src/main/data/connectionConfig.ts
@@ -2,14 +2,6 @@ import type Database from 'better-sqlite3-multiple-ciphers'
 
 export const SQLCIPHER_COMPATIBILITY_VERSION = 4
 
-/**
- * Page cache ceiling in KiB (negative `cache_size` values are KiB). SQLCipher decrypts every page
- * it loads, so the whole-session Tape reads that run before each provider request pay the
- * decryption cost again whenever the session no longer fits in the cache. 64 MiB keeps a
- * 10k-entry session resident; the cache only grows to the pages actually touched.
- */
-export const SQLITE_PAGE_CACHE_KIB = 65_536
-
 export function configureSQLCipherCompatibility(db: Database.Database): void {
   db.pragma("cipher='sqlcipher'")
   db.pragma(`legacy=${SQLCIPHER_COMPATIBILITY_VERSION}`)
@@ -26,5 +18,7 @@ export function configureSQLiteConnection(db: Database.Database, password?: stri
   }
 
   db.pragma('journal_mode = WAL')
-  db.pragma(`cache_size = -${SQLITE_PAGE_CACHE_KIB}`)
+  // 64 MiB page cache (negative = KiB). SQLCipher decrypts every page it loads, so the
+  // whole-session Tape reads before each provider request must stay resident to be cheap.
+  db.pragma('cache_size = -65536')
 }
--- a/src/main/data/connectionConfig.ts
+++ b/src/main/data/connectionConfig.ts
@@ -18,7 +18,9 @@ export function configureSQLiteConnection(db: Database.Database, password?: stri
   }
 
   db.pragma('journal_mode = WAL')
-  // 64 MiB page cache (negative = KiB). SQLCipher decrypts every page it loads, so the
-  // whole-session Tape reads before each provider request must stay resident to be cheap.
+  // 64 MiB page cache ceiling (negative = KiB) for every connection opened through this helper.
+  // It is sized for the main database: SQLCipher decrypts every page it loads, so the whole-session
+  // Tape reads before each provider request must stay resident to be cheap. Other connections
+  // (importer, OCR store, utility hosts) only grow the cache to the pages they actually touch.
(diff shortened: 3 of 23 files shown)


zerob13 (approved):
## Review: `perf(tape): faster turns, atomic message facts`

Reviewed all 8 commits at `c159c1aa` against `dev` (`ec63c768`). I independently verified the load-bearing claims in the PR description against the code:

- **Transaction atomicity is real.** The four transcript writers (`createUserMessage`, `finalizeAssistantMessage`, `setMessageError`, `updateMessageContent`) now run through `runInDatabaseTransaction` ([transcript.ts:233-235](src/main/session/data/transcript.ts#L233-L235)); better-sqlite3 degrades nested calls to savepoints, so callers that already hold a transaction (e.g. pending inputs) nest safely. The "no Execution Journal write is reachable inside" claim holds — the journal service asserts against host transactions and none of the four wrapped bodies write `execution/*` names.
- **The SQL filter is semantically exact.** `getEffectiveViewInputRows` ([tapeEntryStore.ts:1264-1277](src/main/tape/infrastructure/sqlite/tapeEntryStore.ts#L1264-L1277)) selects `message` / `tool_call` / `tool_result` / `anchor` plus `event`/`message/retracted`. The tape entry kind union is exhaustive, `buildEffectiveTapeView`'s message records depend only on those row kinds, and the recall/search consumers that need event rows correctly stay on `getBySessionExcludingContext`. The SQL-vs-predicate equivalence test is the right lock for the deliberate double-maintenance.
- **Skip-set keys match live append keys.** `buildTapeToolFactProvenanceKey` ([factPersistence.ts:276-289](src/main/tape/application/factPersistence.ts#L276-L289)) builds the identical provenance key a live `appendToolFact` assigns (the round path passes no `supersedesEntryId`), so a skipped input is exactly an idempotent hit, and a failed append stays out of the set and retries next round. ToolSurface release is consumed in the same round that first appends the fact, so it is unaffected.
- **Backfill snapshot invalidation signals are complete.** Every `deepchat_messages` UPDATE stamps `updated_at`, and writes that bypass Tape change the row count. The distrust guards (clock regression, swapped database file, host transactions) are documented at the check site.
- **The disclosed behavior change is accurate.** The reserved-namespace throw inside finalize is effectively unreachable for tool facts: MCP/plugin tool names are validated to `^[a-zA-Z0-9_-]+$` and cannot contain `/`.

### Minor observations (non-blocking)

1. **🟡 Clock-regression edge in the fast path** — [reconcilerService.ts:65-73](src/main/tape/application/reconcilerService.ts#L65-L73) distrusts the snapshot only while the clock is at-or-behind the remembered `maxUpdatedAt`. A no-Tape transcript write stamped *during* a clock-regression window and observed *after* the clock recovers can still compare equal (same count, same `maxUpdatedAt`) and be served a stale view. Triggering this requires a mid-process recovery plus a precise regression window, and the code comment already documents the trust model — fine to address in a follow-up. If you want it airtight, permanently distrust the snapshot after any observed regression instead of re-trusting once `Date.now()` moves ahead again.
2. **🟡 `isEffectiveViewInputRow` is exported from production code but only consumed by tests** ([effectiveSemantics.ts:12](src/main/tape/domain/effectiveSemantics.ts#L12)). It is the oracle for the SQL-equivalence test, so the export is defensible; flagging the prod/test boundary only.
3. **🟡 `cache_size = -65536` applies to every connection** configured through `configureSQLiteConnection` ([connectionConfig.ts:23](src/main/data/connectionConfig.ts#L23)), not only the session DB (sync importer, scheduler, etc. share it). Since a negative `cache_size` is a ceiling rather than an allocation and session data uses a single shared main connection, this is fine in practice — noting only that the comment's motivation understates the blast radius.

Validation: static review (the review worktree had no `node_modules`); format/lint/typecheck/test results are per the PR description and CI — the two new real-SQLite suites added to the CI allowlist with `DEEPCHAT_REQUIRE_NATIVE_SQLITE` (fail-not-skip) fix a real silent-skip problem, which is a good catch.

No P1/P2 findings. Well-scoped "stop the bleeding" slice; the description matches the code.

zerob13 (approved):
## Delta re-review — `c159c1aa6` → `684505ac0`

Re-reviewed the single new commit `684505ac0` ("fix(tape): compare transcript rows, not clocks, before skipping backfill") on top of my earlier approve at `c159c1aa6`.

### What the change does

Replaces the `(messageCount, maxUpdatedAt)` skip snapshot with a SHA-256 digest over per-row `(id, orderSeq, status, updatedAt)` and drops both clock guards (`observedAt` trust + strictly-in-the-past remember). Comment-only updates in `connectionConfig.ts` (cache-ceiling rationale) and `deepchatMessages.ts` (invariant restated for the digest).

### Verified

1. **The clock-regression edge from my previous review is closed.** A bypass write stamped by a regressed clock (e.g. `updateMessageStatus` on a failed steer) changes the row's `status`/`updatedAt`, which the per-row digest sees regardless of which direction time moved. The new test (`backfills again after a bypass write stamped by a clock that had stepped back`) reproduces exactly that scenario: `u2` re-stamped `2_000 → 3_000` below `u1`'s `4_000` — count and max unchanged, old snapshot skips, digest doesn't.
2. **Digest fields cover every reachable fact-identity mutation.** Message-fact provenance is `message:{id}:revision:{status}:{updatedAt}` for non-`sent` rows (all fields in digest), and the default `(source.type, id, seq, kind, name)` = `(id, role)` for `sent` rows — `role` has no update path (set only at insert). Content/metadata/traceCount changes re-stamp `updated_at` (digest-visible) and, for `sent` rows, don't alter fact identity anyway — a re-run would be provenance-deduped to a no-op, so skipping them is harmless.
3. **No perf regression to the PR's goal.** The digest is O(n) SHA-256 over ~50 bytes/row of records that were already read and sorted before the skip check; the expensive part (fact diff + appends + backfill event) is still what the skip avoids.
4. **Safe failure direction.** A spurious digest change only forces a full backfill re-run, which provenance dedup keeps a no-op.
5. **Transaction/swap guards retained.** Host-transaction remember guard, incarnation and head checks unchanged; the snapshot shape is a private in-memory map — no persisted format, contract, or IPC change.
6. **Tests are right-sized.** Net −16 lines: the two obsolete clock-guard tests are replaced by one targeted regression test, and the steer-retry test now asserts the actual Tape entries (2 message facts) instead of a count.

### Non-blocking note (P4)

The only residual edge I can construct: a bypass `updateMetadata` that flips a compaction row's `compactionStatus` while `Date.now()` collides with the row's previous `updated_at` millisecond — digest and indicator provenance (`message:{id}:compaction_indicator:{status}:{updatedAt}`) would both stay unchanged. That requires an exact-millisecond collision (sub-ms race or a clock returning to the exact prior ms) and is strictly narrower than the hole this commit fixes; fine to leave as-is.

**Verdict: approve** — refreshed at `684505ac0`.

PR #2232, opened 2026-09-03
refactor(tape): remove dead weight and contract-test the table mock

yyhhyyyyyy:
## Summary

This branch removes Tape code that nothing in production reaches, makes the test double that ~290 Tape tests run on prove it still matches the SQLite store, and trims two things the Tape Inspector did for no reader. No schema, fact name, provenance key or hash format changes, and no Tape write-path behavior changes. The only runtime behavior changes are a stricter session check on one Inspector route and fewer hashes computed for Inspector list rows.

Please merge with a merge commit rather than squash; each commit stands alone and can be reverted independently.

## Legacy compatibility re-exports

The seven `@deprecated` modules under `src/main/session/data/` (`tape`, `tapeEffectiveView`, `tapeFacts`, `tapeViewManifest` and the three `tables/deepchatTape*` files) only re-exported symbols from `src/main/tape/`. Production code stopped importing them after the layering refactor; eight test files were the last consumers and now import the canonical modules. `layerBoundaries.test.ts` loses the frozen export-contract map, the "production must not import legacy paths" guard and their negative fixtures, plus the two storage-boundary exceptions that named the deleted table modules. Every other layer guard (domain purity, Memory route port, concrete facade isolation, physical storage boundary) is unchanged. `tape-system.md` and `session-management.md` now say Tape has no compatibility path under `session/data`.

## Delta-merge fork service

`TapeForkService` (`createFork` / `mergeFork` / `discardFork` and the two `recordExternalFork*` receipts) has had no production caller since #1980 moved subagent finalization to `subagent/tape_linked`; the in-process delta-merge fork was never wired to a product path. It was also broken: its merge filter only excludes the reserved namespaces that existed at the time, so a fork containing `provider/attempt_completed` or `view/tool_surface` cannot be merged because generic `append` rejects those names and the transaction rolls back. Every real provider run inside a fork would have failed to merge. The mock-backed tests did not catch this because the harness `append` had no namespace guard.

The service, its facade forwards, the `TapeForkHandle` type, the fork-only `appendMessageRecordForSession` helper and the fork-only `getBySessionUpToEntryIdExcludingContext` store method are removed from the port, the SQLite store and the test harness. Read paths are untouched: `fork/merge` and `fork/discard` rows written by v1.0.5 to v1.0.9 subagent finalization still parse as legacy child links, the Inspector still classifies `fork/*` under lineage, and `'fork'` stays a source type. `tape-system.md` describes that read-only compatibility instead of a fork lifecycle.

## Reserved namespaces and the test mock

The reserved-namespace rule has two halves: a generic append must not produce a strict writer's facts, and a strict writer must only produce its own. Both halves now live in one domain function, `assertTapeAppendAuthorized` in `domain/reservedNamespaces.ts`. The SQLite store calls it from `appendInternal` and its five strict writer methods drop their own name checks; predicates, order and error messages are unchanged. The strict half is one exhaustive `switch`, the generic half is the original sequence of `if` statements. `isExecutionJournalReservedName` no longer claims `name is string`, which narrowed `name` to `null | undefined` after a negative check.

`createTapeTableMock` in the test harness had drifted from the store in several ways that hid real bugs: no namespace guard on `append` (the fork bug above), three of the seven summary anchor names in `getLatestSummaryAnchor`, bootstrap detection that only recognised `session/start`, a case-sensitive single-phrase `search`, and recovery events ordered by run instead of by entry. The mock now calls the same `assertTapeAppendAuthorized`, its strict writers pass their namespace, and the other drifts are fixed. Five `tapeRecall` fixtures that relied on the missing guard to write `provider/attempt_completed`, `view/programmatic_tool_surface` and a `context` row through generic append now use the strict writers.

New `tapeTableMockContract.test.ts` runs identical operations against the mock and a real `:memory:` store and compares every reader the mock implements with its own logic: append ids and idempotency, bootstrap and anchors, all reserved names through generic append and an off-namespace name through each strict writer, event/message/effective-view readers, search, nested transaction rollback, execution-journal recovery with interleaved runs and nested operations, cross-Tape effective reads and session deletion. Journal and skill facts are produced through the production writers so the fixtures match what runs in the app. Reverting any one of the mock fixes makes the suite fail. It runs in the native CI job, together with `tapeToolSurfacePersistence` and `compactionUsage`, whose real-SQLite cases were never in the allowlist.

## Tape Inspector

`sessions.subscribeTapeInspectorHead` was the only Inspector route that did not go through `SessionQuery.requireSession`: it validated the renderer caller and handed the session id straight to the head watcher, so an unknown session either subscribed (tape rows still present) or failed with the misleading "bootstrap is missing" error. The route now checks the session first; the watcher and its poll loop are unchanged.

`projectTapeInspectorFact` computed SHA-256 over `payload_json` and `meta_json` for every list row, and View rows hashed them a second time in their own branch. The renderer never reads those list-row hashes; it shows `detail.hashes`, which the detail projection computes separately. The cost was paid on every page and on every row scanned by the filter fallback, including rows that never made it into the page. List rows now carry only the View `manifestHash` and `integrity`, which the detail pane uses to announce its integrity section before the detail loads. `tape-trace-inspector/spec.md` records the split.

## Compatibility

- No schema, fact name, provenance key or hash format changes. Existing Tapes read identically; `fork/*` history remains readable.
- No IPC contract changes. `TapeInspectorFactRecord.hashes` and its fields were already optional.
- Behavior changes: subscribing to an unknown session now fails with `Session not found` (previously `bootstrap is missing`, or it succeeded when tape rows outlived the session row); Inspector list rows omit `payloadHash` and `metaHash`.
- Historical SDD folders under `docs/architecture/` are left as written, per `spec-driven-dev.md`.

## Validation

- `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, `pnpm run typecheck` (node and web): pass
- `test:main`, run in directory chunks because the one-shot runner is SIGKILLed in this sandbox: 621 of 634 files executed, 8192 tests passed, 0 failures. 29 SQLite-gated files skip under the Node ABI by design. 15 `test/main/scripts/*` files (build and packaging script tests, untouched here) could not be started in the sandbox.
- Real SQLite under the Electron ABI (`ELECTRON_RUN_AS_NODE=1 DEEPCHAT_REQUIRE_NATIVE_SQLITE=1`) on the updated native allowlist: 13 files, 292 passed, 0 skipped
- `pnpm run test:memory:scope`, `pnpm run test:memory:type`: pass
- `test/renderer/components/tape-inspector`: 8 files, 114 passed

## Summary by CodeRabbit

- **Changes**
  - Removed Tape fork creation, message appending, merging, and discarding capabilities.
  - Tape event writing now enforces stricter namespace and event authorization.
  - Session Tape Inspector subscriptions verify that the requested session exists before starting.
  - Tape Inspector list views no longer display payload and metadata hashes; detailed views continue to provide integrity information.
  - Removed legacy Tape compatibility access paths.

- **Documentation**
  - Updated Tape architecture and integrity documentation to reflect current behavior and compatibility support.

- **Tests**
  - Expanded coverage for Tape storage consistency, search behavior, authorization, and Inspector integrity data.


zerob13 (approved):
## Review at `f3a53f6062` (all 7 commits, base `f25967e05`)

Reviewed the full delta (37 files, +1004/−1647) together with the surrounding code. Every load-bearing claim in the PR description checks out:

- **Legacy re-exports** — the seven `@deprecated` modules were pure re-exports, had zero production importers at the merge-base, and exactly 8 test files consumed them; all 8 migrations are pure import-path swaps, and nothing in `src/`, `test/` or `scripts/` imports the deleted specifiers anymore.
- **layerBoundaries.test.ts** — the 277 deleted lines are exactly the legacy export-contract map, the two legacy guards with their negative fixtures, and the two storage exceptions; the remaining guards are intact.
- **TapeForkService removal** — no production caller since #1980 (verified at the merge-base), and zero dangling references at HEAD (`TapeForkService`, `createFork`/`mergeFork`/`discardFork`, `recordExternalFork*`, `TapeForkHandle`, the fork-only store methods). The unrelated `sessions.fork` route (`lifecycle.forkSession`) is untouched.
- **The removed fork merge was indeed broken** — its filter only excluded the reserved namespaces that existed at the time, while the store's generic append (already at the merge-base) rejects `provider/attempt_completed`, `view/tool_surface` and friends, so any real provider run inside a fork would have rolled the merge back. Good riddance.
- **Read-path compatibility** — historical `fork/merge`/`fork/discard` rows still parse via the legacy external-link snapshot ([traceInspectorProjection.ts:320-331](src/main/tape/application/traceInspectorProjection.ts#L320-L331)), lineage classification keeps `fork/*`, and `'fork'` remains a source type in the domain and shared types.
- **No schema / fact-name / provenance-key / hash-format changes**; the store's append checks are byte-identical to the old inline block, now shared with the mock through `assertTapeAppendAuthorized`. Net write-path behavior: unchanged.
- **Inspector behavior changes** — the head-subscription route now rejects unknown sessions with the same `Session not found` error as the other five Inspector reads ([routes.ts:494](src/main/session/routes.ts#L494)), and the renderer already degrades gracefully on rejection. List-row `payloadHash`/`metaHash` were already `.optional()` in the route contract, no consumer reads them from list rows, and the detail payload still computes full hashes.
- **The contract test is real, not tautological** — it drives a real `:memory:` better-sqlite3 store and the mock through identical operations and compares every reader the mock implements, gated by a smoke check and fail-closed under `DEEPCHAT_REQUIRE_NATIVE_SQLITE=1`, wired into the native CI allowlist. Each of the six mock fixes has a discriminating fixture. Net test volume decreases (contract file added vs. the 562-line fork suite + 277-line guard map deleted).

I also spot-checked the motivation: the harness had drifted badly enough that most recent store-touching commits also patched the mock — so extracting the namespace rule into one domain function consumed by both the store and the mock is the right fix, not gold-plating.

**No P1/P2 findings.** Three non-blocking nits:

1. **Mock search parity edges** — [tapeTestHarness.ts:597-600](test/main/session/data/tapeTestHarness.ts#L597-L600) accepts an unbounded token list while the store caps at 256 clauses (`MAX_TAPE_SEARCH_TOKEN_CLAUSES`, [tapeEntryStore.ts:383](src/main/tape/infrastructure/sqlite/tapeEntryStore.ts#L383)), and the store's `LIKE` folds ASCII case only while the mock's `toLowerCase()` folds Unicode. Test-double-only, and no fixture discriminates them — but this is the exact drift class this PR kills, so a one-line cap mirror (plus a comment on the folding caveat) would close it out.
2. **Stale baseline doc** — [zero-inbound-candidates.md:24-30](docs/architecture/baselines/zero-inbound-candidates.md#L24-L30) still lists all seven deleted modules as deletion candidates. It is a dated snapshot, but this file's whole job is describing the current dead-code frontier — regenerate it or add a one-line superseded note.
3. **"No compat re-creation under `session/data`" is now docs-only** — with the import guard gone, the invariant lives in [tape-system.md](docs/architecture/tape-system.md) alone. Re-imports of the deleted modules still fail typecheck, and a re-created `tables/deepchatTape*` forwarder would still trip the physical-storage guard; only a re-created `session/data/tape.ts` forwarding to application modules would slip through every remaining guard. Acceptable trade for deleting dead policing — noting it so the enforcement gap is a conscious decision.

Net −643 lines, all removals are main-process-internal (no IPC/preload surface ever exposed the fork service or the legacy modules), both behavior changes are explicitly called out in the PR body, and I found nothing breaking that is not explained. Approving.

zerob13 (approved):
## Delta review at `14adcea965` (5 commits since my review at `f3a53f6062`)

All three nits from my previous review are addressed, and correctly. Each fix was verified against the store's real behavior, not just the mock:

- **Mock search parity** — both edges now mirror the store exactly:
  - *Clause cap*: the store adds the per-token clause only when `1 < tokens ≤ MAX_TAPE_SEARCH_TOKEN_CLAUSES` and drops it above the cap ([tapeEntryStore.ts:411](src/main/tape/infrastructure/sqlite/tapeEntryStore.ts#L411)); `matchTokens` in the harness reproduces that predicate verbatim, including the `> 1` half ([tapeTestHarness.ts:601](test/main/session/data/tapeTestHarness.ts#L601)). The new fixtures discriminate the interesting semantics: `tokensAtCap` is exactly 256 tokens in reversed order (only the token clause can match it) and `tokensOverCap` is 257 reversed (nothing can), so a truncate-to-cap mock regression would fail the pair.
  - *ASCII folding*: the store passes `query.trim()` straight into a raw `LIKE … ESCAPE '\'` with no pre-lowering, so case folding is SQLite's ASCII-only rule; `foldAscii` reproduces that instead of the old Unicode `toLowerCase()` ([tapeTestHarness.ts:595](test/main/session/data/tapeTestHarness.ts#L595)). `unicodeCaseKept` (`éclair` must not find `Éclair`) and `unicodeExact` pin that boundary on both sides of the contract.
  - The delta's only production change is adding `export` to that constant so the harness and the contract test bind to a single source of truth — the right shape for this fix, no runtime change.
- **Stale baseline docs** — regenerated: the seven deleted modules are gone from `zero-inbound-candidates.md` (candidate count 20 → 14; the one new candidate and the raw-timer drift come from other merged work, as expected from a regen).
- **Docs-only re-export invariant** — enforced again: `findTapeReexportViolations` ([layerBoundaries.test.ts:181](test/main/tape/layerBoundaries.test.ts#L181)) walks `session/data` with the TS AST and rejects re-exports of anything resolving under `tape/` — direct `export … from`, and re-exports of imported bindings including renamed, namespace, type-only and `export default` forms — reusing the existing `resolveMainImport`/`isInside` helpers rather than re-implementing resolution. Eight negative fixtures plus a positive "consume ports, export your own contracts" fixture prove the detector discriminates ([layerBoundaries.test.ts:576](test/main/tape/layerBoundaries.test.ts#L576)), and the real-tree scan passes at HEAD. The remaining theoretical gap is const-indirection (`import X …; const y = X; export { y }`), which needs full type info to catch — acceptable scope for a static guard whose job is the realistic re-created-forwarder case.

Also checked the extra pin: View list rows keep only `manifestHash` + `integrity` while the detail projection carries `payloadHash`/`metaHash` ([traceInspector.test.ts:157](test/main/tape/traceInspector.test.ts#L157)) — matches the PR's stated list-vs-detail contract and the route's `.optional()` shape.

Verified locally: `layerBoundaries` 46/46 ✓, `traceInspector` 17 tests (new pin passing; 9 pre-existing sqlite-gated skips) ✓, `typecheck` ✓, oxlint on the touched files ✓. The contract file's 8 tests skip on this machine (no Node-ABI `better-sqlite3-multiple-ciphers`) and stay delegated to the native CI allowlist; their four new readers were verified against the store source directly as above.

One non-blocking observation: the regenerated baselines stamp `headCommit: f3a53f6062`, one commit behind the current tip — the intervening commits are test-only, so the src-scoped content is unaffected; only the stamp lags.

No new findings. Approving.

PR #2233, opened 2026-09-04
perf(tape): index-range reads, shared domain primitives

yyhhyyyyyy:
## Summary

Two Tape read paths did more work than their output needed. `getMessageRecords`, which runs before every provider request, read and folded every message, tool and anchor row of the session and parsed each tool payload up to three times, although its result depends only on message rows and `message/retracted` events. And both effective-input queries were written as `WHERE kind IN (...)`, which SQLite answers by walking the whole session on the primary key; under SQLCipher that decrypts every page the session touches. Alongside, `tape/domain` carried ten copies of the same SHA-256 pattern, four copies of `deepFreeze` and a string comparator, and four UUID regexes that disagreed with each other.

No schema, fact name, provenance key, hash format or IPC change. Existing Tapes read identically.

## What changed

**Effective-input reads walk index ranges instead of the session.** `getEffectiveViewInputRows` and the new `getEffectiveMessageInputRows` merge one range per kind (`UNION ALL ... ORDER BY entry_id`, no sort), so only the pages holding the requested rows are read. Measured on a 10k-entry encrypted session with the 64 MiB cache:

| read | before | after |
|---|---|---|
| effective-view inputs, cold | 215 ms | 115 ms |
| `getMessageRecords` inputs, cold | 132 ms (all effective inputs) | 68 ms (messages only) |
| `getMessageRecords` inputs, warm | 4.3 ms | 1.8 ms |

**`getMessageRecords` reads only what decides it.** It now reads message rows and retraction events alone; `getViewManifestSourceMaps` filters with the same predicate. One pair of constants (`EFFECTIVE_VIEW_INPUT_KINDS`, `EFFECTIVE_MESSAGE_INPUT_KINDS`) drives both the JS predicates and the SQL, typed so `event` cannot be listed and duplicate the retraction range.

**The fold parses each tool row once.** Identity, rank and the stored `orderSeq` are read together; the rank comparison no longer re-parses `meta_json`, and the orderSeq projection re-serialises only when a compaction shift moved the message.

**`tape/domain/primitives.ts`.** `SHA256_HEX_PATTERN`, `deepFreeze`, `utf8Length` and the string comparator now live in one module. The comparator is renamed `compareUtf16`: it orders by UTF-16 code unit (what `<` does), not by code point as the old name said, and persisted canonical orderings depend on that comparison. `hasExactKeys` and `isPlainRecord` stay per-file because the three versions differ (optional keys, prototype check).

**UUID identities validate the same way everywhere.** The Execution Journal accepted any version nibble, the contract and tool-surface validators required 1-8, and only two of the four sites normalised case. One `CANONICAL_UUID_PATTERN` plus `canonicalUuid()`: input validators normalise then check, validators of stored values check the canonical form directly, so nothing stored is loosened. `tapeEntriesToEffectiveMessageRecords` had no callers and is gone.

**Tests.** Real-SQLite tests assert both reads match their JS predicates and the table mock. The mock-contract parity gets a second seed: one assistant message written three times (sent, retracted, sent again at a shifted orderSeq, then pending) with production-shaped tool rows in both the object and the legacy JSON-text `toolCall` encoding. Replacement, retraction, the tool join and the orderSeq rewrite are now compared between the SQL and JS views on non-trivial input, and absolute assertions keep the parity from passing on two empty results. A fold test pins that `messageRecords` are derived from message and retraction rows alone.

## Compatibility

- The one behaviour change: the Journal now rejects a `runId` whose UUID version nibble is outside 1-8. Every runtime `runId` has come from `randomUUID()` (v4) since the Journal was introduced, so no stored fact is affected, and the read path lands in the `try/catch` of `classifyExecutionJournalRows` (`malformed_fact`, parked), so even a hypothetical bad row would not block harness construction. A test pins the rejection and the upper-case to canonical idempotency.
- `getEffectiveMessageInputRows` is a new `TapeEntryStore` port method; the test harness mock and the ad-hoc mocks implement it.
- The two range-merge queries bind a named parameter (`$session`), the first use in `src/main`. SQLite binds it once for all five branches.

## Validation

- `test/main`, run in directory chunks because the one-shot runner is killed in this sandbox: 586 files, 8218 tests passed, 0 failures (31 SQLite-gated files skip under the Node ABI by design; `test/main/scripts` and `performance` cannot start in the sandbox)
- Real SQLite under the Electron ABI on the CI native allowlist: 13 files, 295 passed
- `test/renderer/components/tape-inspector`: 8 files, 114 passed
- Query plans and row-for-row equality of old and new SQL checked on a 10k-entry encrypted fixture with the bundled SQLite 3.53

## Summary by CodeRabbit

* **Improvements**
  * Improved tape processing for effective messages, revisions, retractions, pending records, and associated tools.
  * Improved consistency when resolving revised or retracted messages and reconstructing replay views.
  * Standardized UUID and SHA-256 validation, normalization, and deterministic ordering across tape and task workflows.

* **Bug Fixes**
  * Corrected effective-message selection and view replay behavior.
  * Ensured retracted messages and revised assistant responses are handled consistently.

* **Tests**
  * Expanded coverage for message queries, persistence parity, revisions, retractions, tool associations, and UUID validation.

### DIFF
--- a/src/main/tape/application/common.ts
+++ b/src/main/tape/application/common.ts
@@ -1,6 +1,5 @@
 import { TAPE_INCARNATION_META_KEY, type DeepChatTapeEntryRow } from '../domain/entry'
-
-const UUID_PATTERN = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-8][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/
+import { CANONICAL_UUID_PATTERN } from '../domain/primitives'
 
 export function parseJsonObject(raw: string): Record<string, unknown> {
   try {
@@ -24,7 +23,7 @@ export function readCanonicalTapeIncarnationId(row: DeepChatTapeEntryRow): strin
     return null
   }
   const value = parseJsonObject(row.meta_json)[TAPE_INCARNATION_META_KEY]
-  return typeof value === 'string' && UUID_PATTERN.test(value) ? value : null
+  return typeof value === 'string' && CANONICAL_UUID_PATTERN.test(value) ? value : null
 }
 
 export function parseJsonValue(raw: string): unknown {
--- a/src/main/tape/application/factPersistence.ts
+++ b/src/main/tape/application/factPersistence.ts
@@ -7,7 +7,6 @@ import {
 } from '@/tape/domain/facts'
 import type { DeepChatTapeEntryRow } from '@/tape/domain/entry'
 import type { TapeBootstrapStore, TapeEntryStore } from '@/tape/ports/storage'
-import { buildEffectiveTapeView } from '@/tape/domain/effectiveView'
 import {
   parseAssistantBlocks,
   parseTapeJsonObject,
@@ -611,10 +610,4 @@ export function appendMessageRetractionToTape(
   })
 
   return 1
-}
-
-export function tapeEntriesToEffectiveMessageRecords(
-  rows: DeepChatTapeEntryRow[]
-): ChatMessageRecord[] {
-  return buildEffectiveTapeView(rows, { includePending: true }).messageRecords
 }
--- a/src/main/tape/application/factService.ts
+++ b/src/main/tape/application/factService.ts
@@ -281,7 +281,7 @@ export class TapeFactService
   }
 
   getMessageRecords(sessionId: string): ChatMessageRecord[] {
-    return buildEffectiveTapeView(this.table.getEffectiveViewInputRows(sessionId), {
+    return buildEffectiveTapeView(this.table.getEffectiveMessageInputRows(sessionId), {
       includePending: true
     }).messageRecords
   }
--- a/src/main/tape/application/lineageService.ts
+++ b/src/main/tape/application/lineageService.ts
@@ -11,7 +11,8 @@ import {
 } from '../domain/entry'
 import type { TapeApplicationProviders } from '../ports/application'
 import { parseJsonObject, parseJsonValue } from './common'
-import { computeTapeIdentity, TAPE_IDENTITY_PATTERN } from '../domain/tapeIdentity'
+import { SHA256_HEX_PATTERN } from '../domain/primitives'
+import { computeTapeIdentity } from '../domain/tapeIdentity'
 
 type TapeLineageProviders = Pick<
   TapeApplicationProviders,
@@ -141,7 +142,7 @@ function parseSubagentTapeLink(row: DeepChatTapeEntryRow): ParsedSubagentTapeLin
     (linkVersion === 1 && childTapeIdentity === undefined) ||
     (linkVersion === SUBAGENT_TAPE_LINK_VERSION &&
       typeof childTapeIdentity === 'string' &&
-      TAPE_IDENTITY_PATTERN.test(childTapeIdentity))
+      SHA256_HEX_PATTERN.test(childTapeIdentity))
   if (
     row.kind !== 'event' ||
     row.name !== SUBAGENT_TAPE_LINK_EVENT_NAME ||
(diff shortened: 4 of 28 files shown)


zerob13 (approved):
## Review: perf(tape): index-range reads, shared domain primitives

Reviewed at `3303fee2c0` (8 commits, 28 files, +530/−209) against merge-base `68f39684d` in an isolated read-only worktree. Static review plus independent reproduction of the load-bearing claims.

### Independently verified

- **Query-plan claims reproduce.** Ran `EXPLAIN QUERY PLAN` for old vs. new SQL against the real DDL (`PRIMARY KEY (session_id, entry_id)`, `idx_..._session_kind`, partial `idx_..._event_name`) with the bundled `better-sqlite3-multiple-ciphers` (SQLite 3.53.0):
  - Old: a single `SEARCH ... sqlite_autoindex (session_id=?)` — the documented whole-session PK walk.
  - New: per-kind `SEARCH ... (session_id=? AND kind=?)` ranges plus the retraction arm on the partial `event_name` index, merged via `MERGE (UNION ALL)` with no `USE TEMP B-TREE FOR ORDER BY`.
- **Row-for-row equality.** Old vs. new SQL compared on randomized sessions (3 sessions, 685 rows across all 8 kinds, random event names): identical rows in identical `entry_id` order; the message-only set exactly equals the JS predicate's selection.
- **The narrowing is sound.** Traced the whole fold in `buildEffectiveTapeView`: `messageRecords`/`messageEntries` derive from `messageCandidates` (message rows) + `retractedMessageIds` (retraction events) only — `shouldReplaceMessage` consults the record and `entry_id` alone, and tool candidates/anchors cannot influence them. Feeding `getEffectiveMessageInputRows` therefore produces identical output; the new `tapeFacts` invariant test pins exactly this for both `includePending` values.
- **`isEffectiveMessageInputRow` ≡ the old inline filter** in `viewReplayService`, so the replay read is unchanged.
- **Consolidation is behavior-identical everywhere except the one documented site.** Compared every migrated pattern char-by-char: the `common.ts`/`executionContract.ts`/`toolSurfaceFacts.ts` UUID patterns already were the 1–8 canonical form; the SHA-256 patterns differ only in flags; `compareCodePoints` → `compareUtf16` is a pure rename with an identical body (the old name was the misnomer); `deepFreeze`/`utf8Length` bodies unchanged.
- **No dangling references.** `TAPE_IDENTITY_PATTERN` and `tapeEntriesToEffectiveMessageRecords` have zero remaining references; all four mock stores implement the new port method; `reconcilerService`/`factPersistence` correctly keep `getEffectiveViewInputRows` (they need tool rows).
- **Tool-fold rework preserves semantics.** `shouldReplaceToolRow` with precomputed ranks makes the same decisions (rank, then `entry_id`); each tool payload is parsed once and re-serialized only when a compaction shift moved the message.

### Behavior change (acknowledged, acceptable)

The Journal now rejects `runId`s whose UUID version nibble is outside 1–8. Well-reasoned in the PR body: every runtime `runId` is `randomUUID()` v4, and a hypothetical stored bad row lands in `classifyExecutionJournalRows`'s `try/catch` as `malformed_fact` (parked) rather than blocking harness construction. Pinned by the new test, which also verifies upper-case → canonical idempotency with a genuinely case-differing Run ID.

### Notes (non-blocking)

- The second mock-contract seed also fixes a real pre-existing seed bug: the old `message/retracted` event in `seedConversation` carried no `messageId`, so the retraction it was meant to exercise was a no-op. Good catch making that observable through the window assertions.
- Leaving the `compareCodePoints` copies in `agent/deepchat/runtime` untouched is the right call — importing `tape/domain` primitives across that boundary would be worse than the duplication.
- Test additions are proportionate: one invariant test guarding the core claim, one pin for the behavior change, SQL↔JS↔mock parity on a non-trivial seed, and mechanical mock additions. No padding.

No blocking findings. Approving.

PR #2235, opened 2026-09-04
refactor(tape): make implicit contracts explicit

yyhhyyyyyy:
## Summary

Four places in the Tape stack relied on something that was true but written down nowhere, or written down in several places at once. Every SQLite connection ran with `synchronous=NORMAL` because the bundled `better-sqlite3-multiple-ciphers` build compiles that default in. The six strict-writer namespaces were spelled out three times: as generic-append rejections, as a strict-writer switch, and again as the audit-event exclusion list in `effectiveView.ts`. The Tool Surface domain signalled a contract violation by throwing the built-in `TypeError`, and the application layer classified on that. And `createSessionDataFromDatabase` handed out the concrete `SessionTape` class, so one runtime module typed its dependency as `Pick<SessionTape, ...>` because the read it needed had no port.

This branch states each of those in code, once. No schema, fact name, provenance key, hash format or IPC change. Existing Tapes read identically.

## What changed

**`synchronous = NORMAL` is set explicitly.** `configureSQLiteConnection` now issues the pragma after `journal_mode = WAL`. Measured on both encrypted and plain WAL connections under the current native build, the effective value was already `1`; the pragma keeps that durability contract independent of a future rebuild of the native module without `SQLITE_DEFAULT_WAL_SYNCHRONOUS=1`, which would otherwise fall back to FULL and add an fsync per commit.

**Reserved namespaces are one rule table.** `reservedNamespaces.ts` describes each strict-writer slice once: its exact fact names, the entry kind it is bound to, the reserved sibling prefix (`execution/`, `contract/`), whether its events are audit evidence, and its error texts. `assertTapeAppendAuthorized` walks the rules in declaration order, and `effectiveView.ts` derives `DEFAULT_EXCLUDED_TAPE_EVENT_NAMES` from the rules flagged `auditEvents: true` plus the four bookkeeping events written through the generic path. The check order, the resulting exclusion set (14 names) and the strict-writer error texts are unchanged.

While deriving the list I checked whether `provider/attempt_completed` was missing from it. It is not: `getLastEffectiveTapeMetrics` reads the latest attempt outcome from effective rows for the `tape_info` cache metrics, so the rule declares `auditEvents: false` with that reason next to it. Before, that only followed from the name being absent.

**Tool Surface facts throw their own error.** `commitToolSurfaceView` classified every `TypeError` escaping its transaction as `invalid_input`. The forty contract checks in `toolSurfaceFacts.ts` do throw for bad input, but so does any programming error inside the transaction, such as reading a property of an undefined row, and those were reported as if the caller had sent invalid data. The domain now throws `ToolSurfaceFactError` (the other Tape domains already carry their own error classes) and the service classifies on that type. A regression test pins a store `TypeError` inside the transaction to `persistence_failed`.

**The facade is exposed through its ports.** `SessionTape` now implements a single `SessionTapeCapabilities` alias, and `createSessionDataFromDatabase` returns `tapeStore` under that type. The concrete instance stays inside the composition root, whose `SessionTapePort` wrappers are the only callers of the facade's direct read helpers (`info`, `search`, `getContext`, `anchors`, `handoff`, `exportReplaySlice`). Context occupancy gets a `TapeContextOccupancyReader` port, `TapeContextOccupancyEvidence` moves next to it in `ports/capabilities.ts` (still re-exported from `application/contracts.ts`), and `contextOccupancyCoordinator.ts` joins the layer test's list of consumers that must not import the concrete facade. `handoffResult` is removed from the facade and the fact service; nothing in `src`, `test` or `docs` called it.

**Architecture baselines** record the two moved edges (`contextOccupancyCoordinator -> tape/ports/capabilities`, and `effectiveView` no longer importing `executionJournal`). The rest of that diff (six more `main` files, the shared routes digest) predates this branch: the baselines were last generated at f3a53f606 and dev has merged #2229 and #2233 since.

## Compatibility

- `ToolSurfaceProvenanceError.code` for a non-domain failure inside the Tool Surface transaction is now `persistence_failed` instead of `invalid_input`. No caller branches on that code; the loop runner rethrows the error as-is.
- Domain validation errors from `toolSurfaceFacts.ts` have `name === 'ToolSurfaceFactError'` instead of `'TypeError'`. Nothing in `src/main` or `src/shared` tested for `TypeError` other than the one classification site this branch changes.
- A generic append of a `context` row now fails with one message that names both the kind and `skill/materialized`; previously the two cases had separate texts. One test substring was adjusted.
- `SessionData.tapeStore` is typed as `SessionTapeCapabilities`. Consumers can only reach methods some port declares; the runtime object is the same instance as before.
- `DEFAULT_EXCLUDED_TAPE_EVENT_NAMES` is `readonly string[]` rather than a literal tuple, and the order of names inside the SQL `NOT IN (...)` changed. Both consumers only need the set.

## Validation

- `test/main`, run in directory chunks because the one-shot runner is killed locally by `scripts/buildCli.test.ts` (it executes the CLI bundle with `process.execPath`, which in this sandbox is not a plain Node binary; unrelated to this branch): `tape` + `session` + `data` 54 files, 878 passed; `agent` 110 files, 2109 passed; remaining directories 576 files, 8134 passed; `scripts` 29 of 30 files passed
- Real SQLite under the Electron ABI on the CI native allowlist: 13 files, 296 passed (one more than #2233: the new regression test)
- `architectureBaseline.test.ts` passes before and after the refresh; cycle counts are unchanged (13 / 2 / 0 / 0)

## Summary by CodeRabbit

- **New Features**
  - Added context-occupancy evidence access for session runtime components.
  - Added dedicated validation errors for tool-surface data.
  - Centralized reserved audit-event and namespace authorization rules.

- **Bug Fixes**
  - Database connections now use NORMAL synchronization for consistent durability behavior.
  - Tool-surface persistence errors are classified more accurately.

- **Refactor**
  - Simplified session tape capability exposure and internal boundaries.
  - Removed the obsolete handoff result API.

- **Documentation**
  - Refreshed architecture baseline and dependency reports.

### DIFF
--- a/docs/architecture/baselines/agent-system-layered-runtime-baseline.json
+++ b/docs/architecture/baselines/agent-system-layered-runtime-baseline.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 2,
   "goal": "agent-system-layered-runtime",
-  "headCommit": "f3a53f60629162c692eed56b480a823d973f82fa",
+  "headCommit": "ed7adc34c5fa87a681f324de2f5f1ac1a66b50ef",
   "relevantWorkingTree": {
     "dirty": false,
     "files": []
@@ -495,7 +495,7 @@
       "src/shared/contracts/routes/window.routes.ts",
       "src/shared/contracts/routes/workspace.routes.ts"
     ],
-    "sha256": "f846ec8ca18abe70f548b5739007c2fc232c527de20606d039bc007489227d62"
+    "sha256": "fe38e410c549358f61969450c43acf85a057bb560b6bfa85b8b21c1116ee01d4"
   },
   "storage": {
     "sqlite": {
--- a/docs/architecture/baselines/dependency-report.md
+++ b/docs/architecture/baselines/dependency-report.md
@@ -4,8 +4,8 @@ Generated on 2026-09-04.
 
 ## main
 
-- Total files: 853
-- Internal dependency edges: 3009
+- Total files: 859
+- Internal dependency edges: 3021
 - Cycles detected: 13
 
 ### Top outgoing dependencies
@@ -33,14 +33,14 @@ Generated on 2026-09-04.
 - `data/baseTable.ts`: 44
 - `routes/routeRegistry.ts`: 43
 - `remote/types.ts`: 39
+- `tape/ports/capabilities.ts`: 36
 - `agent/settings.ts`: 35
 - `config/settingsStore.ts`: 35
-- `tape/ports/capabilities.ts`: 35
 - `tape/domain/entry.ts`: 34
 - `tape/domain/canonicalJson.ts`: 31
 - `agent/deepchat/runtime/types.ts`: 28
-- `tape/domain/executionJournal.ts`: 26
 - `agent/deepchat/runtime/toolSurface.ts`: 25
+- `tape/domain/executionJournal.ts`: 25
 - `memory/types.ts`: 24
 - `agent/deepchat/instance/deepChatAgentRuntime.ts`: 23
 
--- a/src/main/agent/deepchat/runtime/contextOccupancyCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/contextOccupancyCoordinator.ts
@@ -1,7 +1,7 @@
 import type { SessionContextOccupancySnapshot } from '@shared/types/agent-interface'
 import { toAppSessionId } from '@/agent/shared/agentSessionIds'
 import type { DeepChatAgentRuntime } from '@/agent/deepchat/instance/deepChatAgentRuntime'
-import type { SessionTape } from '@/tape/application/sessionTape'
+import type { TapeContextOccupancyReader } from '@/tape/ports/capabilities'
 import type { SessionSettingsCoordinator } from './sessionSettingsCoordinator'
 import type { SessionStateResolver } from './sessionStateResolver'
 import { resolveEffectiveContextBudget } from './contextBudget'
@@ -10,7 +10,7 @@ type ContextOccupancyDependencies = {
   runtime: Pick<DeepChatAgentRuntime, 'getOrHydrateScope'>
   sessionState: Pick<SessionStateResolver, 'getSummary'>
   sessionSettings: Pick<SessionSettingsCoordinator, 'getEffectiveGenerationSettings'>
-  tape: Pick<SessionTape, 'getContextOccupancyEvidence'>
+  tape: TapeContextOccupancyReader
 }
 
 export function unavailableContextOccupancy(): SessionContextOccupancySnapshot {
--- a/src/main/data/connectionConfig.ts
+++ b/src/main/data/connectionConfig.ts
@@ -18,6 +18,12 @@ export function configureSQLiteConnection(db: Database.Database, password?: stri
   }
 
   db.pragma('journal_mode = WAL')
+  // NORMAL is what every connection has been running with: the bundled
+  // better-sqlite3-multiple-ciphers build compiles in SQLITE_DEFAULT_WAL_SYNCHRONOUS=1, which
+  // applies only while the application never sets the pragma itself. Stating it here keeps the
+  // durability contract (commits survive a process crash, a power loss may drop the newest
+  // committed transactions but never corrupts the file) independent of that build flag.
+  db.pragma('synchronous = NORMAL')
   // 64 MiB page cache ceiling (negative = KiB) for every connection opened through this helper.
   // It is sized for the main database: SQLCipher decrypts every page it loads, so the whole-session
   // Tape reads before each provider request must stay resident to be cheap. Other connections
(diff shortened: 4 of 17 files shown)


zerob13 (approved):
## Review at `a056d3cf9` (6 commits on `7ae720dce`, #2233 merge point)

All four claims in the PR body were re-verified against the code independently, not taken from the description.

**Namespace rule table is behavior-preserving.** I walked `assertTapeAppendAuthorized` ([reservedNamespaces.ts:104](src/main/tape/domain/reservedNamespaces.ts#L104)) old vs new for all six slices, both the strict-writer and the generic path:

- `isToolSurfaceTapeReservedName` is exact membership in the 3-name `TOOL_SURFACE_TAPE_EVENT_NAMES` ([toolSurfaceFacts.ts:37](src/main/tape/domain/toolSurfaceFacts.ts#L37)), so replacing it with `rule.names.includes` narrows nothing. The `execution/` / `contract/` prefix guards survive as `reservesName`, so unknown sibling names stay reserved for generic appends while strict writers stay bound to their exact declared names — exactly the old split.
- Kind binding: only `skill-materialized` declares `kind: 'context'`; `ownsFact` ([reservedNamespaces.ts:121](src/main/tape/domain/reservedNamespaces.ts#L121)) reproduces the old `kind && name` conjunction, and `claimsGenericAppend` claims any `context` row plus the bare `skill/materialized` name, matching the old two checks. All other slices stay name-only, as before.
- `DEFAULT_EXCLUDED_TAPE_EVENT_NAMES` ([effectiveView.ts:50](src/main/tape/domain/effectiveView.ts#L50)): 4 bookkeeping names + `RESERVED_AUDIT_TAPE_EVENT_NAMES` (4 execution + 2 contract + 3 tool-surface + 1 compaction = 10) is set-identical to the old 14-name literal. The only consumer is the `Set` in the same file, so the `readonly string[]` widening and the `NOT IN` reordering are inert. The `auditEvents: false` reasons (provider-attempt feeds `tape_info` metrics; `context` rows excluded by kind) match what the old absence-of-name encoded.

**Error reclassification blast radius is zero.** `instanceof ToolSurfaceProvenanceError` guards rethrow, `instanceof ToolSurfaceFactError` → `invalid_input`, everything else → `persistence_failed` ([toolSurfaceProvenanceService.ts:1056](src/main/tape/application/toolSurfaceProvenanceService.ts#L1056)). Grep confirms: no `instanceof TypeError` remains anywhere in `src/`, `ToolSurfaceProvenanceError` is not referenced outside its own module, and the only `persistence_failed` branch assertion in tests belongs to `ExecutionJournalError`. The new regression test pinning a store `TypeError` → `persistence_failed` is the right test for this fix. `ToolSurfaceFactError` follows the established per-domain error class pattern (`TaskContractError`, `ExecutionJournalError`, `TaskEvaluationError`, …).

**Port exposure is clean.** `SessionTapeCapabilities` ([sessionTape.ts:149](src/main/tape/application/sessionTape.ts#L149)) replaces the 30-interface `implements` list with the same intersection as a named alias; the composition root keeps the concrete instance and hands out the narrowed type ([session/data/index.ts:40](src/main/session/data/index.ts#L40)). After this, `session/data/index.ts` is the only importer of the concrete facade module in `src/`; `contextOccupancyCoordinator.ts` now depends on `TapeContextOccupancyReader` ([capabilities.ts:353](src/main/tape/ports/capabilities.ts#L353)) and is added to `CAPABILITY_SCOPED_CONSUMER_FILES` enforced by `findConcreteTapeFacadeImportViolations` — the boundary is now guarded, which the old `Pick<SessionTape, …>` typing actively violated. `handoffResult` has no remaining references in `src`, `test` or `docs`; dead code confirmed.

**Baselines.** The two recorded edges match the code (coordinator → `tape/ports/capabilities`; `effectiveView` no longer importing `executionJournal`, visible as its outgoing-dependency count dropping 26 → 25). The rest of the regen is dev drift since `f3a53f606`, as stated.

Verified locally: `test/main/{tape,session,data}` 54 files / 878 passed (matches the PR's numbers exactly), `architectureBaseline.test.ts` 2/2, `contextOccupancyCoordinator.test.ts` 16/16, `typecheck:node`, `oxfmt --check` and `oxlint` on all touched files. SQLite-gated suites stay delegated to the native CI allowlist, as usual.

Findings — 0 P1 / 0 P2 / 2 P3:

1. **P3** [skillMaterialization.test.ts:848](test/main/tape/skillMaterialization.test.ts#L848) — the adjusted assertion loosened to `.toThrow('context entry kind')`. Pinning the full merged message ('The context entry kind and skill/materialized are reserved for the strict materialization writer.') would also lock the `skill/materialized` mention into the contract instead of just the kind half.
2. **P3** [reservedNamespaces.ts:89](src/main/tape/domain/reservedNamespaces.ts#L89) — `RULES_IN_ORDER` depends on object-literal key insertion order. It is documented ("Declaration order is the order generic appends are checked in") and order only selects which message fires on a multi-claim append, and the `Record<TapeReservedNamespace, …>` typing buys exhaustiveness (adding a namespace forces a rule), so this is an acceptable trade — just noting the coupling.

No over-engineering: every hunk converts an implicit invariant into a declared one or deletes a third spelling of the same set; net LOC is down. Compatibility notes in the PR body are accurate and complete. Approving.

PR #2236, opened 2026-09-05
fix(chat): restore queue slots and allow 10

zerob13:
Allow up to 10 waiting messages per session and free a waiting slot as soon as a queued message is claimed. Previously, the UI hid the running message while the main process still counted it toward the five-message limit, so Enter and Queue could fail even with a visible count of 4/5.

One shared constant controls main-process admission, renderer submit gating, and the waiting-lane counter/full hint. Recovery preserves every accepted input and its queue order: a released or restarted claim can temporarily produce 11/10 waiting messages. The lane shows the actual count and both admission gates block new inputs until fewer than 10 remain.

```text
BEFORE
+-------------------------------+
| Waiting 4/5                   |
| Four waiting messages         |
+-------------------------------+
| Draft            [Queue]      | --> Limit reached (one running)
+-------------------------------+

AFTER
+-------------------------------+
| Waiting 9/10                  |
| Nine waiting messages         |
+-------------------------------+
| Draft            [Queue]      | --> Waiting 10/10
+-------------------------------+

RECOVERY
[Waiting 11/10] --> [Waiting 10/10] --> [Waiting 9/10]
[Queue disabled]   [Queue disabled]   [Queue enabled]
```

The native pending-input suite is included in the CI job that rebuilds SQLite for Node and sets `DEEPCHAT_REQUIRE_NATIVE_SQLITE=1`. Its migration and transaction fixtures use real temporary-directory operations and the current session table owners.

Validation:

- `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, and `pnpm run typecheck` pass.
- All 21 native pending-input tests pass under Electron with `DEEPCHAT_REQUIRE_NATIVE_SQLITE=1`, including full-queue refill, four release/restart recovery paths, migrations, and transaction rollback.
- 81 main-process tests pass across pending input storage, ownership, admission, and pumping.
- 151 renderer tests pass across composer submission, pending input state, ChatPage, and PendingInputLane, including the actual overflow count and capacity recovery.

## Summary by CodeRabbit

- **Bug Fixes**
  - Claimed inputs no longer count toward the active pending-input capacity.
  - Recovered and claimed items remain visible in the queue, even when the displayed count temporarily exceeds the limit.
  - Queue counts and capacity messages now reflect the actual number of pending inputs.
  - Capacity becomes available only after enough pending items are deleted, preventing premature additions while recovery overflow remains visible.


zerob13 (commented):
## Review: Approve ✅

*(Comment review — GitHub does not allow approving one's own PR. Reviewed as if approving.)*


Verified against `f886d6e9b` (merge-base with `dev`):

**Core fix — correct.** `countActiveQueue` now excludes `claimed` rows, which aligns the JS admission count with the pre-existing SQL semantics in `countActiveBySession` (`AND NOT (mode = 'queue' AND state = 'claimed')`, [deepchatPendingInputs.ts:208](src/main/session/data/tables/deepchatPendingInputs.ts#L208)) and with the renderer's `activeCount` (derived from `listPendingInputs`, which already filtered `claimed`). This resolves the UI-shows-4/5-but-Enter-fails mismatch described in the PR.

**Claimed inputs remain tracked.** Still returned by `getInput`/`listActiveInputs` (restart recovery), `hasClaimedInput` (drain gating), and `getClaimedInput`; release paths (`releaseClaimedQueueInputTo`, `blockClaimedInput`) correctly re-occupy a slot.

**Races — none.** `ensureWithinLimit` + insert run synchronously on the single-threaded main process; the async gap in `sendQueuedMessage` is re-checked authoritatively at the final synchronous admission point. Delete/move are transactional.

**No stale assumptions of 5.** Constants, error strings, i18n `queueCount`/`limitReached` (all 20 locales use `{count}/{max}`), and existing tests checked — no leftover `5` semantics. Off-by-one correct on all three gates (main throw, renderer `isAtCapacity`, lane hint).

**Test — appropriate.** The single new `it` covers the persistence contract (reject at 11, refill after claim with ordering, refill after delete while claimed stays active) via the public `SessionPendingInputs` API. Smallest durable coverage, nothing implementation-coupled.

### Non-blocking notes (follow-up only)

1. **P3** — The limit now lives in three hand-synchronized places: `MAX_ACTIVE_PENDING_INPUTS` ([pendingInputs.ts:12](src/main/session/data/pendingInputs.ts#L12)), `MAX_PENDING_INPUTS` ([pendingInput.ts:6](src/renderer/src/stores/ui/pendingInput.ts#L6)), and the prop default in [PendingInputLane.vue:295](src/renderer/src/components/chat/PendingInputLane.vue#L295). All three were updated consistently here, but a future limit change must again touch all three or the UI counter will disagree with the admission gate. Consider a shared constant in `src/shared` as a follow-up.
2. **P3** — `countActiveQueue` materializes rows and filters in JS while its sibling `countActive` delegates to a SQL `COUNT`. A SQL predicate would match the sibling's pattern and avoid row materialization on every enqueue. Not required in this PR given the smallest-change rule.

yyhhyyyyyy (commented):
Three inline notes. The first is a small regression, the other two are minor.

src/main/session/data/pendingInputStore.ts
```diff
@@ -61,7 +61,7 @@ export class SessionPendingInputStore {
   countActiveQueue(sessionId: string): number {
     return this.database.deepchatPendingInputsTable
       .listActiveBySession(sessionId)
-      .filter((row) => row.mode === 'queue').length
+      .filter((row) => row.mode === 'queue' && row.state !== 'claimed').length
```
yyhhyyyyyy, line 64: Excluding `claimed` here fixes the 4/5 mismatch, but the release paths (`releaseClaimedQueueInputTo`, `releaseClaimedInput`, restart recovery) move the item back to `pending` / `retry_required` with no capacity check. With 10 waiting + 1 claimed, a pre-user-fact rollback, `release-for-mutation`, or a restart release leaves 11 waiting items and the lane shows `11/10`. Reproduced locally; before this change the same flow saturated at 5. Probably want admission to keep one slot reserved while an item is claimed, or cap the displayed count.

src/main/session/data/pendingInputs.ts
```diff
@@ -9,7 +9,7 @@ import type {
 import { SessionPendingInputStore } from './pendingInputStore'
 import type { SessionTranscript } from './transcript'
 
-const MAX_ACTIVE_PENDING_INPUTS = 5
+const MAX_ACTIVE_PENDING_INPUTS = 10
```
yyhhyyyyyy: Nit: the limit now lives in three places (this constant, renderer `MAX_PENDING_INPUTS`, and the `activeLimit` default in `PendingInputLane.vue`, which `ChatPage` never passes). A single shared constant would stop them drifting again.

test/main/session/data/tables/deepchatPendingInputsTable.test.ts
```diff
@@ -32,6 +32,47 @@ describe('DeepChatPendingInputsTable migrations', () => {
   })
 })
 
+describeIfNativeSqlite('SessionPendingInputs queue capacity', () => {
```
yyhhyyyyyy, line 36: FYI this suite is `describe.skip` under plain Node (17 skipped locally), and the `test-main` CI job runs plain Node, so this regression test won't execute in CI unless the file is added to the `DEEPCHAT_REQUIRE_NATIVE_SQLITE=1` list.

PR #2237, opened 2026-09-05
fix: restore prompt scrolling and notifications

zerob13:
Custom prompt forms can extend past the settings window without a scrollbar, leaving the confirm button outside the viewport. Session completion also has no active system-notification delivery path despite the setting remaining available.

- Preserve `undefined` for the optional `DcSheetPanel.scrollBody` boolean so panel appearance enables scrolling by default. Allow the scroll body to shrink while keeping header and footer visible. Explicit scroll opt-outs and plain appearance retain their behavior.
- Deliver successful DeepChat and ACP turn completions through the existing main-process notification service, independently of configured command hooks. Use actual window focus to suppress foreground notifications; a visible, unfocused window qualifies even when it is not minimized, including when another app is fullscreen.
- Reuse existing localized completion text and the existing notification click route. Exclude canceled, failed, paused, subagent, and CLI turns; honor the notification setting.
- Preserve normal-build provider and ACP registry refreshes in a separate commit, as required by the repository guidelines.

```text
BEFORE                        AFTER
+---------------------+       +---------------------+
| Add Prompt        X |       | Add Prompt        X |
| Long form           |       | Form              | |
| No scroll container |       | Scrollable        v |
+---------------------+       +---------------------+
  Confirm below viewport      | Cancel      Confirm |
                              +---------------------+
```

Validation:

- `pnpm run build` (includes Node and renderer typechecks)
- `pnpm run format:check`, `pnpm run lint`, `pnpm run i18n`
- 31 main-process tests: notification delivery, runtime terminal projection, ACP compatibility adapters
- 2 renderer tests: prompt attachment submission and sheet close-button drag behavior
- Electron smoke test at 1000x640 and 1100x900: footer visibility, wheel scrolling to the file section, save and reload persistence, Escape dismissal and focus restoration; screenshots inspected

Native notification banners over a separate fullscreen application have not been manually verified. Automated notification tests cover the corresponding unfocused, visible, non-minimized window state and verify delivery to Electron's notification API.

## Summary by CodeRabbit

* **New Features**
  * Added localized desktop notifications when an agent session finishes while the app is unfocused.
  * Completion notifications include the session title and support notification click handling.

* **Bug Fixes**
  * Improved prompt editor scrolling in smaller windows while keeping action buttons accessible.
  * Prevented completion notifications for paused, aborted, failed, or tool-error sessions.

* **Chores**
  * Updated packaged agent release versions and distribution metadata.


zerob13 (commented):
## Review: Approve ✅

*(Comment review — GitHub does not allow approving one's own PR. Reviewed as if approving.)*

Verified against `f886d6e9b` (merge-base with `dev`). All source diffs reviewed; the Vue boolean-casting mechanism was checked against the actual installed Vue 3.5.40 runtime (`resolvePropValue`) and compiler-sfc output.

**scrollBody fix — mechanism verified correct.** At merge-base `withDefaults` omitted `scrollBody`, so Vue's boolean casting (`isAbsent && !hasDefault → false`) defeated `shouldScrollBody = props.scrollBody ?? !isPlain.value` — panel sheets lost their ScrollArea entirely (regression traced to #2084, which replaced the pre-existing `ScrollArea class="flex-1 h-0"`). The explicit `scrollBody: undefined` default restores the `?? !isPlain` fallback (compiler-sfc emits `default: undefined` verbatim; runtime `hasOwn(opt, 'default')` is true, so the un-cast `undefined` survives). Consumer audit: `PromptEditorSheet.vue` and `SystemPromptEditorSheet.vue` (panel default — both are the intended fix targets; the latter's `h-64` textarea is unaffected), `McpServers.vue` (plain — `false` either way), and the three explicit `:scroll-body="false"` panels. No consumer accidentally gains scrolling. `min-h-0 flex-1` is the standard flex-column fix, equivalent to the old `h-0` trick; header/footer are already `shrink-0`.

**Notification path — end-to-end checks pass.** Single terminal projection: DeepChat runs via `RunLifecycleCoordinator` → `RuntimeHookSink.observeTerminal` (paused excluded by `status !== 'completed'`; aborted/error excluded by `!userStop && !facts.error`); ACP runs via the compatibility adapter (`completed: input.status === 'completed'`). Subagent sessions excluded via `sessionKind !== 'regular'`; CLI runs via `resolveSessionRunId` truthiness. Settings honored; click reuses the existing `systemNotificationClicked` route (pre-existing, unmodified). Delivery is independent of command hooks (directly tested with no hook subscriptions). No double-delivery with hooks; no listener leaks (callback, not subscription). Focus semantics: notify unless some window is visible + un-minimized + focused — visible-but-unfocused (incl. another app fullscreen) notifies, matching the stated design; aggregation matches the existing multi-window pattern.

**i18n — safe today.** All 20 `SUPPORTED_LOCALES` ship `chat.notify.generationComplete` and `getLanguage()` always resolves to a supported locale, so the `import.meta.glob` lookup cannot miss at runtime. Reusing the translated key is the right call over duplicating 20 locales in `@shared`.

**Tests — reasonable, not excessive.** `runtimeHookSink.test.ts` (+35): two contract-focused tests (which settlements fire the callback + error isolation). `notification.test.ts` (+55): focused-matrix, settings, click-route id, zh-CN title — user-visible behavior. `34-prompt-editor-scroll.smoke.spec.ts` (+52): genuine regression spec (small window, sticky footer visible, real scrollTop polling, deep section reachable); selectors verified against actual component/i18n. Serial numbering and naming match the suite.

**Registry refresh — plausible.** `registry.json` is pure version/URL bumps (antigravity 1.1.1, claude-acp 0.75.0, codebuddy 2.146.0, codex-acp 1.10.0, cursor 2026.09.02), no structural edits, isolated in its own commit per guidelines. `providers.json` diff header is consistent with a generated catalog refresh.

### Non-blocking notes (P3, follow-up only)

1. **Spurious completion toast from pending-input pre-stream handoff** — [turnCoordinator.ts:1358](src/main/agent/deepchat/runtime/turnCoordinator.ts#L1358) settles a handoff as `status: 'completed'` with `stopReason: PENDING_INPUT_ABORT_REASON`, which `observeTerminal` maps to a completion notification for a turn that never generated. Only reachable if the app is unfocused during handoff stabilization; inherits the same semantics the Stop/SessionEnd hooks already use. Suggestion: exclude `PENDING_INPUT_ABORT_REASON` from `completed`, or document the parity.
2. **Notification id reuse without closing** — [notification.ts:32](src/main/desktop/notification.ts#L32) reuses `chat/${session.id}/completed` per settled turn, and `clearNotification` removes the wrapper from the Map without calling `.close()`, so stale banners linger in Notification Center. The pre-refactor code used per-event ids. Suggestion: include the turn id in the notification id, or make `clearNotification` call `notification.close()`.
3. **i18n lookup has no fallback** — [notification.ts:29](src/main/desktop/notification.ts#L29): `notificationMessages[...]` dereferences without a guard; a missing locale key would TypeError and the `.catch` would silently swallow the notification. Safe today, but `messages?.generationComplete` with an `en-US` fallback (or moving the dependency into `@shared` with a typed contract) would be sturdier.
4. **`resolveSessionRunId` tri-state semantics undocumented** — [composition.ts:909](src/main/app/composition.ts#L909) returns `null` (no CLI ancestor), a truthy id (CLI run), or `undefined` (unknown ancestor / cycle guard); the caller's `!== null` treats `undefined` as "skip notification" (conservative). Defensible, but a one-line comment (or collapsing to `string | null`) would prevent future misreads.

PR #2238, opened 2026-09-05
fix(chat): preserve drafts across agent switches

zerob13:
Switching away from a new conversation now preserves a separate draft for each agent, including text, attachments, selected skills and inline references. Drafts restore when returning or reloading. Successful sends consume their submitted snapshot, while failed sends and newer edits remain intact.

Existing-session drafts are saved before the input editor unmounts, preserving skills and document nodes. Pending attachment checks are invalidated when their originating view changes or unmounts.

```text
BEFORE
Agent A [draft A] -> all agents -> Agent A [empty]
Agent A [draft A] -> Agent B [draft A]

AFTER
Agent A [draft A] <-> Agent B [draft B]
Leave / reload -> restore the corresponding complete draft
```

Draft restoration copies documents and skill arrays only when the editor needs restoration. The page and composable share the editor handle type, and window listeners use the same environment guard as existing-session drafts.

Known follow-up: deleting agents or sessions still leaves orphaned draft keys. Lifecycle cleanup and regression acceptance criteria are tracked in [Composer Draft Cleanup on Deletion](https://github.com/ThinkInAIXYZ/deepchat/blob/codex/preserve-agent-composer-drafts/docs/issues/composer-draft-deletion-cleanup/spec.md).

Validation:

- `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, and `pnpm run typecheck` passed.
- `pnpm exec electron-vite build` passed.
- `NODE_OPTIONS=--max-old-space-size=6144 pnpm run test:renderer --silent` passed: 267 files and 2373 tests, including deep-link and onboarding coverage using the real draft store with isolated state. Regressions cover agent isolation, full document restoration, debounce/window-close flushing, late send acceptance and attachment checks.
- `pnpm exec playwright test -c test/e2e/playwright.config.ts test/e2e/specs/35-agent-composer-drafts.smoke.spec.ts` passed with isolated Electron user data and no provider requests.

## Summary by CodeRabbit

- **New Features**
  - New-thread composer drafts are now saved separately for each agent.
  - Drafts preserve message text, attachments, skills, and editor content when switching agents, deselecting an agent, remounting, or reloading the page.
  - Drafts are saved automatically, including during navigation and other lifecycle transitions.
  - Switching agents during submission now prevents stale draft content from being submitted.

- **Documentation**
  - Added specifications for cleaning up drafts when agents or sessions are deleted.


zerob13 (commented):
## Review: ✅ APPROVE — 0 P1 / 0 P2 / 5 P3 follow-ups

Reviewed at `bc4b8c2c0` against merge-base `e9f909519` (9 files, +544/-50). Full diff and changed files read in place; dependencies (`composerDraftState`, `composerDraftPersistence`, `ChatInputBox` handle API, `useComposerSubmit`, `useSkillsData`) cross-checked at merge-base. Verified in an isolated worktree: `typecheck:web`, `oxfmt --check`, `oxlint` all clean; 50/50 `NewThreadPage` + 95/95 `ChatPage` renderer tests pass; the new e2e smoke spec passes against a real build (7.1s, isolated user data).

### Key behaviors verified

- **PR claim "successful sends consume their submitted snapshot; failed sends and newer edits remain intact" — TRUE.** `applyAcceptedComposerSubmission` (`composerDraftState.ts:94`) treats content as unchanged only when `revision` AND `rawMessage` AND document all match; any edit during send (revision bump) preserves current text while still subtracting the sent files and their document nodes. Failed/cancelled paths never call `acceptSubmission`.
- **Watcher choreography on agent switch is correct.** The composable's sync persist fires before the page's sync cancel watcher, capturing the old agent's live editor state before cancellation side effects. The `restoring` flag suppresses skills echoes (`useSkillsData` clears on agentId change; ACP sessionId re-emit), the post-flush watcher re-seeds, and later echoes are fingerprint no-ops. No clobber path found.
- **No restore/emit feedback loop.** `restoreDocumentSnapshot` runs inside `syncEditorContent` with sync meta, so restores never re-emit `update:modelValue`/`draft-change`; `updateDraft`'s fingerprint check short-circuits any residual echo.
- **Mid-submission agent switch fully guarded.** Ambient values (`agentId`, `projectDir`, `targetAcpSessionId`) captured at entry before the first `await`; `submission.cancelled` re-checked after every `await`; the `finally` identity check prevents a resolving old submission from clobbering the switched-to state. Cross-agent sends consume the origin agent's draft only.
- **The `ChatPage.vue` teardown move is a genuine fix, not churn.** `disposeComposerSubmit` → `flushComposerDraftPersist` needs a live `chatInputRef` to capture the rich document; in `onUnmounted` the child would already be unmounted and the persisted draft silently degrades to text-only. The extended ChatPage test pins document + activeSkills surviving unmount.
- **Attachment pending-check invalidation works.** `attachmentFilterToken` bumped on agent switch and unmount; late capability results discarded; raw files re-filtered at submit.

### Over-engineering / test assessment

- The 170-line composable mirrors the established `useComposerSubmit` pattern at ~1/8 the size; store placement is required (`acceptSubmission` must mutate shared state after the page unmounts). No premature generality; the store changes are purely additive.
- Test delta is proportionate, not bloat: +226 in `NewThreadPage.test.ts` is six tests, each mapping to a durable category (per-agent restore, unmount-flush persistence, debounce/beforeunload contract, send-resolves-after-remount concurrency, cross-agent consumption, late capability invalidation). The 40-line e2e smoke is the only test proving persistence across a real reload and the deselect→unmount path.

### P3 follow-ups (non-blocking)

1. **E2e spec numbering collision** — `34-agent-composer-drafts.smoke.spec.ts` reuses prefix `34-`, already taken by `34-prompt-editor-scroll.smoke.spec.ts`. Rename to `35-`.
2. **`ComposerHandle` type duplication** — `useNewThreadComposerDraft.ts:14-21` and `NewThreadPage.vue:311-321` re-declare the same handle shape. Export the type from the composable and reuse it.
3. **Missing `typeof window` guard** — `useNewThreadComposerDraft.ts:152-153` registers `pagehide`/`beforeunload` unguarded, while sibling `useComposerSubmit.ts:569` guards with `typeof window !== 'undefined'`. Harmless in the renderer; inconsistent with the pattern being copied.
4. **Orphaned draft keys on agent deletion** — deleting an agent leaves `deepchat.composerDraft.v1.new-thread:<id>` in localStorage forever (`clearComposerDraftFromStorage` has zero callers in `src/`). Pre-existing behavior class shared with session drafts — follow-up ticket for both.
5. **Per-keystroke cost note** — the post-flush restore watcher (`useNewThreadComposerDraft.ts:115-135`) does a full `copyComposerDraft` deep copy plus `JSON.stringify` comparisons on every draft change. Fine in practice (documents carry only file metadata); ChatPage's version caches fingerprints to avoid redundant copies — worth aligning eventually.

Merge-ready as-is; the P3s can be follow-ups.

zerob13 (commented):
## Review: fix(chat): preserve drafts across agent switches (follow-up `f2f0625`)

**Verdict: approve.** The follow-up refactor does exactly what it claims — same watcher choreography, cheaper comparisons — and all 5 prior P3s are addressed. No new blocking findings.

### Refactor assessment

The delta touches the post-flush restore watcher's *internals* only; structure, flush modes, and registration order are untouched:

- **Skills comparison** (`useNewThreadComposerDraft.ts:122-127`): `JSON.stringify` equality replaced by length + element-wise strict equality — algebraically equivalent for `string[]` (order-sensitive both ways; an `undefined` return is caught by the `!skills` guard and lands in the same mismatch branch as before). Length is checked first, so the `some()` index access is safe.
- **Snapshot copy removal** (`:121`): the watcher now reads `draft.value` directly instead of `copyComposerDraft`-ing it per run. Every field that still escapes to the composer handle remains isolated: `[...snapshot.activeSkills]` at `:131` and `copyComposerDocument(snapshot.document)` at `:136` — per-field equivalents of what `copyComposerDraft` (`composerDraftState.ts:62-70`) provided for those two fields. The store itself also copies on write (`stores/ui/draft.ts:78`).
- **`ComposerHandle` single-sourced** (`:20` export; `NewThreadPage.vue:231-234` import) — the `JSONContent` import is gone with no residual references.
- The sync draft watcher (`:87-102`), sync agent-switch watcher (`:104-114` — persist-before-swap preserved), and post-flush re-seed watcher all keep their flush modes and ordering, so the skills-echo / ACP sessionId re-emit / fingerprint no-op choreography verified in the last round is intact by construction.
- ChatPage teardown (`ChatPage.vue:1392-1394` → `dispose()` flushing with a live `chatInputRef`, `useComposerSubmit.ts:1286-1287`) and the `attachmentFilterToken` lifecycle (`NewThreadPage.vue:1370/1373/1541/1596`) are untouched by this commit.

No new tests is acceptable here: the change is a provably-equivalent replacement inside code already covered end-to-end by the 42-test NewThreadPage mount suite (e.g. "restores separate agent drafts with attachments, skills and inline references") plus the unchanged e2e smoke spec.

### Prior P3 follow-ups status

- **P3-1 (e2e prefix collision): addressed.** Renamed to `35-agent-composer-drafts.smoke.spec.ts` with zero content change; no dangling `34-` references anywhere in the repo (including workflows), `35-` is now unique, and `playwright.config.ts` globs `specs/` so the rename is inert.
- **P3-2 (`ComposerHandle` duplication): addressed.** Exported from the composable and re-imported by `NewThreadPage.vue`.
- **P3-3 (`typeof window` guard): addressed.** Added around both add (`:155-158`) and remove (`:162-165`), matching the sibling pattern in `useComposerSubmit.ts:569-572`.
- **P3-4 (orphaned draft keys): addressed** via `docs/issues/composer-draft-deletion-cleanup/spec.md` — debt documentation in the established location (27th entry under `docs/issues/`). Content checks out against the code: `clearComposerDraftFromStorage` indeed has zero production callers; the key format and ownership claims all match.
- **P3-5 (per-keystroke copy + stringify): addressed.** The full `copyComposerDraft` and the skills `JSON.stringify` comparison are gone from the watcher; remaining per-change work (`getDocumentSnapshot` + `composerDocumentsMatch`) is unchanged from before.

Two non-blocking nits, no action required: `const snapshot = draft.value` (`:121`) is no longer a snapshot so the name slightly overstates; and `ChatPage.vue:1076-1087` still carries its own pre-existing inline handle type that could eventually adopt `ComposerHandle & { ... }` too.

### Verification

- `pnpm vitest run --config vitest.config.renderer.ts` — NewThreadPage suites 50/50 (42 + 7 + 1) and ChatPage 95/95; draft model suites (useComposerSubmit / composerDraftState / composerDraftPersistence) 47/47.
- `pnpm run typecheck:web` — clean. `oxfmt --check` on all touched files — clean. `oxlint` on touched sources — 0/0.
- Whole-PR test diff vs `e9f909519`: zero deleted test cases.

Clean follow-up — good to merge from my side.

PR #2239, opened 2026-09-06
feat(plugins): support Codex user packages

zerob13:
Users can install Codex-format plugin packages directly from public HTTPS Git repositories or ZIP files. Packages reuse DeepChat's existing Skills and MCP services, with reviewed context hooks for `SessionStart`, `UserPromptSubmit`, and `SubagentStart`.

## Changes

- Inspect packages without executing plugin code, including JavaScript Skill front matter. Install private snapshots with selected capabilities and manage enablement from the detail page.
- Persist hook invocations and attributed context in Tape. Reuse results for matching prompts, handle edits on the same message, invalidate history when messages are cleared, and discard obsolete output before provider dispatch.
- Start each hook boundary's execution budget after queue admission, record cancellation, decode complete UTF-8 output, and bound the session cache. Revalidate queued retries against session incarnation, owner, invocation status/input and current turn/update activity. Preserve committed compaction projection when hook dispatch fails and exclude ACP requests from plugin context checks.
- Import stdio, HTTP and explicit SSE configuration through existing MCP owners. Require HTTPS for remote endpoints and validate headers after substitution. User plugins resolve only explicitly supplied variables; credentials are encrypted separately from configuration, templates remain intact, and the detail form supports rotation.
- Support enable, disable, reviewed updates, rollback and uninstall with stable installation identities and Skill assignments. Failed recovery disables the plugin and releases the update lock. Failed-revision cleanup cannot skip restoration of the previous MCP configuration and encrypted credentials. Preserve official plugin activation when a user Skill has the same name.
- Localize consent across all 20 locales, display environment/header declarations directly, and use the existing checkbox and uninstall-confirmation components.
- Add authoring instructions, static package validation, a documented directory-snapshot developer workflow, and durable security/runtime regression coverage.

## UI

```text
BEFORE
Plugins
  Built-in tools
  Official plugins                     [Enable]

AFTER
Plugins               [Install from Git] [Install from ZIP]
  Built-in tools
  Official plugins                     [Enable]
  User packages                        [Details]

Review package
  Source, commands, endpoints, compatibility findings
  Environment and header templates     visible declarations
  [x] Skills  [ ] Context hooks  [ ] MCP connections
  [Cancel]                             [Install]

Plugin details
  [Enable / Disable] [Review update] [Reload] [Uninstall]
  Source, selected components, hook activity, MCP setup/status
  Explicit MCP values                  [Save / replace]

Uninstall confirmation
  Remove plugin, private data and owned resources?
  [Cancel]                             [Uninstall]
```

## Validation

Validated with dev at b4d93c926 (draft preservation, queue capacity/recovery, mode-save notifications and Tape hash contracts).

- Format/format check, i18n, lint, main/renderer typecheck, and app/CLI builds passed.
- Plugin, session, ACP, route and runtime suites: 1,160 passed; 132 native SQLite cases skipped under Node are covered by the Electron run.
- Native session and Tape suites under Electron: 700 passed, including transcript projection, Queue/Steer rollback and real SIGKILL recovery. One standalone child-worker fixture is exercised through its parent crash tests.
- Renderer draft, mode, queue and chat suites: 172 passed across eight files.
- Two real Electron plugin smoke tests passed: ZIP lifecycle with working stdio MCP and uninstall confirmation; authenticated HTTP MCP with credential rotation and hook context in actual provider requests. Outgoing credentials change while persisted headers remain templates.
- Static CLI validation passed for original Ponytail 4.9.0. Original-package Git/ZIP and hook acceptance covers upstream commit 0a4dd63ad4541f4f655c4108a295916f3c1d8fda, including modes, child inheritance, shared state, resume, compact and replay.
- Native and visual validation covers macOS. Windows/Linux native validation remains pending.

## Compatibility

This implements a context-hook and MCP configuration subset. Private Git authentication, marketplaces, Codex `SessionStart/clear`, tool/permission decision hooks, ACP propagation and OpenAI connector IDs are outside this delivery. Clearing messages starts fresh hook history on the next input. Native hook commands run with the user's permissions; snapshot integrity checks are not an OS sandbox.

Normal build-generated provider and ACP catalog refreshes are included. No dependencies were added.

[Technical specification](https://github.com/ThinkInAIXYZ/deepchat/blob/codex/user-plugin-compatibility-plan/docs/features/user-plugins/spec.md) · [Implementation and verification](https://github.com/ThinkInAIXYZ/deepchat/blob/codex/user-plugin-compatibility-plan/docs/features/user-plugins/plan.md) · [Authoring guide](https://github.com/ThinkInAIXYZ/deepchat/blob/codex/user-plugin-compatibility-plan/docs/features/user-plugins/authoring.md)

## Summary by CodeRabbit

* **New Features**
  * Added support for installing user-authored plugins from Git, ZIP archives, or local directories.
  * Added package inspection, selective Skills, hooks, and MCP approval, plus enable, disable, update, uninstall, and retry workflows.
  * Added secure MCP variable and credential configuration, plugin context contributions, diagnostics, and localized settings UI.
  * Updated available agent and model metadata.

* **Security & Reliability**
  * Strengthened package, archive, hook, and MCP validation.
  * Prevented disabled or outdated plugin contributions from executing.

* **Documentation**
  * Added comprehensive user-plugin authoring and lifecycle guidance.


zerob13 (commented):
## Review: feat(plugins): support Codex user packages

**Verdict: Request changes** — 0 P1 / 0 P2 / 3 P3 / 2 nits. No must-fix blockers; the P3s are consistency/robustness items worth addressing before or shortly after merge.

### Strengths (verified)

- Supply-chain hardening is thorough and correct: git transport lockdown (`core.hooksPath=/dev/null`, empty credential helper, `protocol.file/ext.allow=never`, `http.followRedirects=false`, `GIT_CONFIG_NOSYSTEM`/`GIT_CONFIG_GLOBAL=/dev/null`, depth-1, no submodules), hand-parsed ZIP central directory rejecting links/special files/encrypted entries, `O_NOFOLLOW` + inode re-check against TOCTOU, symlink and `nlink > 1` rejection, NFC-normalized duplicate-path detection.
- Package parsing validates everything untrusted: `safe-regex2` on hook matchers, 1 MiB config caps, 64-hook/32-server limits, traversal-safe path joining with reserved-name checks, strict hook output schema (only `hookSpecificOutput`/`systemMessage`/`suppressOutput` accepted).
- Crash-safety is real, not decorative: pending-update journal with MCP snapshot rollback, atomic `.publishing-*` rename publication, previous-digest retention for interrupted-update recovery, promise-chain serialization of all mutations.
- Context hooks: invocation identity persisted in Tape, budgets enforced (8 KiB context / 64 KiB stdout / 1 MiB stdin / 10 s boundary), prompt-injection framing in projected sections, authority re-validation before provider dispatch, hook failure never blocks the turn.
- No new dependencies (reuses `gray-matter`, `safe-regex2`, `fflate`); all shared-type changes additive; i18n complete across all 20 locales with identical key sets; docs follow the `docs/features/<name>/{spec,plan,authoring}.md` convention.

### P3 findings

1. **Raw checkbox inputs instead of the shadcn-vue primitive** — `src/renderer/src/pages/plugins/UserPluginInstallDialog.vue:223,236,249` use `<input type="checkbox">` while the analogous import flows (`SkillsPanel.vue`, `ImportSkillsFromAgentDialog.vue`) use `@shadcn/components/ui/checkbox`. Repo convention prefers existing shadcn-vue primitives; this also renders differently from every other checkbox in Settings.

2. **Install-root layout duplicated across modules** — `src/main/skill/index.ts:5296` reconstructs `userData/plugins/user` by hand, which must stay in lockstep with `PluginService.getPluginInstallRoot()` (`src/main/plugin/index.ts:1874`) and `UserPlugins.root()`. The `ownerPluginId` check still covers the primary path, so this is defense-in-depth that can silently rot when the layout changes. Inject the root or share a constant instead of reconstructing it.

3. **Unbounded session-history cache** — `UserPluginHooks.sessions` retains every session's invocation history (including stored inputs, up to 1 MiB prompt per invocation) for the process lifetime with no eviction. Long-lived processes with many sessions will accumulate memory. Consider evicting on session close or capping the cache.

### Nits

4. `pluginsInspectSourceRoute.output` uses `z.custom<PreparedUserPlugin>()` (pass-through) while sibling plugin routes validate their outputs — inconsistent contract strictness.
5. The `directory` source kind is supported end-to-end (contract, main process, install dialog for updates) but has no first-install UI entry. If it is the intended developer path, a line in `authoring.md` on how it is triggered would help.

### Verification

- `pnpm run typecheck` — pass
- `pnpm run lint` — 0 warnings, 0 errors
- `pnpm run i18n` — pass (20/20 locales, key sets identical)
- `vitest run test/main/plugin/userPlugins.test.ts test/main/plugin/userPluginLifecycle.test.ts test/main/skill/skillServiceSharedSkills.test.ts test/main/skill/skillExecutionAuthority.test.ts` — 35/35 pass

src/main/plugin/userPluginPackage.ts
```diff
@@ -0,0 +1,355 @@
...
+      ? [relative]
+      : fs
+          .readdirSync(skillDirectory, { withFileTypes: true })
+          .filter((entry) => entry.isDirectory())
+          .map((entry) => `${relative}/${entry.name}`)
+    for (const skillRoot of roots) {
+      const filename = pluginRelativePath(root, `${skillRoot}/SKILL.md`)
+      if (!fs.existsSync(filename)) continue
+      if (fs.statSync(filename).size > 1024 * 1024) throw new Error('SKILL.md exceeds 1 MiB')
+      const metadata = matter(fs.readFileSync(filename, 'utf8')).data
```
yyhhyyyyyy: **P0.** `gray-matter` evaluates `---js` front matter with `eval`. A SKILL.md that starts with `---js` runs arbitrary Node code in the main process the moment the user clicks Inspect, before anything is shown or approved. I reproduced this locally against gray-matter 4.0.3.

Same call in `skill/index.ts:3759`. The `javascript` engine has to be disabled on every `matter()` call (the existing skill import paths share the exposure, this PR just adds the remote URL route to it).

src/main/mcp/environmentBindings.ts
```diff
@@ -0,0 +1,15 @@
...
+export function resolveMcpEnvironmentBinding(
+  value: string,
+  variables: unknown,
+  environment: NodeJS.ProcessEnv = process.env
+): string {
+  if (!Array.isArray(variables)) return value
+  const allowed = new Set(variables.filter((item): item is string => typeof item === 'string'))
+  return value.replace(/\$\{(?:env:)?([A-Za-z_][A-Za-z0-9_]*)\}/g, (match, name: string) => {
+    if (!allowed.has(name)) return match
+    const resolved = environment[name]
```
yyhhyyyyyy, line 14: **P1.** The allowed list comes from the package itself and the value comes from `process.env`. An HTTP server entry with `bearer_token_env_var: "OPENAI_API_KEY"` and its own URL sends the host's key as a header on first connect, no local code involved. The detail page only asks the user to fill in variables that are missing from `process.env`, so anything already present is used silently.

src/main/skill/index.ts
```diff
@@ -3749,6 +3756,17 @@ export class SkillService implements SkillServicePort {
       throw new Error(`Plugin skill "${input.id}" is missing SKILL.md`)
     }
 
+    const candidate = matter(fs.readFileSync(skillPath, 'utf8')).data.name
```
yyhhyyyyyy, line 3760: **P1, regression.** This check also runs for official plugins. If the user's `skills/` directory has a skill named `computer-use` or `feishu-tools`, `discoverSkills` puts the user copy in `metadataCache` with no `ownerPluginId`, this throws, and the whole CUA / Feishu plugin (MCP included) fails to activate on restart. Before this PR the duplicate was logged and skipped.

src/main/plugin/userPluginHooks.ts
```diff
@@ -0,0 +1,458 @@
...
+    const deadline = AbortSignal.timeout(10000)
+    const signal = input.signal ? AbortSignal.any([deadline, input.signal]) : deadline
+    const operation = this.queue.then(async () => {
+      if (signal.aborted) return
+      const history = this.history(input.sessionId)
+      for (const owner of this.owners.values()) {
+        const previous = [...history.values()].filter(
+          (item) => item.pluginId === owner.pluginId && item.digest === owner.digest
+        )
+        if (!input.source && previous.some((item) => item.messageId === input.messageId)) continue
```
yyhhyyyyyy: **P1.** Edit-and-resend keeps the same messageId (`editUserMessage` updates in place, then `retryMessage` passes it as `preStreamAnchorMessageId`), so this skips `UserPromptSubmit` and the model gets context computed from the old prompt. Tape keeps the old prompt in `input` too. The reuse key needs the prompt hash, not only the messageId.

src/main/plugin/userPlugins.ts
```diff
@@ -0,0 +1,571 @@
...
+        } catch (error) {
+          await this.deactivate(record, false)
+          if (previous) {
+            await this.deps.mcpSettings.restorePluginUpdate(previous.pluginId)
+            const restored = {
+              ...previous,
+              enabled: previous.enabled && !this.revoked.has(previous.pluginId)
+            }
+            this.save(restored)
+            if (restored.enabled) await this.activate(restored)
```
yyhhyyyyyy, line 246: **P2.** If this `activate(restored)` (or the `deactivate` on L215) throws, `writePending(null)` on L225 never runs. The record stays `enabled: true` with nothing registered, and install/enable/configureMcp all return "Restart DeepChat" until restart. `initialize()` L124-130 has the same gap: a failed recovery never clears pending, so the lock survives restarts and blocks every user plugin.

yyhhyyyyyy (commented):
Read through the plugin host, hook runtime, MCP import and the new UI. The Git/ZIP hardening and the ownership model look solid. Inline comments below are the things I think need to land before merge: one P0 (code runs on Inspect), three P1s, and a handful of P2s.

src/main/plugin/userPluginHooks.ts
```diff
@@ -0,0 +1,458 @@
...
+
+  hasHooks(): boolean {
+    return this.owners.size > 0
+  }
+
+  private history(sessionId: string): Map<string, Invocation> {
+    let history = this.sessions.get(sessionId)
+    if (!history) {
+      history = new Map()
+      for (const row of this.tape.getBySession(sessionId)) {
```
yyhhyyyyyy, line 102: **P2.** This cache is filled once and never invalidated. `clearMessages` deletes every Tape row for the session via `resetTape`, but the cached history still says SessionStart ran, so it is not re-run, and the `Tape entry N` references injected into the prompt point at rows that no longer exist.

src/main/plugin/userPluginHooks.ts
```diff
@@ -0,0 +1,458 @@
...
+      entryId: row.entry_id
+    })
+  }
+
+  async accept(input: PluginContextInput): Promise<void> {
+    if (!this.hasHooks()) return
+    const deadline = AbortSignal.timeout(10000)
+    const signal = input.signal ? AbortSignal.any([deadline, input.signal]) : deadline
+    const operation = this.queue.then(async () => {
+      if (signal.aborted) return
```
yyhhyyyyyy: **P2.** The 10s deadline starts at `accept()` but the queue is shared across sessions. With a few concurrent sessions or subagents the later ones reach this line already aborted and return silently: no `failed` row in Tape, nothing in diagnostics. Persisting a failed invocation here would at least make it visible.

src/renderer/src/pages/plugins/UserPluginDetails.vue
```diff
@@ -0,0 +1,257 @@
...
+      }}</DcButton>
+    </section>
+    <UserPluginInstallDialog
+      v-model:open="updateOpen"
+      :kind="details.source.kind"
+      :source="details.source"
+      :plugin-id="plugin.id"
+      @installed="catalog.commitPluginMutation"
+    />
+    <Dialog
```
yyhhyyyyyy: **P2.** Uninstall is destructive and the design system requires `AlertDialog` for delete. `DcConfirmDialog` (used by `SkillDetailDialog`) already wraps it. A plain `Dialog` here also means `lint:alert-dialog-contract` never sees this confirmation.

src/renderer/src/pages/plugins/UserPluginInstallDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+              t('settings.userPlugins.allowMcp')
+            }}</span></label
+          >
+          <div v-for="server in prepared.package.mcpServers" :key="server.name" class="space-y-1">
+            <p class="font-medium">{{ server.name }} · {{ server.type }}</p>
+            <pre class="whitespace-pre-wrap break-all rounded bg-muted p-2 text-xs"
+              >{{ server.url ?? [server.command, ...(server.args ?? [])].join(' ')
+              }}{{ server.cwd ? `\nCWD: ${server.cwd}` : '' }}</pre
+            >
+            <details v-if="server.env || server.headers" class="text-xs">
```
yyhhyyyyyy: **P2.** `env` and `headers` are collapsed by default, but the checkbox says the user reviewed "these commands and endpoints". Headers are where `${OPENAI_API_KEY}`-style bindings live (see the `environmentBindings.ts` comment), so they should be visible without expanding.

PR #2243, opened 2026-09-06
refactor(tape): pin hash contracts, remove dead verification paths

yyhhyyyyyy:
## Summary

This branch writes down which Tape hashes are on-disk contracts and why, removes two hash checks that nothing depends on, and deletes a replay observation reader that has had no caller since #1952. Nine hash-related paths were ablated one at a time against the tape, session, agent, skill and tool suites before deciding what to remove. No Tape schema, fact name, provenance key or hash algorithm changes.

## What changed

**The hash contract is documented where people will trip over it.** `canonicalJson.ts` now says which persisted identities still come from the legacy `hashJson` (the schema-4 ViewManifest that ordinary chat writes, tool fact provenance keys, Execution Journal v1 operation and run keys, `responseHash` and `errorHash`), that those producers depend on its `undefined` coercion, and what a drifted digest does: duplicate tool facts on backfill, failed manifest verification, and every historical run classified as corruption because Journal parsing rebuilds each provenance key and compares it with the stored one. The old comment implied the legacy path was read-only. `tape-system.md` gains a paragraph in the compatibility section that separates the two reader kinds: ViewManifest passes stored fields through and recomputes over them, so writer field changes leave old rows verifiable while a canonicalizer or hashable-transform change does not; ExecutionContract, tool surface facts and Journal payloads use exact-key checks, so removing any field or adding a required one makes old rows malformed.

**`verifyExecutionContractHash` is gone.** It had no production caller. `isDeepChatExecutionContract` already recomputes the contract hash after the exact-key and ceiling checks, and keeping a weaker hash-only verifier beside the strict guard invited the wrong pick. The two tests that used it now assert through the strict guard.

**`isStoredExecutionProvenance` no longer recomputes `hashData(ceilings)`.** It compared the result with `provenance.internalExecutionPolicyHash` on every contract validation, which runs on each tool dispatch, stored-manifest verification and pause/resume restore. Both `ceilings` and the provenance block sit inside the draft that `contractHash` covers, so the comparison could only fail when `contractHash` was about to fail anyway. Ablating it changed no result across 3194 tests. The field is still written as provenance; the tampering test now also pins that a stale policy hash is rejected through `contractHash`.

**`readCausalObservationSlice` is removed.** It joined a message's manifests, provider traces and terminal transcript row into a `DeepChatCausalObservationSlice`, and nothing has called it since the layered runtime migration in #1952: no IPC route, no agent tool, no Inspector projection. The Trace Inspector covers the same ground through its own list and detail projection. Deleting it takes the facade forward, the `getTerminalMessageReader` provider that only it used, the `DeepChatCausalObservation*` type family, thirteen tests whose only subject was the method, and five harness helpers. `exportReplaySlice` and `buildReplaySlice` are untouched.

## Ablation

| Removed | Failing tests | Reader | Result |
| --- | --- | --- | --- |
| `internalExecutionPolicyHash` recompute | 0 of 3194 | self-check only | removed |
| `internalExecutionPolicyHash` field | 94 across 8 files | exact-key | on-disk, kept |
| tool surface `projectionHash` | 2 historical V1 fixtures | verify path | on-disk, kept |
| Journal `errorHash` | 1 (spec test) | none | spec contract, kept |
| `messageContentHashByMessageId` | 5, incl. 4 harness integration | pinned fail-closed | load-bearing, kept |
| post-extraction skill `assertAuthorityCurrent` | 2 | TOCTOU layer | deliberate, kept |
| replay `sliceHash` / `sliceId` / per-entry hashes | 1 / 0 / 0 | none | see below |

## Compatibility

- Tape: no schema, fact name, provenance key, payload field or hash algorithm change. Contracts still write `internalExecutionPolicyHash`.
- `isDeepChatExecutionContract` accepts a contract whose policy hash disagrees with its ceilings when `contractHash` is consistent. Such a contract can only be built by hand; `buildExecutionContract` always writes both from the same ceilings.
- Removed from `@shared/types/tape-replay`: `DeepChatCausalObservation*`. Removed from tape ports: `TapeTerminalMessageRow`, `TapeTerminalMessageReader`, `TapeApplicationDatabase.deepchatMessagesTable`, `TapeApplicationProviders.getTerminalMessageReader`. Removed from `SessionTape`: `readCausalObservationSlice`. Nothing in `src/renderer` or the route contracts referenced any of them.
- Left alone on purpose: the `sessions.exportMessageTapeReplaySlice` route has no renderer caller, which is why the replay `sliceHash`, `sliceId` and per-entry hashes have no reader; whether that export surface stays is a product call, not a hash one. `getViewManifestSourceMaps` hashes every effective user record while only the pinned one is read; fixing that changes a port signature for a millisecond-scale win and was not worth it here.

## Validation

- `format:check`, `lint`, `typecheck` (node and web): pass.
- `test/main/tape`, `test/main/session`, `test/main/routes`: 1017 passed, 161 skipped (all `itIfSqlite`, native module unavailable under the Node ABI).
- `test/main/agent/deepchat`, `test/main/skill`, `test/main/tool`, `architectureBaseline`: 2680 passed.
- `test/main/scripts/buildCli.test.ts` fails locally on the packaged-host launcher case; the branch does not touch CLI or packaging and the failure was not reproduced against `dev`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Tape replay compatibility, persisted hash contracts, manifest invalidation, and canonicalization rules.
  * Updated Inspector documentation to reflect the removal of message replay/export capability.

* **Changes**
  * Removed replay-slice export and causal-observation replay capabilities.
  * Removed terminal-message persistence interfaces.
  * Execution-contract validation now uses structural provenance and contract comparisons.

* **Tests**
  * Updated replay and execution-contract coverage for the revised behavior.


zerob13 (changes requested):
## Review: refactor(tape): pin hash contracts, remove dead verification paths

**Verdict: Request changes** — 0 P1 / 0 P2 / 2 P3 / 1 nit. No must-fix blockers; both P3s are small and touch files this PR already edits, so they are worth taking before merge.

Reviewed at `c9af82494` against merge-base `61839ea9c` (10 files, +48/-1061).

### What was verified

- **All removed symbols are dead, worktree-wide.** `readCausalObservationSlice` (service + `SessionTape` delegation), the seven `DeepChatCausalObservation*` shared types, `verifyExecutionContractHash`, `TapeTerminalMessageReader`/`TapeTerminalMessageRow`, `getTerminalMessageReader`, and the `deepchatMessagesTable` key on `TapeApplicationDatabase` have zero remaining references across `src/`, `test/`, `docs/`, `scripts/`, `plugins/`. No barrel re-export, DI registration, IPC route, or preload exposure touches them; the renderer imports only the surviving `DeepChatTapeReplay*` types. No orphaned private helpers remain in `viewReplayService.ts`.
- **Both removed hash checks are behaviorally redundant.** `verifyExecutionContractHash` had test-only callers since it was introduced. The `internalExecutionPolicyHash === hashData(ceilings)` recheck is subsumed by the full binding at the end of `isDeepChatExecutionContract` (`buildContractHash(draft) === contractHash`): the producer computes the policy hash and the contract hash from the same input in one atomic build, so divergence requires a writer bug the recheck never guarded against in practice. The new `staleProvenance` test case correctly pins the surviving tamper detection.
- **Tests executed** in a clean worktree at the PR head: `vitest run --config vitest.config.ts test/main/tape/executionContract.test.ts test/main/session/data/tapeViewReplay.test.ts` → **42/42 pass**.
- **Docs producer taxonomy is accurate** against the remaining code: `hashJson` still produces schema-4 manifest hashes (`viewManifest.ts:176,475-476`), tool fact provenance keys (`factPersistence.ts:191`), Journal v1 keys/hashes (`executionJournal.ts:491,510,601,631`); `canonicalJsonStringifyData` covers hashVersion 3+ manifests, contract hashes, tool surface facts, Journal v2, skill materialization.
- "No tape schema, fact name, provenance key or hash algorithm changes" — confirmed: constants and both hash function bodies are untouched; only comments/docs were added.

### P3-1 — Stale docs claim left directly below the new paragraph

`docs/architecture/tape-system.md:444` still says `SQLite row parsing、message trace 和 terminal evidence 读取仍属于 TapeViewReplayService`. This PR deleted the only terminal-evidence reading in that service — `getTerminalMessageReader()` existed solely for the removed `readCausalObservationSlice`. Since the new paragraph sits directly above this sentence, please drop the dead phrase, e.g. `SQLite row parsing 和 message trace 读取仍属于 TapeViewReplayService，不能反向放进 domain。`

### P3-2 — The deleted "latest physical attempt" test also covered live code

The removed test `binds replay to the latest physical attempt for a request sequence` mostly exercised `readCausalObservationSlice` (dead — fine to delete), but it was also the **only** coverage of `selectLatestTrace`'s tie-break ordering (`physical_attempt` desc → `created_at` desc → `id` desc, `viewReplayService.ts:1265-1276`) through the surviving `exportReplaySlice` path. After this PR, `tapeViewReplay.test.ts` has zero references to `physical_attempt`; the remaining trace tests only vary `request_seq`. Please keep a slimmed variant: two `createTraceRow`s with `request_seq: 2`, `physical_attempt: 1/2` and inverted `created_at`, one manifest at requestSeq 2, assert `exportReplaySlice('s1', 'a1', { requestSeq: 2 })` returns attempt 2.

### Nit — `hashes` object is also excluded from the manifest hashable

The new paragraph pins `ViewManifest 除 assembledAt、viewId 外的全部 stored 字段都进 manifestHash`, but `buildManifestHashable` (`viewManifest.ts:163-173`) additionally replaces the whole `hashes` object with only `{ promptHash, toolDefinitionsHash }`. Since the paragraph's purpose is to pin the hashable contract exactly, make that exclusion explicit.

### Not verified from the PR body

- The nine-path ablation and the full tape/session/agent/skill/tool suite runs are process claims; I only executed the two directly touched test files (both green). Static analysis found no dangling references, signature mismatches, or layer-boundary violations that would predict other-suite failures.
- "No caller since #1952": git history shows the reader never had a production caller at any point — the dating is imprecise in the direction that favors removal, so no concern.

zerob13 (changes requested):
## Review: refactor(tape): pin hash contracts, remove dead verification paths (re-review)

**Verdict: Request changes** — all three items from the previous review (`c9af82494`) are addressed and verified, but the legacy-`hashJson` pin list this PR introduces omits one live on-disk producer: `errorHash`. 0 P1 / 0 P2 / 1 P3.

Re-reviewed at `3356fef66` (incremental diff since the previous review: 2 commits, 2 files, +20/-5, docs + test only; the deletion set and everything else verified at `c9af82494` is untouched).

### Previous findings — all resolved ✅

- **P3-1 (stale terminal-evidence claim) — fixed.** `docs/architecture/tape-system.md` now reads `SQLite row parsing 和 message trace 读取仍属于 TapeViewReplayService`; the dead phrase is gone.
- **P3-2 (lost `physical_attempt` tie-break coverage) — fixed.** New test `binds replay to the latest physical attempt for a request sequence` is exactly the slimmed variant requested: two trace rows at `request_seq: 2` with `physical_attempt: 1/2` and inverted `created_at`, one manifest at requestSeq 2, asserting `exportReplaySlice('s1', 'a1', { requestSeq: 2 })` returns attempt 2 — pinning `selectLatestTrace`'s primary tie-break through the surviving export path.
- **Nit (`hashes` exclusion) — fixed.** The hashable paragraph now states the `hashes` object is reduced to `{ promptHash, toolDefinitionsHash }`; matches `buildManifestHashable` (`viewManifest.ts:163-172`) exactly.

### P3 — Pin list omits `errorHash`, a live legacy `hashJson` producer

`docs/architecture/tape-system.md` (回放和兼容 section) pins legacy `hashJson` as still producing `schema-4 manifest hash、tool fact provenance key 与 Execution Journal v1 key/responseHash`. It also still produces **`errorHash`**: `buildRunTerminalData` writes `errorHash: hashJson(input.errorMessage)` for protocol-version-1 run-terminal facts (`executionJournal.ts:631`). The field is persisted, exact-key checked, and re-derived on read (`executionJournal.ts:936-950`) — the same on-disk category as the listed `responseHash` (`executionJournal.ts:601`). Migrating it to `hashJsonData` would mark every stored error-run terminal as malformed, which is exactly the drift this paragraph exists to prevent. Please extend the list, e.g. `Execution Journal v1 key/responseHash/errorHash`.

I audited every remaining `hashJson(` call site in `src/main` against the pin list — this is the only omission. For the record, the two that look like gaps but are not: `toolSurfaceProvenanceService.ts:244` re-derives the schema-≤4 manifest `toolDefinitionsHash` on the verification side (same value as `viewManifest.ts:476`, covered by the manifest-hash item), and `sliceId`/`sliceHash` (`viewReplayService.ts:1194`, `domain/replay.ts:321`) are transient export products never persisted or re-validated by main, so they are correctly outside the on-disk pin scope.

### Verified

- Tests executed at the PR head in a clean worktree: `vitest run --config vitest.config.ts test/main/session/data/tapeViewReplay.test.ts test/main/tape/executionContract.test.ts` → **43/43 pass** (28 + 15, including the new case).
- Incremental commits touch no production code beyond the already-reviewed deletion set.

### Not verified from the PR body

Unchanged from the previous review: the nine-path ablation and full-suite runs remain process claims; no static issue found in the incremental commits.

zerob13 (approved):
## Review: refactor(tape): pin hash contracts, remove dead verification paths (follow-up `3d4130ce4`)

**Verdict: Approve.** All three incremental commits verified against `3356fef66..3d4130ce4`. The prior CHANGES_REQUESTED is fully resolved and no new blocking findings. 0 P1 / 0 P2 / 0 P3.

### Prior P3 resolution (verified)
- **errorHash pin-list omission — fixed** by `3d4130ce4`. `docs/architecture/tape-system.md` now lists `errorHash` alongside the Journal v1 key/`responseHash` among legacy `hashJson` producers that must not be migrated to `hashJsonData`. This matches `buildRunTerminalData`, which writes `errorHash: hashJson(errorMessage)` on protocol-version-1 run-terminal facts, and the existing `canonicalJson.ts` comment. Minimal, correct docs-only delta.

### New commit: `41eef8228` — share the exact-key and record guards
The consolidation is justified (five `hasExactKeys` copies + four `isRecordObject` bodies under two names) and provably behavior-preserving. Verified caller-by-caller:

- The shared `hasExactKeys` in `domain/primitives.ts` is required-set ⊆ enumerable own keys ⊆ required∪optional — exactly what every prior local copy enforced on JSON data. `executionContract`'s optional-keys variant keeps the identical signature and call shape (`PROMPT_SECTION_KEYS.slice(0, 3), PROMPT_SECTION_KEYS.slice(3)`).
- `executionJournal.requireExactKeys` delegates with the same `${label} has unsupported or missing fields.` throw — no error-message or stored-shape change.
- The `taskContract` / `toolSurfaceFacts` / `traceInspectorProjection` exact-equality variants (sorted-array compare, length+includes) are set-equivalent to the shared form with `optionalKeys = []`.
- `toolSurfaceFacts` call sites still pre-check `isPlainRecord` before `hasExactKeys` (checked at merge-base and at HEAD) — the prototype guard is not bypassed by the shared signature's widening to `unknown`.
- `traceInspectorProjection.isPlainObject` → shared `isRecordObject` is equivalent for every falsy/non-object input; the separate `hasOnlyKeys` subset check is correctly retained for the optional-field projections.

### New commit: `897693b77` — remove the unused replay export surface
The dead-surface claim is true and the removal is complete:

- At merge-base, the only renderer reference to `sessions.exportMessageTapeReplaySlice` was the `SessionClient` wrapper's own definition/export — zero callers in pages, stores, or preload. The full chain (route contract → registry entry → session route handler → `SessionQuery` → data port → `SessionTape.exportReplaySlice` → `TapeViewReplayService.exportReplaySlice`/`buildReplaySlice` → `DeepChatTapeReplay*` type family) was reachable only from tests.
- No dangling references at HEAD: `exportReplaySlice` / `buildReplaySlice` / `selectLatestTrace` / `findReplayTrace` / `withReplaySliceHash` / `sliceHash` / `sliceId` / `types/tape-replay` resolve to zero hits across `src/` and `test/`.
- `getMessageTraceReader` is not orphaned — `traceInspectorService.ts` still consumes it for the shipped Inspector path; only the replay service's dependency was dropped.
- `TapeViewReplayService` retains its live consumers (manifest append/read, source maps, `getViewManifestByExecutionBinding`, runtime skill-view recovery) — the class is not gutted beyond the dead export.
- Test delta is a fair trade, not coverage loss: the 9 deleted tests (including the `4bfa230ce` latest-physical-attempt test) exercised only the removed export surface; manifest-integrity assertions were repurposed onto the live `getViewManifestByExecutionBinding` path, and the `layerBoundaries.test.ts` allowed-import fixture substitution (`tape-replay` → `tape-view-manifest`) is mechanical.
- Docs stay consistent: `tape-system.md` drops the export-specific sentences (ReplaySlice derivation clause, message-trace reading ownership) and the Inspector spec drops the delegated "Message replay/export" row.

### Breaking-change assessment
Surface removal only: an internal route/IPC contract plus its renderer wrapper, zero callers, no persisted-shape changes, no migration, no user-visible behavior. Nothing requiring flagging.

### Verification
- `pnpm vitest run --config vitest.config.ts` on `tapeViewReplay.test.ts` (19), `query.test.ts` (12), `session.integration.test.ts` (135), `layerBoundaries.test.ts` (46), `dispatcher.test.ts` (72) — 284/284 pass.
- `pnpm run typecheck` (node + web) — clean.
- `oxfmt --check` on all 19 changed source files — clean; `oxlint` — 0 warnings, 0 errors.

Good to merge from my side.

PR #2246, opened 2026-09-07
fix(chat): preserve workspace session history

zerob13:
Workspace history could disappear when starting a conversation: background refreshes replaced loaded pages and raced with pagination, while automatic Agent selection changed the sidebar filter and hid other Agents' history.

Preserve loaded sessions and their pagination position during background refreshes, block next-page requests until the first-page request settles, and retain stale-response guards. Refreshing an exhausted list keeps pagination complete.

Keep the sidebar Agent filter explicit across conversation creation, activation, and restoration. A workspace's new conversation uses the explicit Agent selection first. Under All Agents, it uses the Agent of that workspace's latest regular, non-draft conversation, including pinned conversations. Unavailable Agents, empty history, and failed reads use the existing fallback. A bounded lightweight database query finds unloaded history, and delayed results cannot override newer navigation or an explicit Agent filter change.

```text
BEFORE                          AFTER
[All Agents] -> workspace +     [All Agents] -> workspace +
[Agent A]                      [All Agents]
Workspace                      Workspace
  Agent A history only           Agent A history
  [older pages may vanish]       Agent B history
  [scroll cannot load more]      ... load more
New conversation: Agent A       New conversation: latest workspace Agent
```

Validation:

- Session store: 100 tests; Agent store: 5 tests; sidebar: 72 tests passed with the CI heap limit, `NODE_OPTIONS=--max-old-space-size=6144`.
- Main-process query, session service, route contracts, and session table: 78 tests passed using Electron's Node runtime with `DEEPCHAT_REQUIRE_NATIVE_SQLITE=1`. The persisted-history regression covers the typed query through real SQLite, filtering before pagination and excluding ineligible prioritized records.
- Regression coverage includes new-session creation with a delayed background refresh, stale pagination responses, exhausted pagination, explicit Agent precedence, unavailable history, and superseded workspace lookups.
- `pnpm run format`, `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, `pnpm run typecheck`, `pnpm run architecture:renderer-baseline:check`, and `pnpm run icons:check` passed.

## Summary by CodeRabbit

- **New Features**
  - Added separate Agent filtering that remains stable when opening or restoring conversations.
  - New workspace conversations can select the most recently used enabled Agent, with fallback behavior when unavailable.
  - Session history can now be filtered by workspace and draft status.

- **Bug Fixes**
  - Refreshing session history preserves previously loaded entries and pagination position.
  - Prevented duplicate or overlapping page loads during active requests.
  - Exhausted session history remains correctly marked after refreshes.

### DIFF
--- a/src/main/session/data/tables/newSessions.ts
+++ b/src/main/session/data/tables/newSessions.ts
@@ -271,6 +271,8 @@ export class NewSessionsTable extends BaseTable {
     limit?: number
     cursor?: SessionListPageCursor | null
     agentId?: string
+    projectDir?: string
+    includeDrafts?: boolean
     includeSubagents?: boolean
     parentSessionId?: string
   }): SessionListPageResult {
@@ -278,6 +280,15 @@ export class NewSessionsTable extends BaseTable {
     let sql = 'SELECT * FROM new_sessions'
     const conditions: string[] = []
     const params: unknown[] = []
+
+    if (options?.projectDir !== undefined) {
+      conditions.push('project_dir = ?')
+      params.push(options.projectDir)
+    }
+
+    if (options?.includeDrafts === false) {
+      conditions.push('is_draft = 0')
+    }
 
     if (options?.agentId) {
       conditions.push('agent_id = ?')
--- a/src/renderer/src/stores/ui/session.ts
+++ b/src/renderer/src/stores/ui/session.ts
@@ -725,6 +725,10 @@ export const useSessionStore = defineStore('session', () => {
   const hasActiveSession: ComputedRef<boolean> = computed(() => activeSessionId.value !== null)
 
   const newConversationTargetAgentId = computed(() => {
+    if (agentStore.filterAgentId) {
+      return agentStore.filterAgentId
+    }
+
     const selectedAgentId =
       typeof agentStore.selectedAgentId === 'string' ? agentStore.selectedAgentId.trim() : ''
     if (selectedAgentId) {
@@ -762,7 +766,7 @@ export const useSessionStore = defineStore('session', () => {
       return
     }
 
-    agentStore.setSelectedAgent(targetAgentId)
+    agentStore.setSelectedAgent(targetAgentId, { preserveFilter: true })
   }
 
   const applySessionStatus = (sessionId: string, status: string, version?: number): boolean => {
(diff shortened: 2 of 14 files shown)


src/renderer/src/stores/ui/session.ts
```diff
@@ -1229,7 +1238,36 @@ export const useSessionStore = defineStore('session', () => {
...
+        if (latestAgentId && agentStore.enabledAgents.some((agent) => agent.id === latestAgentId)) {
+          targetAgentId = latestAgentId
+        }
+      } catch (lookupError) {
+        console.warn('[sessionStore] Failed to resolve workspace Agent:', lookupError)
+      }
+
+      if (
+        activationNavigationRequestId !== requestId ||
+        agentStore.filterAgentId !== filterAgentId
```
zhangmo8, line 1264: This guard compares the filter by value against the pre-await snapshot. If the user toggles the filter to a different Agent and back to the same value while the lookup is in flight, the stale result is still applied and the navigation proceeds with the old target. Consider a monotonically increasing filter-change counter compared by identity, mirroring how `activationNavigationRequestId` guards navigation. Not blocking — the race window is tiny.

src/main/session/data/tables/newSessions.ts
```diff
@@ -279,6 +281,15 @@ export class NewSessionsTable extends BaseTable {
     const conditions: string[] = []
     const params: unknown[] = []
 
+    if (options?.projectDir !== undefined) {
```
zhangmo8, line 284: `!== undefined` treats an explicit `null` as a real filter (`project_dir = NULL` never matches, so the query silently returns an empty page). The route schema (`z.string().min(1)`) protects IPC callers, but internal main-process callers of `SessionQuery`/this table bypass it. Suggest `options?.projectDir != null`, or a comment documenting that callers must pass a non-empty string.

zhangmo8 (commented):
Reviewed the full diff. Overall this looks correct and well-tested — nothing blocking from my side.

**What I like**

- Splitting `selectedAgentId` / `filterAgentId` in the agent store is the right fix at the root cause: conversation activation no longer silently rewrites the sidebar filter, and `All Agents` stays sticky as the spec intends.
- The retained-cursor refresh logic is sound: first-page refreshes merge into the existing cursor chain, an exhausted list stays complete after refreshes, and next-page requests are blocked while the initial load is in flight.
- Strong test coverage — especially the real-SQLite regression exercising route parsing → `SessionQuery` → `AppSessionService` → table together, and the `it.each` supersession tests covering all three race sources (newer workspace call, session activation, explicit filter change).

**Non-blocking observations**

1. In `startNewConversation`, the stale-guard compares `agentStore.filterAgentId` by value against the snapshot taken before the await. If the user switches the filter to another Agent and back to the original while the lookup is in flight, the stale result would still be applied. A monotonically increasing change counter (like `activationNavigationRequestId`) would close that window.

2. In `NewSessionsTable.list`, `projectDir !== undefined` also treats an explicit `null` as "filter by `project_dir = NULL`", which matches nothing. The Zod route schema rejects `null` for IPC callers, but internal main-process callers bypass it — `options?.projectDir != null` (or a comment stating the schema guarantees a non-empty string) would make it robust.

3. Cursor-chain retention tradeoff: sessions created in another window between the retained cursor boundary and the newest row only arrive through event-driven merges. If such an event is missed (e.g., created while the window was still hydrating), that row stays invisible until a full reload. The spec appears to accept this — just flagging it; if it ever becomes a problem, a periodic gap check keyed on the current top `updatedAt` could self-heal.

Otherwise LGTM — CI is green, and the design matches the spec document.

PR #2247, opened 2026-09-07
fix(memory): vector admission, fork cursor, scope repair and maintenance drain

yyhhyyyyyy:
## Summary

This PR fixes four bugs in the memory system: vector recall could get stuck in a state where every lease was refused, forking a session re-ran extraction over the whole cloned history, one malformed scope row could keep the app from starting, and database maintenance left consolidation passes running while SQLite was closed and reopened. Each fix has its own commit and a regression test. It touches no schema version, migration, Tape fact or anchor name.

## What was wrong

**Vector admission could close and never reopen.** Five paths closed the `accepting` gate and each bumped `leaseEpoch`, then reopened only if the epoch still matched its own capture. When an embedding config change was draining and a lease timed out, the timeout observation recorded `resumeAllowed=false` and bumped the epoch again; on settle it deferred to the "existing owner", while the transition saw a foreign epoch and also gave up. The agent ended with `accepting=false` and `health='healthy'`: every lease failed with `admission-closed`, recall silently ran FTS-only, and `getRecallHealth` still reported available. The in-lease identity switch had the same shape, and eviction could bump the epoch underneath a draining transition.

**Forking a session replayed its whole history through extraction.** A fork target starts with a NULL memory cursor while its cloned transcript is renumbered from `orderSeq` 1. The first settled turn saw `tail - cursor >= 6` and extracted the entire clone again (dozens of chunks of LLM work). Provenance keys ignore `sessionId`, so re-extracting a superseded claim before its successor revived the old fact and retired the current head.

**A dirty scope row kept the whole app from starting.** `assertCurrentSchema` threw on any `agent_memory` row whose scope pair violated the invariant. That error matched no schema or corruption classifier and propagated to `app.quit()`. Databases migrated through v51 only carry the column CHECKs, so an external tool can leave such rows behind, and unlike the temporal path there was no repair before the assertion.

**Database maintenance did not actually stop maintenance.** `stopBackgroundMaintenance` cleared two timers and left the rest armed, so consolidation passes fired into the window where SQLite is closed, replaced by an import and reopened. `maintenanceStarted` was never reset, so the `startBackgroundMaintenance` after maintenance returned early and dropped the startup arm and prewarm.

## What changed

- `vectorStoreManager`: admission is owned through a hold count. Every owner (identity transition, in-lease switch, timeout observation, drain-and-close) takes one hold and releases it when it settles; admission reopens only when the last hold is released on a healthy state whose open store matches the logical identity, otherwise the store is converged first. A permanent close keeps its hold until the state is settled away. Eviction skips agents whose admission is owned. The observation no longer needs an identity snapshot.
- Fork: `cloneSentMessagesToSession` maps the source's committed extraction cursor onto the filtered, densely renumbered clone. `resetForkTarget` skips only that extracted prefix; later turns extract the unprocessed cloned tail and new messages even if the source never resumes. Cursor seeding advances the target epoch so older extraction work cannot commit its cursor or anchor.
- `agentMemory`: the scope repair runs before the assertion, mirroring `ensureTemporalArtifacts`, and never widens applicability. Agent rows drop a stray id, User rows resync the shadow from the authoritative id or recover a missing or malformed id from a well-formed shadow, Project/Session rows drop a stray shadow. A narrow-scope row whose identity cannot be recovered is deleted and the FTS mirror marked dirty, because `memoryScopeFromRow` would throw on it from every listing, forget or restore, and turning it into an Agent-scope row would widen it. Before deleting a claim covered by a pending clear, scope repair persists its recoverable provenance tombstone and adds every deleted row to the job's removed count in the same transaction. Tombstones use the clear job's timestamp, and no content scope is invented. Both startup repairs run with the clear-job guard suspended in the same transaction, since a pending clear on the same agent otherwise vetoed the repair UPDATEs and the app failed to open on every launch. The trigger predicate, the assertion and the repair share one row-level SQL constant.
- `maintenanceService`: stop clears every timer, refuses new arming and passes, and invalidates the execution fence of every agent with an in-flight pass, which aborts its provider requests so the pass and the challenge/reflection/persona sub-services stop at their next checkpoint instead of waiting out a 60s provider deadline. `drainBackgroundMaintenance` waits for tracked passes within a bounded timeout and returns the agents still running; `runDatabaseMaintenance` fails instead of closing SQLite while a pass is live. Passes are tracked by agent so directly requested passes are drained too, and start re-arms after stop.

## Docs

`docs/architecture/memory-system.md` records the fork cursor rule, the scope repair contract (no widening, preserve pending-clear provenance tombstones, delete unrecoverable rows, clear guard suspended during repair) and the stop/drain/start sequence. `docs/architecture/memory-vector-store-v2/plan.md` describes the hold-count ownership of `accepting`.

## Verification

- `pnpm run format:check`, `pnpm run lint`, `pnpm run i18n`, `pnpm run typecheck`
- `pnpm run test:memory`: 54 files, 899 tests
- Four native SQLite memory suites under Electron's Node: 143 tests
- Focused transcript/runtime suites: 125 tests
- Expanded Electron-hosted session/harness/app coverage: 1,324 passed; two startup migration fixture tests fail on both the original PR head and this implementation because their fixture omits `new_session_active_skills`.
- Targeted fork-cursor, pending-clear tombstone, and stale-extraction regressions fail with their production fixes reverted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Forked sessions now resume memory processing at the correct point in copied history, preventing duplicate extraction.
  - Startup recovery repairs malformed memory records, preserves recovery history, removes unrecoverable entries, and maintains cleanup safeguards.
  - Vector-store transitions and timed-out operations now recover more reliably without leaving healthy stores unavailable.
  - Background maintenance now shuts down safely, reports unfinished work, supports bounded draining, and can restart without losing pending tasks.

- **Tests**
  - Added coverage for forked-session memory handling, recovery repairs, maintenance lifecycle behavior, and vector-store recovery.


src/main/agent/deepchat/runtime/transcriptMutationCoordinator.ts
```diff
@@ -106,9 +109,10 @@ export class TranscriptMutationCoordinator {
     this.deps.runLifecycle.transitionCurrentStatus(sessionId, 'idle')
   }
 
-  resetForkTarget(targetSessionId: string): void {
+  resetForkTarget(targetSessionId: string, clonedTailOrderSeq: number): void {
     const targetInstance = this.deps.registry.getOrHydrateScope(toAppSessionId(targetSessionId)).instance
     this.deps.compaction.reset(targetSessionId, targetInstance)
+    this.deps.memory.seedExtractionCursor(targetSessionId, clonedTailOrderSeq)
```
zerob13: [P2] Preserve cloned messages that the source has not extracted

The source cursor need not have reached the fork point. For example, four short messages below the extraction thresholds leave the source cursor at 0; this call nevertheless marks all four cloned messages as extracted. If the user continues only the fork, its next extraction starts at row 5, and the original four messages never reach `extractAndStore`: forking schedules no catch-up extraction on the source. I reproduced this with a source that makes zero extraction calls and a fork that extracts only rows 5–10; leaving its cursor at 0 includes the original messages. Please distinguish the cloned prefix already covered by successful source extraction from its unprocessed tail, or ensure that tail is ingested before advancing the target cursor past it.

src/main/memory/data/tables/agentMemory.ts
```diff
@@ -1141,6 +1186,87 @@ export class AgentMemoryTable extends BaseTable implements MemoryRepositoryPort
...
+        .all() as Array<{ id: string }>
+    ).map((row) => row.id)
+    let deletedRows = 0
+    if (deletedIds.length) {
+      deletedRows = this.db
+        .prepare(
+          `DELETE FROM agent_memory
+           WHERE (${AGENT_MEMORY_SCOPE_INVALID_ROW_SQL})`
+        )
+        .run().changes
```
zerob13, line 1309: [P2] Preserve pending-clear tombstones before deleting corrupt rows

If a persisted `agent_memory_clear_job` covers a row with an unrecoverable scope, this raw deletion removes the row before `processMemoryClearBatch()` can create its forgetting tombstones. The clear then completes successfully, but the same claim can be inserted again. I reproduced this on the v51 migrated table shape: insert a project-scoped claim with a valid scoped provenance key, lose its `scope_id`, call `beginMemoryClear`, reopen/assert the schema, and finish the clear. `insertClaimUnlessTombstoned` with the original scope/provenance returns a new active row; the control without scope corruption returns `null`. Preserve the recoverable provenance tombstone for rows covered by a pending clear inside the repair transaction before deleting them, so startup recovery does not undo the user's clear request.

zerob13 (commented):
Reviewed commit `35a9aa299387036b24b78b06eec844414da687d2`. Found two P2 correctness issues, detailed inline:

- Fork cursor seeding skips cloned history that the source has never extracted.
- Startup scope repair bypasses pending-clear tombstones, allowing a cleared claim to be inserted again.

Both were reproduced with focused executable checks; the corresponding controls passed (unseeded fork cursor and a clear without scope corruption).

Validation passed: `format:check`, `lint`, `i18n`, `typecheck`, and `test:memory` (896 tests). Electron-hosted native checks passed for the four memory suites (142 tests) and the focused transcript/session suites (65 tests). Related agent-memory/session/app checks under host Node reported 874 passed and 134 skipped.

zhangmo8 (changes requested):
Reviewed the rework at head `35a9aa299` (all CI green: build, static, test-main, test-native-memory, test-renderer, package-impact). The admission hold-count, maintenance fence/drain, and scope-repair machinery are solid and well tested. For transparency: the two P2s zerob13 already flagged inline (fork cursor seeding vs. unextracted cloned history; scope repair bypassing pending-clear tombstones) are both still present at head — no commits were pushed after that review — so I'm marking this as request changes on that basis rather than repeating the details.

Three additional (all P3) findings:

1. `src/main/app/composition.ts:3373-3378` — a >5 s drain (`MAINTENANCE_DRAIN_TIMEOUT_MS`) now throws and fails the entire database-maintenance operation (backup/vacuum/rekey). The non-abortable native-vector-work wedge is exactly the case the new test demonstrates; the fence reopens/re-arms before rethrowing, but this introduces a new user-visible failure mode. Consider a warning-level degradation or a longer/adaptive timeout instead of failing the whole operation.

2. `src/main/agent/deepchat/memory/memoryRuntimeCoordinator.ts:182` — `seedExtractionCursor` writes the cursor with no epoch bump; the "fresh target, no in-flight extraction" invariant is comment-only. If `resetForkTarget` ever ran on a session with an extraction in flight, it would race the chunk commit at line ~587 (last-writer-wins). Cheap to harden with an epoch bump or assert.

3. `src/main/memory/infra/vectorStoreManager.ts` (`evictable()`) — eviction now requires `state.accepting`, so stores held by a pending admission owner are exempt from soft-cap eviction. Bounded and transient (and looks deliberate), but currently undocumented by a test.

Also verified while reviewing: `finishSuspectObservation` is idempotent via its `observation.completed` flag, so no double-release on the suspect path.

src/main/agent/deepchat/runtime/transcriptMutationCoordinator.ts
zerob13: The fork cursor is mapped from the source's committed extraction cursor onto the filtered, densely renumbered clone. Only the extracted prefix is skipped; an unprocessed cloned tail remains eligible even if the source never resumes. Coverage includes a null source cursor, partial extraction, gaps, failed messages, compaction markers, and a source cursor beyond the fork point. Cursor seeding also advances the target epoch, preventing older extraction work from committing its cursor or anchor. The cursor-mapping and stale-extraction regressions fail against the previous implementation and pass with this change.

src/main/memory/data/tables/agentMemory.ts
zerob13: Scope repair persists the provenance tombstone before deleting an unrecoverable claim covered by a pending clear's agent, claims phase, and rowid cutoff. Both operations share the repair transaction, and the tombstone retains the clear job's timestamp. No content scope is invented. The native SQLite regression verifies that replaying the original scoped claim is blocked after the clear completes, while identical content in other scopes and another agent remains writable. It fails against the previous implementation and passes with this change.

zerob13:
Review disposition:

- Both P2 threads are resolved with regressions covering unprocessed fork history and exact forgetting during scope repair.
- Cursor seeding advances the extraction epoch; an in-flight extraction test verifies that an older result cannot commit its cursor or anchor.
- The maintenance drain timeout remains a failure of the database operation. Proceeding while a pass is live would let it access SQLite during close/import/reopen. Provider requests are already aborted; the remaining non-abortable native work still needs to settle. Increasing a fixed timeout does not resolve that race.
- Admission-owned stores remain exempt from soft-cap eviction while a transition or reset owns their lifecycle. This is intentional temporary oversubscription: eviction must not close the store underneath its owner. Existing capacity, idle eviction, active-lease eviction, and transition/timeout suites pass; no production change is needed.

Validation: format, lint, i18n, typecheck, 899 Memory behavior tests, 125 focused transcript/runtime tests, and 143 native SQLite memory tests passed. The targeted fork, tombstone, and epoch regressions fail when their production fixes are reverted. Expanded Electron-hosted session/harness/app coverage passed 1,324 tests; two unrelated startup migration fixture tests fail because their fixture omits `new_session_active_skills`. The same two failures reproduce on the original PR head before these changes.

src/main/memory/data/tables/agentMemory.ts
```diff
@@ -1141,6 +1191,119 @@ export class AgentMemoryTable extends BaseTable implements MemoryRepositoryPort
...
+      for (const row of clearingRows) {
+        if (!row.provenance_key || !isTombstoneEligibleMemoryKind(row.kind)) continue
+        const identity = buildMemoryProvenanceTombstoneIdentity(row.agent_id, row.provenance_key)
+        insertTombstone.run(
+          row.agent_id,
+          identity.identityKind,
+          identity.identityHash,
+          row.created_at
+        )
+      }
```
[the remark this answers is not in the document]
zerob13: Scope repair adds the number of invalid rows covered by each claims-phase job's agent and rowid cutoff to that job's removed_count before deleting the rows, in the same transaction. Counting includes rows without provenance; tombstone creation remains limited to recoverable eligible identities. The native regression covers multiple rows, two pending-clear agents, an unrelated agent, a valid row left for the normal batch, and reopening without double counting. The counting regression fails without the update and passes with it; all 143 native memory tests and the local static/Memory checks pass.

PR #2248, opened 2026-09-07
perf(provider): coalesce model discovery requests

zerob13:
Creating an enabled provider starts model discovery immediately. Initialization plus eight concurrent model-list queries produced 9 HTTP requests on `dev`.

Share the pending discovery request within each provider instance. The same workload produces 1 HTTP request, and a subsequent explicit refresh makes a fresh request. Configuration changes start a new request; results from the old configuration cannot overwrite the current cache or persisted catalog, regardless of completion order. Each caller receives an independent model snapshot, including endpoint arrays, so normalization and caller mutations cannot leak between requests or into cached models. Caller-specific error suppression remains intact, and persistence failures propagate.

Validation:

- Provider suite: 764 passed; the 5 native SQLite cases also passed under Electron (769 distinct tests).
- Regression coverage includes the 9-to-1 request count, retry and mixed error policies, both stale/current completion orders, nested model isolation, and persistence failures.

The request-count workload uses a deterministic HTTP fetch stub; it measures duplicate requests rather than internet latency.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model discovery reliability by sharing concurrent requests instead of issuing duplicates.
  - Ensured provider configuration changes trigger fresh discovery and prevent outdated results from being retained.
  - Preserved retry behavior after failed discovery attempts.
  - Prevented changes to returned model data from affecting cached or persisted results.

- **Tests**
  - Added coverage for concurrent request sharing, stale results, error handling, retries, data isolation, and refreshes after configuration changes.

### DIFF
--- a/src/main/provider/baseProvider.ts
+++ b/src/main/provider/baseProvider.ts
@@ -51,6 +51,7 @@ export abstract class BaseLLMProvider {
   protected isInitialized: boolean = false
   protected providerSettings: ProviderSettingsPort
   private readonly locale: ProviderLocalePort
+  private modelFetchPromise: Promise<MODEL_META[]> | null = null
 
   protected defaultHeaders: Record<string, string> = {
     'HTTP-Referer': 'https://deepchatai.cn',
@@ -113,6 +114,7 @@ export abstract class BaseLLMProvider {
 
   public updateConfig(provider: LLM_PROVIDER): void {
     this.provider = { ...provider }
+    this.modelFetchPromise = null
     this.loadCachedModels()
   }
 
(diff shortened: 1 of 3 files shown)


zerob13:

### P1 — stale in-flight request can overwrite models after updateConfig

`src/main/provider/baseProvider.ts:114-119, 265-300` resets the pending promise on config update, but an already-running discovery request can still complete and write stale models through `this.models` and `providerSettings.setProviderModels`. Promise-slot invalidation does not invalidate the result.

Please associate discovery with a generation/config snapshot and ignore stale results before mutating state or persistence. Add a regression test where the old request resolves with non-empty models after `updateConfig()` and the new request is in flight.

### P2 — shared model objects are mutable

The coalesced result returns the same `MODEL_META[]` object references to concurrent callers, while validation and downstream consumers mutate model objects. Please clone results or make the shared-result immutability contract explicit.

### Verdict
CHANGES_REQUESTED

PR #2251, opened 2026-09-07
perf(toolchains): release probe response bodies

zerob13:
Toolchain mirror probes inspect response headers but left archive bodies open. Servers that ignore the one-byte Range request could continue sending an unused archive while the actual download starts.

Cancel each successful or failed probe body as soon as headers arrive. Observe cancellation errors without awaiting body cleanup, so a stalled cancellation cannot hang URL selection or distort response-time ranking. Existing status handling, request timeouts, resumable downloads, and checksum verification remain intact.

Validation:

- Toolchain suite: 84 tests passed across 9 files.
- A real loopback server ignores Range; the regression verifies that the unused stream closes.
- URL selection completes when body cancellation never settles or rejects, and failed mirrors still fall back to the official URL.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact probing to stop downloading response bodies once the required headers are received.
  * Prevented unnecessary full archive downloads when servers ignore range requests.
  * Ensured probe response streams are closed even when mirror checks fail.
  * Prevented stalled or failed response cleanup from delaying download URL selection beyond the configured probe timeout.

### DIFF
--- a/src/main/toolchains/downloader.ts
+++ b/src/main/toolchains/downloader.ts
@@ -38,7 +38,7 @@ export async function probeArtifactUrl(
       signal
     })
     // Probes only need headers; a server may ignore Range and send the whole archive.
-    await response.body?.cancel().catch(() => undefined)
+    void response.body?.cancel().catch(() => undefined)
     if (response.status === 200 || response.status === 206) {
       successfulProbeCache.set(url, url)
       return true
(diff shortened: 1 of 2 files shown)


zerob13:

### P1 — unbounded response.body.cancel() can hang probe forever

`src/main/toolchains/downloader.ts:41` awaits `response.body?.cancel()` without a bound. A stream implementation, proxy, or test double whose `cancel()` never settles can keep `probeArtifactUrl()` pending forever. The existing timeout cannot interrupt an already pending `body.cancel()` await.

Please make cleanup non-blocking or bounded, and add a regression test for a never-settling `cancel()` implementation.

### Verdict
CHANGES_REQUESTED

PR #2252, opened 2026-09-07
perf(knowledge): abort canceled embeddings

zerob13:
Deleting a knowledge file canceled its queue task but left the active embedding request running because `KnowledgeBase` omitted the existing embedding port's `AbortSignal`. The request could keep using a network connection and provider capacity after the file disappeared.

Pass the task signal through the existing port to the provider. Completed imports, failure progress, queue cancellation, and persistence ordering retain their current behavior. Replace two private-method tests with public file-import workflows covering successful vectors, failed embeddings, and deletion during an active request.

Validation: the deletion regression fails against `dev` because no abort reaches the embedding boundary; all 18 knowledge tests pass with the fix. Repository format, i18n, lint, and both typechecks pass. This change does not alter UI layout or copy.

## Summary by CodeRabbit

- **Bug Fixes**
  - File processing now responds correctly when an in-progress operation is canceled, preventing unnecessary background work.
  - Improved handling of embedding failures during file imports, with clearer processing status and cleanup.
  - File deletion now reliably stops associated processing and removes related temporary tasks.

- **Tests**
  - Expanded coverage for successful imports, processing failures, progress updates, completion events, and cancellation during deletion.

### DIFF
--- a/src/main/knowledge/knowledgeBase.ts
+++ b/src/main/knowledge/knowledgeBase.ts
@@ -217,7 +217,8 @@ export class KnowledgeBase {
       const vectors = await this.embeddingPort.getEmbeddings(
         this.config.embedding.providerId,
         this.config.embedding.modelId,
-        [chunkMsg.content]
+        [chunkMsg.content],
+        signal
       )
 
       if (!vectors || vectors.length === 0) {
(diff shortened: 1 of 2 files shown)


zerob13:
## Review result — APPROVE

No P0-P3 issues found. The change correctly propagates the existing task AbortSignal into `getEmbeddings`, and the public workflow tests cover successful import, embedding failure, and deletion during an in-flight request. No over-design, breaking change, or excessive test scope observed. CI is green.

PR #2253, opened 2026-09-07
feat(chat): simplify tool activity display

zerob13:
Tool calls, reasoning and delegated tasks used dense bordered containers. Reply completion could hide manually opened details, and `update_plan` calls disappeared when plan execution marked them internal. This change renders compact activity rows, preserves group and child disclosure state, and presents each completed plan call as a readable historical checklist. Context compaction uses the same disclosure style for its recorded summary or error.

## Behavior

- Render manual context compaction without a synthetic assistant reply underneath it. Track the pending command per session through preparation, compaction and history refresh, and use the restored compaction state when reopening a busy session. Preserve real reply rows and busy-session controls; completion, failure and no-op paths release the local state.
- Keep title-bar controls clickable in Electron, including returning from a delegated child session to its parent. Apply the drag exclusion to the rendered buttons inside the title bar, including buttons wrapped by tooltip providers.
- Accept multiple `in_progress` steps in `update_plan` for parallel work, including delegated subagent tasks. The tool description, Agent/Minimal Mode guidance and both Code Mode frontends describe the same complete-snapshot contract. Plan displays retain every active step and only auto-collapse after all steps complete.
- Use neutral text rows and action icons, with inline task controls and indented tool details. Browser tools use a compass, including YoBrowser, legacy aliases and CUA/MCP browser commands; plan calls use a todo-list icon. Running and failure indicators retain precedence. Activity summaries use a fixed activity icon, and thinking rows use a fixed brain icon through completion, group composition changes and disclosure toggles.
- Group only consecutive runs of at least two completed visible activity blocks. Single tools, reasoning and provider search entries stay standalone, including after reply completion. Empty reasoning signatures do not count toward the minimum; MCP Apps stay outside disclosures.
- Collapse automatically opened output after success. Preserve manually opened standalone activity when the reply finishes. Expanding a child inside a completed group keeps the group header, membership and keyboard focus in place; closing and reopening the group restores each child's open or closed state, including MCP tool details.
- Keep failed, cancelled, unfinished and action-required activity outside completed groups, including settled history. Preserve already expanded failure output.
- Retain `update_plan` calls when the internal marker arrives, execution completes and history is loaded. Omit their raw JSON summary; expanding a successful call shows its own recorded checklist, completion count and explanation using the dock's status presentation. Later plan updates do not overwrite historical steps. Failed or unreadable calls retain diagnostic details, and MCP tools keep their ordinary rendering.
- Render compaction as a collapsed activity row. Expanding a successful record shows the summary from that attempt's stored anchor; failures stay in history and expand to show the error. Expanded details use a thin themed border, compact padding, 12px body text, 14px headings and tighter paragraph/list spacing at the default font setting. Typography scales with the application font size and stays scoped to compaction details. Preserve cancellation cleanup and the existing context-boundary fallback, explain unavailable or rejected summaries, and keep list height measurement working as details open and close.
- Show thinking durations below one second as "less than 1s" for completed reasoning across all 20 locales. Running labels retain their existing wording. Use the unrounded duration so 999ms stays below the one-second boundary.
- Use stable block IDs plus occurrence counts so inserted content does not rebuild identified activity rows and repeated tool-call lifecycle blocks remain distinct. Keep the existing positional fallback for legacy blocks without IDs.
- Support keyboard disclosure controls, reduced motion, long labels and narrow windows using existing theme and layout primitives. Collapsed reasoning controls omit references to unmounted content.

## UI

```text
BEFORE

Manual /compact:
  Previous assistant reply
  [loader] Compacting context...
  Model avatar + empty assistant spinner

[Back to parent] click -> child session remains

Assistant response
[ green dot | read /workspace/file | allowed > ]
+--------------------------------------------+
| Tool card                                  |
| + Parameters card -----------------------+ |
| + Response card -------------------------+ |
+--------------------------------------------+
[ Delegated task card | status badge ]
update_plan running -> removed on completion
> Thought for 0s
> Thinking for 0s...
v Worked for 42s · 1 reasoning segment
  > Thought for 28s
---------- Context compacted ----------

AFTER

Manual /compact:
  Previous assistant reply
  [loader] Compacting context...
  -> Expand the recorded summary or error when finished

[Back to parent] click -> parent session opens

Assistant response
  > [activity] Worked for 42s · 3 tool calls
  > [brain] Thought for 28s           [single; no group]
  v read /workspace/file              [single; manual]
    Parameters                         Copy
    Response                           Copy
  v exec pnpm lint · Failed
    Error output remains visible
    Task title · Running         Open Stop >

  [compass] load_url https://example.com
  [compass] cdp_send Page.captureScreenshot
  v [activity] Completed activity     [stays visible]
    v [list-todo] update_plan
      Plan · 1/4 completed
      [x] Inspect code
      [~] Adjust layout
      [~] Review UI with a subagent
      [ ] Verify screenshots
    > read /workspace/file
  > [brain] Thought for less than 1s
  > [brain] Thinking for 0s...

  [loader] Compacting context...
  v [fold-vertical] Context compacted
    +---------------------------------------+
    | Small heading                         |
    | Compact recorded summary and lists    |
    +---------------------------------------+
  v [alert] Compaction failed
    +---------------------------------------+
    | Provider error details                |
    +---------------------------------------+
```

## Validation

- Passed 397 main-process checks for the harness and compaction coordinator, plus 266 renderer checks for the chat page, message list, composer, message projection and session/agent stores. Compaction retains summaries and errors alongside plugin hook dispatch, and failed hooks do not interrupt final state projection. Inspected the actual Vue row and Markdown renderer in an isolated preview with light/dark themes, narrow columns, 100%/125% application font settings and failure details. Verified compact typography, a 1px themed border, wrapped paths, keyboard disclosure and unchanged typography in the adjacent normal message.
- Passed 138 renderer checks for the chat page, composer submission and display-message projection. Deferred compaction verifies that preparation, progress and settlement do not append an empty reply, while normal reply placeholders return afterward. Failure and no-op checks cover session switching; restored compaction state preserves real assistant rows.
- Executed `/compact` in the running macOS app and inspected the resulting screenshot. That session had no eligible history and displayed the no-op notice without a new assistant row; delayed and failed requests are covered by the controlled regression checks.
- Passed 255 relevant checks for the top bar, chat page, session store, message list and harness compaction state. Compaction assertions verify disabled disclosure while running and a persisted error when summary generation fails, while retaining the context-boundary fallback checks.
- Reproduced the ignored return-to-parent click in the running macOS Electron app, then verified the same native mouse click opens the parent after the CSS fix. Inspected screenshots before and after navigation; title-bar and chat layout remain intact.
- Passed the relevant main-process and renderer suites, including 170 checks for compaction, transcript persistence, message projection, disclosure behavior and icon availability. Coverage verifies historical compaction summaries, persisted failures, cancellation and runtime replacement, the two-block grouping threshold, empty reasoning filtering, standalone MCP App identity, parallel plan submission and progress snapshots, tool-mode guidance, multiple active steps in current and historical plans, interruption presentation, completion-driven collapse, group membership, restored child disclosure, keyboard focus and MCP App identity.
- Passed `pnpm run format:check`, `pnpm run i18n`, `pnpm run lint`, `pnpm run typecheck`, `pnpm run architecture:renderer-baseline:check` and `pnpm run icons:check`. Generated icon collections match current renderer references, and the icon whitelist test passes.
- Captured and inspected Electron screenshots using isolated user data and fixture messages: light and dark themes, 760–800px narrow windows, long paths, expanded details, task controls, successful completion and retained failure output. Verified keyboard expansion and retained details after inserting content ahead of a tool call. Activity and thinking icons remain distinct, stable and 16px across completion, group composition changes and group reopening. Single tools and reasoning retain their DOM identity through completion, keyboard disclosure works, and automatically opened tool output still collapses after success.
- Verified that manual expansion retains its DOM node across completion, pending review stays outside collapsed groups, the composer keeps its position and dimensions, and the checked views have no horizontal overflow or renderer page errors.
- Verified `update_plan` retains its row when marked internal and completed, remains inspectable after the live plan float disappears, and appears in a freshly loaded history view. Electron captures show distinct 1/3 and 3/3 historical checklists, unchanged after the dock receives a new plan. Group headers remain in place and child details survive group collapse and reopening.
- Captured compaction progress, independent historical summaries, failure details and summary-rejection explanations in light mode and a 760px dark window. Verified keyboard disclosure, stable triggers on completion, updated row measurements after expansion and collapse, unchanged adjacent tool geometry and no horizontal overflow or renderer errors.
- Captured and inspected Chinese and English thinking labels in light mode and German labels in an 800px dark window. Confirmed no label truncation or horizontal overflow, working manual expansion, unchanged composer dimensions across languages and no renderer page errors.

## Summary by CodeRabbit

- **New Features**
  - Activity, thinking, and tool-call sections support clearer manual expand/collapse behavior.
  - Plan tool calls display accessible checklist summaries with progress and completion details.
  - Activity controls include improved accessibility, focus handling, responsive icons, and reduced-motion support.
  - Very short thinking durations are labeled as less than one second.

- **Bug Fixes**
  - Expanded sections remain open after activity updates and regrouping.
  - Unfinished, failed, or action-required activity remains visible.
  - Plan-related tool calls remain visible throughout progress and history reloads.

- **Localization**
  - Added translations for the new duration label across supported languages.


zhangmo8 (commented):
Reviewed the full diff. Nice work overall — the flat activity-row model is a clear improvement over the bordered-card layout, and the details are handled well:

- The `hasManualToggle` guard in `MessageBlockThink` correctly handles the race between the async `think_collapse` read and an early user click, and there's a test for it.
- Preserving manual expansion via `expandedBlockKeys` is a clean way to keep blocks standalone across the pending → grouped transition without remounting; the DOM-identity tests prove the intent.
- Good coverage of reduced motion, focus-visible rings, `aria-expanded`/`aria-controls`, and the sub-second thinking label landed in all 20 locales.
- The existing CodeRabbit comments (missing `manual-toggle` forwarding in `MessageBlockActivityGroup`, invalid `cancelled` fixture status) already cover the issues I found there too, so I won't repeat them.

A few additional notes, none blocking:

1. `buildBlockKey` embeds the block's array index (`${messageId}:${stableId}:${index}`), and `manuallyExpandedBlockKeys` persists those keys across renders. As long as blocks are append-only during streaming this is fine, but if a block is ever inserted before an already-expanded one, indices shift and the expanded block silently falls back into the completed group. Consider keying on the stable id alone, or documenting the append-only assumption.

2. In `ThinkContent.vue`, the trigger's `aria-controls` references the content div, but the content is removed with `v-if` while collapsed, so the attribute points at a non-existent id in the collapsed state. Minor, but keeping the element mounted (hidden) or dropping `aria-controls` while collapsed would be more correct.

3. `isCompletedActivityBlock` now excludes anything that isn't `status === 'success'`, so `error` and `cancelled` blocks always render standalone — including in older history messages. Right call for live sessions, but long histories with several cancelled calls will show a fairly long ungrouped list. Is that intended, or should terminal-but-unsuccessful blocks eventually group as well?

CI (build/static/test-main/test-renderer) is green at d455aa1. Happy to approve once the forwarded `manual-toggle` fix lands.

zerob13:

### P1 — grouped activity rows drop manual-toggle events

`src/renderer/src/components/message/MessageBlockActivityGroup.vue:46-65` renders children that emit `manual-toggle`, but the group listens to/re-emits only `toggle-collapse`. `MessageItemAssistant` therefore cannot record the child key for completed grouped activity, so manual expansion can be lost on regroup or rerender. Forward the exact standalone key, including the `:tool` suffix for MCP-app tool calls.

### P2 — invalid cancellation status in tests

`test/renderer/components/message/messageActivityGroups.test.ts:397-403` and `test/renderer/components/message/MessageBlockToolCall.test.ts:779-783` use `cancelled`, while production supports `cancel` (`src/renderer/src/features/chat-page/model/displayMessage.ts:144-151`).

### Verdict
CHANGES_REQUESTED

zhangmo8 (commented):
Reviewed the delta since d455aa1 (`5824527`..`8be4e1c` plus the `1808d98` test fixes). All three items from my earlier review are addressed — occurrence-count-based block keys, the conditional `aria-controls` in `ThinkContent`, and the grouped `manual-toggle` forwarding (including the `:tool` suffix for MCP tool calls) now line up with `buildAssistantRenderItems` and the standalone keys. Nice.

A few new observations on the delta, none blocking:

1. **Compaction is marked "failed" even when the boundary was committed.** In `compactionRuntimeCoordinator.ts` (~line 423), when summary generation fails, `compactionService` still calls `commitBoundaryOnly` and the anchor **is** committed — the context is successfully compacted, just without a summary. The coordinator then writes status `'failed'` (and the transcript row gets status `'error'`), so the UI shows "Compaction failed" for a compaction that actually succeeded. The existing `summary_unavailable` boundary reason and its "compacted without summary" label already express this exact state. Consider keeping status `'compacted'` with the boundary reason, and surfacing the summary error as secondary detail inside the expanded view — reserving `'failed'` for cases where no anchor was committed.

2. **Raw provider error messages are persisted unredacted.** The catch path stores `error.message` verbatim into `compactionError` metadata and the error block content, while the logging path in `compactionService` deliberately runs through `redactRuntimeErrorForLog`. Provider errors can contain base URLs or request internals, and this text lands in the persisted transcript, message search, and export. Worth applying the same redaction before persisting.

3. **`materializeCompactionMetadata` adds per-row work to the hot load path** (`transcript.ts` ~1049). It now JSON-parses every row's metadata and, for each compacted message, performs a synchronous anchor lookup during list materialization — an N+1 on long histories with many compaction messages. A per-attempt-id cache or a batched lookup (e.g., resolving anchors once per materialization) would keep this flat.

4. **Manual collapse state of a standalone think block is lost if it later joins a group.** `manuallyExpandedBlockKeys` only tracks expanded keys, so a user-collapsed standalone think block that subsequently becomes part of a completed activity group remounts with no `activityExpansion` entry and falls back to the global preference. Edge case, but slightly asymmetric with the "preserve manually opened standalone activity" behavior the PR advertises.

CI note: test-main/test-renderer were red at `8be4e1c` (harness compaction status expectations, and the removed `.compaction-divider__label--compacting` selector in `MessageList.test.ts`). `1808d98` updates both tests to the new behavior — test-main/static/build are already green on the new run, test-renderer was still in progress when I posted this.

PR #2256, opened 2026-09-07
fix(data): restore partial legacy schemas

zerob13:
Partial legacy databases could fail before schema repair became available. `deepchat_sessions` compared the database's global version (for example 43 or 67) with that table's last migration (31), incorrectly treating supported databases as downgrades. A second path installed the delegation child-binding trigger before missing legacy Session columns were repaired, causing later migration statements to fail.

Check downgrade compatibility once against the complete Main schema catalog before creating application tables. Use the existing prerequisite-aware delegation trigger installation during migration, then restore its constraints in the Session schema repair callback. No schema version bump or destructive recovery is introduced.

Update legacy tests to use the current SessionDatabase/AgentDatabase owners. Replace the mocked per-table version guard with native coverage for actual future databases, both with and without a missing table. Retain legacy row/revision/usage recovery checks and verify repaired databases reject an invalid delegation child.

Validation: six native regressions fail on `dev`; all 536 tests across the 38 affected database/startup/session suites pass with this change. Format, i18n, lint, and both typechecks pass. Existing destructive-recovery isolation remains covered.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database startup safeguards by preventing unsupported future schemas from being opened or downgraded.
  - Improved recovery for session data by restoring related live-delegation records when required tables are repaired.
  - Corrected migration behavior so live-delegation tables and their required safeguards are fully established.
  - Added validation to prevent live-delegation records from referencing invalid sessions.


zerob13:
## Review result — APPROVE

No P0-P3 issues found. Global schema-version validation, trigger installation prerequisites, idempotent repair, and test ownership boundaries are coherent. No over-design, breaking change, or excessive test scope observed. CI is green.

PR #2257, opened 2026-09-07
test(renderer): reuse session store fixtures

zerob13:
The session store suite reloaded its Vue dependency graph for each fixture. Its 91 existing cases exhausted the default roughly 4 GB Node heap and aborted the renderer suite. Reuse hoisted mock factories with fresh per-fixture return values, keep runtime identity at its existing asynchronous boundary, and dispose each store's Vue effect scope after the test.

All 91 behavioral cases are retained. The isolated suite now finishes in 1.94 seconds with a reported final heap of 149 MB and maximum resident memory of about 306 MiB. No heap-limit increase or production scheduling change is needed.

Validation: the complete renderer suite passes (268 files, 2,377 tests). Format, i18n, lint, and both typechecks pass. The measurements describe test-runner resource use on macOS arm64, not application memory.

### DIFF
--- a/test/renderer/stores/sessionStore.test.ts
+++ b/test/renderer/stores/sessionStore.test.ts
@@ -1,9 +1,42 @@
-import { reactive } from 'vue'
+import { effectScope, reactive, type EffectScope } from 'vue'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
   GUIDED_ONBOARDING_RESUME_REQUESTED_EVENT,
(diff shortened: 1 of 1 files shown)


zerob13:
## Review result — APPROVE

No P0-P3 issues found. The test-only mock/lifecycle refactor uses file-level mocks and effect scopes consistent with production cleanup, preserves the 91 behavior tests, and fixes the fixture dependency graph without production changes. Focused test passed: 91 tests.

PR #2258, opened 2026-09-07
fix(migration): preserve session recency

zerob13:
The disabled-tool startup migration changed `updated_at` while normalizing legacy tool names. An older affected conversation could jump ahead of a newer conversation, and the existing recency indexes were rewritten even though no user activity occurred.

Update only the disabled-tool payload and revision inside the existing transaction. Preserve conversation timestamps and environment recency, keep both normalized stores atomic, and leave already-completed migrations alone.

Repair the native fixture by creating the current active-Skills relation before inserting sessions. Keep the original order/recency and rollback assertions, verify revision advancement and rollback, and remove the mocked assertion that copied the faulty SQL verbatim. The repaired native test fails on the baseline because the older session is moved first.

Validation: all 14 focused migration tests pass in the Electron-compatible SQLite runtime. Format, i18n, lint, and node/web typechecks pass. No new schema version, dependency, or UI layout change.

Before / after (conversation order after startup migration):

```text
BEFORE                         AFTER
Older chat (migration time)     Newer chat (200)
Newer chat (200)                Older chat (100)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Session data migrations now preserve existing session update timestamps when normalizing disabled tools.
  * Migration revision tracking is now handled consistently for migrated, unchanged, and rolled-back sessions.
  * Improved migration reliability across supported session database scenarios, including rollback handling.

### DIFF
--- a/src/main/app/startupMigrations/sessionDataMigrations.ts
+++ b/src/main/app/startupMigrations/sessionDataMigrations.ts
@@ -442,16 +442,12 @@ export async function runDisabledAgentToolCapabilityCleanupMigration(
        ORDER BY id ASC
        LIMIT ?`
     )
-    const updateSessionDisabledTools = db.prepare<[string, number, string]>(
-      'UPDATE new_sessions SET disabled_agent_tools = ?, updated_at = ?, revision = revision + 1 WHERE id = ?'
+    const updateSessionDisabledTools = db.prepare<[string, string]>(
+      'UPDATE new_sessions SET disabled_agent_tools = ?, revision = revision + 1 WHERE id = ?'
     )
     const persistSessionDisabledTools = db.transaction(
       (sessionId: string, disabledAgentTools: string[]): boolean => {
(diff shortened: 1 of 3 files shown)


zerob13:

### P1 — stale in-flight request can overwrite models after updateConfig

`src/main/provider/baseProvider.ts:114-119, 265-300` resets the pending promise on config update, but an already-running discovery request can still complete and write stale models through `this.models` and `providerSettings.setProviderModels`. Promise-slot invalidation does not invalidate the result.

Please associate discovery with a generation/config snapshot and ignore stale results before mutating state or persistence. Add a regression test where the old request resolves with non-empty models after `updateConfig()` and the new request is in flight.

### P2 — shared model objects are mutable

The coalesced result returns the same `MODEL_META[]` object references to concurrent callers, while validation and downstream consumers mutate model objects. Please clone results or make the shared-result immutability contract explicit.

### Verdict
CHANGES_REQUESTED

PR #2259, opened 2026-09-07
test(e2e): repair streaming state coverage

zerob13:
The generation-completion helper read `data-generating` from `chat-page`, while the attribute belongs to `chat-page-shell`. Read the current owner and scope the send button and composer to that shell.

A local HTTP provider holds a Markdown response open, allows editing the active composer during generation, then verifies the final response and preserved next-message draft. Attach a rejection observer immediately when starting the completion wait, while retaining its final await so completion failures still fail the test. This prevents an intermediate assertion failure from leaving an unhandled rejection.

Refresh the E2E guide with fixture-owned temporary profiles, live-provider gates, current model/environment defaults, packaged mode, and artifact locations.

Validation:

- The local streaming Electron workflow passes without provider credentials.
- The combined performance-audit production build passed the full smoke suite: 33 passed, 5 existing live-provider skips.

## Summary by CodeRabbit

- **Documentation**
  - Updated end-to-end testing guidance with temporary profiles, seeded onboarding and diagnostics, supported commands, packaged-app configuration, CI coverage, artifact retention, and chat state expectations.
  - Documented local provider-integration requirements and expanded default test coverage.

- **Tests**
  - Added coverage for streaming chat responses, including editing a draft while generation is in progress.
  - Expanded validation of local provider and model integrations, streamed completions, final responses, and error-free chat behavior.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -1,71 +1,67 @@
 # DeepChat E2E Smoke
 
-This suite runs manual smoke regression against the real local desktop environment by default.
+The suite launches the built Electron application. By default, each fixture creates a temporary
+`userData` profile, seeds completed onboarding and diagnostic logging, and removes the profile
+after closing the app. It does not use your normal desktop profile by default.
 
-It does not use mock providers, alternate `userData` directories, or E2E-only bootstrap state.
-The default `pnpm run e2e:smoke` command runs against the same local profile that the app normally
-uses.
+## Coverage
 
-## Scope
+The default suite covers startup, Settings navigation, typed IPC boundaries, workspace watchers,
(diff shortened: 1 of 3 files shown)


zerob13:

### P2 — completion rejection may become unhandled before it is awaited

`test/e2e/specs/36-chat-streaming.smoke.spec.ts:108` creates `const completion = waitForGenerationDone(app.page)` and performs further actions before awaiting it. If the helper rejects during that interval, Node/Playwright may report an unhandled rejection and fail the worker with misleading timing. Attach a rejection handler immediately while preserving the original promise for the final await.

### Verdict
CHANGES_REQUESTED

PR #2261, opened 2026-09-07
fix(browser): restore macOS preview capture

zerob13:
Creating a YoBrowser preview host on macOS can leave the page without a capturable compositor surface. Preview frames repeatedly fail with `Current display surface not available for capture`, and CDP screenshots can fail or stall. Calling `setBackgroundThrottling(false)` while the host is hidden triggers [Electron #52844](https://github.com/electron/electron/pull/52844).

On macOS, leave the existing throttling setting unchanged and activate the preview host with `showInactive()`. The shown host continues producing frames. On other platforms, disable throttling only if it was enabled. Both explicit release and host-close cleanup restore it only when preview setup changed it.

Validation:

- Native Electron 41.10.4 on macOS, using the repository's preview-host creation and disposal methods: the original implementation fails its initial capture and retry; the updated implementation captures successfully on initial load and after 10 consecutive host reopenings, through both `capturePage` and `Page.captureScreenshot`.
- Background page liveness: 120 animation frames and 20 callbacks from a 100 ms interval over 2 seconds. Captured images inspected visually.
- 87 tests passed across YoBrowserPresenter, BrowserTab, YoBrowserToolHandler, agentToolManagerYoBrowser, and AgentPreviewCoordinator.
- Parameterized preview lifecycle tests cover macOS, Windows, and Linux platform branches with initially enabled or disabled throttling, through explicit release and host-close cleanup.
- Format, i18n, lint, and both typechecks passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preview behavior across platforms by preserving the existing background-throttling setting.
  * Prevented unnecessary changes to background throttling when previews are already configured without throttling.
  * Ensured background throttling is restored correctly when preview hosts are closed or released.

### DIFF
--- a/src/main/desktop/browser/YoBrowserPresenter.ts
+++ b/src/main/desktop/browser/YoBrowserPresenter.ts
@@ -1495,9 +1495,12 @@ export class YoBrowserPresenter implements IYoBrowserPresenter {
       host.contentView.addChildView(state.view)
       state.view.setBounds({ x: 0, y: 0, ...PREVIEW_VIEWPORT })
       state.view.setVisible(true)
-      state.page.contents.setBackgroundThrottling(false)
       if (isMac) {
+        // The shown host keeps frames active. Unthrottling while hidden breaks capture on macOS.
+        // https://github.com/electron/electron/pull/52844
         host.showInactive()
+      } else {
+        state.page.contents.setBackgroundThrottling(false)
       }
       state.previewHost = host
       host.once('closed', () => {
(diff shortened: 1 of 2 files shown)


zhangmo8 (approved):
Verified this fix end-to-end:

- The linked upstream change is real: electron/electron#52844 ("fix: blank WebContentsView after setBackgroundThrottling(false) while hidden") matches the symptom described in the comment, so skipping the unthrottle on macOS is the right call.
- `backgroundThrottling` is a proper boolean property on WebContents in Electron 41, so the cleanup guard is type-safe and effective — it also has the nice side effect of preserving a deliberately disabled throttling state set by other features.
- The mock now models state transitions (`backgroundThrottling` updated by the setter) instead of asserting raw call arguments, which makes the cleanup test meaningfully exercise the guard.
- CI (build/static/test-main/test-native-memory/test-renderer) is green.

(CodeRabbit's suggestion to track and restore the exact pre-preview throttling state is a reasonable hardening follow-up; nothing here blocks.)

PR #2267, opened 2026-09-08
perf(build): speed up typecheck (TNB bridge.16, incremental)

zhangmo8:
## Summary

Full `pnpm run typecheck` was taking ~2.8 minutes. Root-cause investigation (see measurements below) showed the tsgo-backed TNB checker is working correctly, but every run was a full rebuild with no cache, the TNB fork was 9 bridge versions behind, and `tsconfig.node.json` carried an unused wildcard path mapping.

### Changes

- **Upgrade `typescript` override**: `typescript-native-bridge` `6.0.3-bridge.7` → `6.0.3-bridge.16` (latest; includes multiple resolution/correctness fixes, e.g. virtualization of `.vue` sources under node_modules)
- **Remove dead wildcard mapping** `"*": ["./*"]` from `tsconfig.node.json` — no import in the repo relies on it; it forced an extra resolution attempt per import
- **Incremental typecheck**: `typecheck:node` / `typecheck:web` now pass `--incremental --tsBuildInfoFile node_modules/.cache/...` so warm runs reuse the prior check state (cache lives under `node_modules`, never committed)
- **Untrack stale** `tsconfig.node.tsbuildinfo` (committed months ago, unused since `--composite false` disables it) and ignore `*.tsbuildinfo`

### Measurements (node project: 2525 files; web project: ~878 source files)

| | before | after (cold) | after (warm) |
|---|---|---|---|
| `typecheck:node` | ~51s | **18.9s** | **3.3s** |
| `typecheck:web` | ~115s | **30.8s** | **16.4s** |
| full `typecheck` | ~166s | **~50s** | **~20s** |

0 new/removed diagnostics on both projects; `oxfmt --check` passes.

## Test plan

- [x] `pnpm install` resolves `typescript-native-bridge@6.0.3-bridge.16` (lockfile updated)
- [x] `pnpm run typecheck` passes with 0 errors on both projects
- [x] Second run confirms incremental cache hits (node 3.3s, web 16.4s)
- [x] `pnpm run format:check` passes

## Summary by CodeRabbit

- **Chores**
  - Improved development build and type-check performance through incremental checking and caching.
  - Updated TypeScript tooling to support the latest incremental compilation improvements.
  - Refined project configuration to use explicit module path aliases, improving consistency and predictability.
  - Added build metadata to ignored files so local development artifacts do not appear as project changes.


zerob13 (dismissed):
## Review verdict

REQUEST_CHANGES — 1 P1, 1 P2.

### P1 — `typecheck:web` incremental cache can become stale across branch/worktree changes

**Location:** `package.json:40`

The new command enables TypeScript incremental compilation and stores the build info under `node_modules/.cache/tsc-web.tsbuildinfo`. The cache is intentionally outside Git and is therefore shared by all branches that use the same checkout/worktree `node_modules`. TypeScript's incremental program can reuse prior graph/state based on file metadata and does not provide a repository/branch boundary. Switching branches or rebasing while timestamps/content combinations are reused can make typecheck accept stale diagnostics or produce results unrelated to the checked-out source.

Please either make the cache identity branch/commit-specific, or explicitly invalidate/recreate the build-info file when the source/config dependency identity changes. Add a reproducible regression check for switching between incompatible revisions if this cache is part of the supported workflow.

### P2 — `composite` is disabled while the config still declares project references

**Location:** `package.json:39-40`, `tsconfig.node.json:11`

The scripts continue to pass `--composite false`, while `tsconfig.node.json` still sets `compilerOptions.composite: true`. This makes the incremental behavior depend on command-line overrides and weakens the project-reference/build-info contract without explaining why. Please either keep the existing non-composite behavior and document the intended cache semantics, or align the config/scripts around one explicit incremental strategy and verify the project-reference behavior.

### Verification

`pnpm run typecheck:node` could not execute in this checkout because `node_modules` is absent (`tsc: command not found`); the checkout remained read-only.

zerob13 (approved):
## Review verdict

APPROVE

The build-only change is narrowly scoped and consistent with the repository. Incremental typecheck state is explicitly stored under `node_modules/.cache`, stale tracked `tsbuildinfo` is removed, and the TypeScript native bridge upgrade is synchronized in the lockfile. Existing aliases remain intact, with no runtime/API changes or unnecessary tests.

Verification reported by review: `pnpm install --frozen-lockfile --ignore-scripts`, `pnpm run typecheck:node`, and `pnpm run typecheck:web` pass; warm checks are materially faster. Full CI-runtime confidence remains subject to the repository's Node 24 / platform matrix.

PR #2269, opened 2026-09-08
feat(provider): add API Route

DennyHo0917:
## Summary

- add API Route as a disabled built-in OpenAI-compatible provider
- discover available models from the provider's `/models` endpoint instead of hardcoding a catalog
- add the API Route brand icon to the desktop provider list
- disable embeddings because the documented integration currently exposes chat completions and models

## UI

Before:

```text
Provider list
└── API Route is not available
```

After:

```text
Provider list
└── API Route [brand icon]
    ├── API URL: https://global.api-route.com/v1
    ├── API Key
    └── Refresh models
```

## Validation

- `vitest` provider defaults and registry tests: 31 passed
- `vitest` ModelIcon tests: 6 passed
- Node and web typechecks passed
- lint guards and oxlint passed
- i18n validation passed (20 locales)
- icon collection check passed
- changed-file formatting check passed

The full-repository formatting check reports existing Windows line-ending differences across the checkout; no unrelated files were reformatted.

## Summary by CodeRabbit

- **New Features**
  - Added API Route as a built-in OpenAI-compatible provider option.
  - Included API Route branding and provider details in the provider selection interface.
  - The provider is currently disabled and unavailable for use by default.

- **Tests**
  - Added coverage to verify API Route provider configuration, credentials, model discovery, and default settings.

### DIFF
--- a/src/main/provider/defaults.ts
+++ b/src/main/provider/defaults.ts
@@ -32,6 +32,21 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
     }
   },
   {
+    id: 'api-route',
+    name: 'API Route',
+    apiType: 'openai-completions',
+    apiKey: '',
+    baseUrl: 'https://global.api-route.com/v1',
+    enable: false,
+    websites: {
+      official: 'https://www.api-route.com/',
+      apiKey: 'https://www.api-route.com/api-keys',
+      docs: 'https://www.api-route.com/docs/quickstart',
+      models: 'https://www.api-route.com/pricing',
+      defaultBaseUrl: 'https://global.api-route.com/v1'
+    }
+  },
+  {
     id: 'runinfra',
     name: 'RunInfra',
     apiType: 'openai-completions',
--- a/src/main/provider/providerRegistry.ts
+++ b/src/main/provider/providerRegistry.ts
@@ -132,6 +132,14 @@ const PROVIDER_ID_REGISTRY = new Map<string, AiSdkProviderDefinition>([
     })
   ],
   [
+    'api-route',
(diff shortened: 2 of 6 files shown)


zhangmo8 (commented):
Reviewed the diff and ran the provider test suites locally (both `defaultProviders.test.ts` and `basicApiKeyProviders.test.ts` pass, 31/31). The change follows the existing provider conventions correctly: disabled by default, `apiKey: ''`, standard `websites` block, `OPENAI_BASE` spread with `credentialStrategy: 'api-key'`. The default endpoint and docs links all resolve (the 401 on `/v1/models` without auth is expected). Tests cover both the registry resolution and the defaults entry. Left-aligned with `apimart`/`aihubmix` etc. — no blockers.

Two non-blocking notes:

1. **Embedding strategy**: this declares `embeddingStrategy: 'none'`, while other OpenAI-compatible aggregators (e.g. `aihubmix`, `qiniu`) keep the default `'openai'`. If API Route does expose `/v1/embeddings`, keeping `'openai'` would give users embedding support for free; if it doesn't, the current choice is right. Worth confirming against their docs.

2. **`websites.models` points to the pricing page** (`https://www.api-route.com/pricing`). Other providers point this entry at an actual model-list endpoint/page (see `apimart`, `qiniu`). Minor, but pointing it at the real models list would be more consistent.

DennyHo0917:
Thanks for the review. I confirmed that API Route currently documents Chat Completions and Models, but not an embeddings endpoint, so `embeddingStrategy: 'none'` is intentional. The pricing page is also the public, up-to-date model catalog users can browse; the authenticated `/v1/models` endpoint remains the runtime source of truth.

PR #2270, opened 2026-09-08
perf(renderer): memoize artifact extraction per block reference

zhangmo8:
## Summary

During streaming, every chunk replaces the streaming record in `messageCache`, which invalidates `WorkspacePanel`'s `artifactItems` computed. The recompute re-ran `extractArtifactsFromContent` (regex scan) over **every assistant block in the whole conversation** per chunk — cost grows linearly with conversation length.

This memoizes extraction per block object reference (`WeakMap<DisplayAssistantMessageBlock, readonly ParsedArtifactPart[]>`). The message store reuses block references for settled blocks (`reuseStableAssistantBlocks`), so a per-chunk rescan only re-extracts blocks whose reference actually changed (typically the streaming tail). Status changes always yield a new block object, so status is covered by the key. Cached arrays are frozen to protect the shared-cache contract.

Per-chunk cost drops from O(all blocks × content length) to O(1) lookups + one re-extract of the streaming tail. Reactive contract is unchanged — streaming artifacts still appear live.

## Test plan

- [x] `pnpm vitest run test/renderer/composables/useArtifacts.test.ts` (new memo-contract case)
- [x] `pnpm vitest run test/renderer/components/WorkspacePanel.test.ts test/renderer/components/ChatSidePanel.test.ts`
- [x] `pnpm typecheck:web`, `oxfmt --check`, `oxlint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact detection in the workspace panel, including while content is still streaming.
  * Ensured artifact listings stay consistent and update correctly when message content changes.
  * Improved responsiveness by reusing results for unchanged message blocks.
* **Reliability**
  * Prevented accidental changes to extracted artifact results, helping keep displayed information stable.

### DIFF
--- a/src/renderer/src/components/sidepanel/WorkspacePanel.vue
+++ b/src/renderer/src/components/sidepanel/WorkspacePanel.vue
@@ -196,7 +196,7 @@ import { useI18n } from 'vue-i18n'
 import { createFileClient } from '@api/FileClient'
 import { createProjectClient } from '@api/ProjectClient'
 import { createWorkspaceClient } from '@api/WorkspaceClient'
-import { extractArtifactsFromContent } from '@/composables/useArtifacts'
+import { extractArtifactsFromBlock } from '@/composables/useArtifacts'
 import WorkspaceFileNode from '@/components/workspace/WorkspaceFileNode.vue'
 import LiveDelegationPanel from './LiveDelegationPanel.vue'
 import WorkspaceViewer from './WorkspaceViewer.vue'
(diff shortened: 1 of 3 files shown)


zerob13 (approved):
## Review verdict

✅ Approve

I found no blocking correctness, compatibility, or over-design issues.

The `WeakMap` cache is correctly keyed by the stable assistant block references produced by the message store. Streaming updates replace changed block objects, so changed artifact content is re-extracted while settled blocks reuse cached results. Freezing the cached arrays also protects the shared-result contract.

The focused artifact extraction tests pass: 6/6.

PR #2271, opened 2026-09-08
perf(renderer): fix message row render bailout

zhangmo8:
## Summary

During streaming, **every mounted message row re-rendered on every chunk** — settled rows never bailed out of updates. Measured in the real renderer (200 settled rows, windowed; 50 synthetic stream chunks driven through the real message store):

| | before | after |
|---|---|---|
| reactive render+patch per chunk | avg 33.5ms, p50 17.2ms, **p95 128ms**, max 351ms | avg 22.4ms, p50 16.8ms, **p95 44.3ms**, max 199ms |
| `MessageListRow` patches | ~15 per chunk (all mounted rows) | ~0 (rows exit the top patchers entirely) |
| `MessageList` patch avg | 78ms | 4.6ms |

### Root cause

Two dynamic props on `MessageListRow` were recreated on every `MessageList` render, so Vue's `shouldUpdateComponent` props comparison always found a difference and re-rendered each row's whole subtree (assistant body, tooltips, toolbar):

1. `@animationend="onEntranceAnimationEnd(item, $event)"` — inline closure capturing the `v-for` `item` cannot be cached by the compiler; the handler prop gets a fresh function identity per render.
2. `:class="{ 'message-row-entrance': ... }"` — object literal identity differs per render.

### Fix

- `@animationend="onRowAnimationEnd"` — stable top-level handler; the row root already carries `data-message-id`, so the listener resolves the message from `event.currentTarget`.
- `:class="entranceClassFor(item)"` — stable string (`''` or `'message-row-entrance'`), equal across renders.

Entrance-animation behavior is unchanged (class still applies while animating; removed on `animationend`).

## Test plan

- [x] `pnpm vitest run test/renderer/components/MessageList.test.ts test/renderer/components/MessageListRow.test.ts test/renderer/components/ChatPage.test.ts` (121 passed)
- [x] `pnpm typecheck:web`, `oxfmt --check`, `oxlint`
- [x] Live-renderer A/B measurement described above (synthetic stream probe against the real message store)

## Summary by CodeRabbit

* **Performance**
  * Improved chat message list rendering during streaming by reducing unnecessary updates to already-mounted message rows.
  * Stabilized message row animations and event handling across parent re-renders.

### DIFF
--- a/src/renderer/src/components/chat/MessageList.vue
+++ b/src/renderer/src/components/chat/MessageList.vue
@@ -18,9 +18,9 @@
         :is-read-only="isReadOnly"
         :allow-guard-stop-continue="item.id === latestAssistantMessageId"
         :disable-markdown-virtualization="shouldDisableMarkdownVirtualization"
-        :class="{ 'message-row-entrance': shouldAnimateEntrance(item) }"
+        :class="entranceClassFor(item)"
         :data-entrance-feedback="shouldAnimateEntrance(item) || undefined"
-        @animationend="onEntranceAnimationEnd(item, $event)"
+        @animationend="onRowAnimationEnd"
         @retry="onRetry"
         @delete="onDelete"
         @fork="onFork"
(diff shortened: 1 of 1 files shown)


zerob13:
## Review result: APPROVE

I found no actionable P0–P3 issues.

### Scope

- Replaces per-row dynamic class objects with a stable class string.
- Replaces per-row `animationend` closures with a stable handler and resolves the row ID from the existing `data-message-id` root attribute.
- The target/currentTarget guard remains in place, so descendant animation events are not incorrectly handled by the row.

The change is small, directly addresses Vue render-bailout performance, matches the existing Vue 3/TypeScript style, and does not add over-engineered abstractions or excessive tests.

### Verification

- `git diff --check origin/dev...fe2d900d03f2aa3ab6b9ced7032ba2e9433259c8` passed.
- PR CI passed: build, static, renderer tests, main tests, native-memory tests, package-required, pr-required, and CodeRabbit.
- Local dependency-based checks were unavailable in the isolated worktree (`node_modules` absent; repository requires Node >=24.18.0 <25).

zerob13 (approved):
Reviewed at fe2d900d03f2aa3ab6b9ced7032ba2e9433259c8. No actionable P0-P3 findings; see the detailed review comment.

PR #2274, opened 2026-09-09
refactor(memory): unify write paths, stage recall, extract breaker and maintenance passes

yyhhyyyyyy:
## Summary

This PR restructures the write, recall and maintenance services of the memory system without changing any schema, migration, Tape fact or anchor name. `rememberMemory` and extraction now share one decision kernel instead of two implementations that had drifted apart; `retrieve` is split into keyword, vector, refill and prune stages over one per-request state; the heavy maintenance steps are an ordered pass table under a scheduler that only owns cooldown, fence, concurrency, budget and audit; the query-embedding breaker is its own class; claim ownership is classified in one place; and three `MemoryService` self-loops in the wiring are gone. Two defects found while reviewing the result are fixed with regression tests. The behavior changes that fall out of the merge are listed below.

## What was wrong

**Two write paths had drifted.** `coordinateWriteAttempt` (single remember) and `coordinateBatchWrites` (extraction) implemented the same provenance, tombstone, recall, decision and apply sequence separately. Only the single path checked the tombstone before recall, so a forgotten candidate in an extraction batch still paid for neighbor recall and a decision call. The single path retried a CAS loss by re-running recall and the decision model; the batch path reused its query vectors. The single path recalled neighbors through the injection path, so the warm-recall breaker that exists to protect first-token latency also gated background writes, and every remember ran the adaptive refill loop for three neighbors.

**Six sites composed the same ownership judgment.** `resolveProvenance`, `handleProvenanceHit`, `supersedeHead` and the challenged/live head checks were spelled out inline six times with different flag combinations, and the response to each state (restore an archived owner, fold temporal metadata into a duplicate, leave suppressed and challenged chains alone) was repeated four more times.

**`retrieve` was one 450-line function.** Keyword search, the breaker-guarded vector stage, the adaptive refill loop with authoritative revalidation, inline vector pruning, fusion and the cancelled/failed classification shared closures and repeated the fence-plus-read-epoch check six times. The breaker itself was two maps and eight private methods on the service and leaked its in-flight entry to callers.

**The maintenance scheduler and its steps were one function.** `runConsolidationPassInternal` spelled out challenge, merge, reflection and persona inline, each with its own try/catch, stat accounting, touched flag, fence check and audit write.

**The facade sat inside the graph it composes.** `EmbeddingPipeline` reached its own reindex/backfill and `MaintenanceService` reached its own consolidation pass through `MemoryService` lambdas, so the timer callback depended on a facade method staying a pass-through.

## What changed

- `coordinateBatchWrites` is the only decision kernel. A remember is a one-candidate batch carrying its dispatch-commit callback in a `WriteContext`. Provenance-settled candidates (archived, duplicate, suppressed, challenged, forgotten) apply synchronously before any provider round trip can age their owner snapshot; unowned candidates check the tombstone before recall; the rest go through `retrieveForDecisions`, one batch decision call, apply, and one bounded retry that reuses query vectors. The commit boundary fires outside every store transaction right before the first row write, which matters because memory and Tape share one SQLite connection and the Execution Journal refuses to commit inside a host transaction. `coordinateWrite`, `retrieveForDecision` and the dead `allowInsert` / `excludeConflictParticipants` switches are removed.
- `MemoryRowMutations.resolveClaimOwnership` / `classifyClaimOwner` return a `ClaimOwnership` union (`unowned`, `archived`, `duplicate`, `suppressed`, `challenged`, `superseded`) and every write site switches on it; `settleOwnedClaim` is the single response to an owned claim and `isForgottenClaim` the single tombstone lookup. `supersedeHead` and `handleProvenanceHit` are private again. The live-target and challenged-head predicates live in `domain/stateModel`.
- `RetrievalService.retrieve` orchestrates `recallVectorCandidates`, `refillCandidates` and `pruneStaleVectors` over a `RecallState` literal; cancellation points, degradation causes and latency accounting are unchanged, and the candidate counts reported on a cancelled recall reflect the last completed round. `QueryEmbeddingCircuitBreaker` (`infra/queryEmbeddingCircuit.ts`) owns the breaker and exposes `start`, `state`, `reset`, `clear`. `retrieveForDecisions` embeds the query only once the vector store is available and has a ready certificate.
- `MaintenanceService` runs `heavyPasses` (challenge, merge, reflection, persona) in order under one budget; each pass returns a uniform step result and is fenced independently. Pass audits go through a fail-open helper so a failed audit insert cannot erase a step's LLM accounting.
- `EmbeddingPipeline` and `MaintenanceService` call their own methods directly; the scheduling tests observe the owner instead of the facade.
- Review fix: the retry-phase settle loop had no failure accounting, so a store error while restoring an archived owner escaped the kernel and skipped the failed audit, the domain-mutation mark and finalization for candidates already written in the same batch. Both preparation passes now run through `settleImmediateCandidates` with the same accounting as apply.
- Review fix: moving the reflection and persona audit writes into their passes meant an audit insert failure dropped the step's call counts and could classify a successful reflection next to a failed merge as all-steps-failed, rolling back the watermark and arming the failure cooldown.

## Behavior changes

- The dispatch commit for the `memory_remember` tool moves from before neighbor recall to right before the first row write, after recall and the decision call. A remember that ends in a NOOP or no-neighbor retry no longer commits a dispatch it never executed. An already-aborted tool call may spend one recall and one decision round before the abort surfaces; the result is discarded.
- Single remembers use the batch decision prompt with one candidate and the batch neighbor recall, so they no longer pass through the warm-recall breaker or the adaptive refill loop.
- A forgotten candidate in an extraction batch costs no embedding or decision call.
- On a cold vector store, decision recall does not embed the query, so a CAS retry in that window stays FTS-only where it previously reused a vector the first round had computed anyway.
- A failure while preparing an extraction batch records the failed `memory/extract` audit and finalizes already-written candidates, the same as a failure during apply.
- `casRetries` counts retried decisions only; a retried candidate that settles on provenance is not a retry.
- Failed heavy maintenance passes log as `<step> pass failed`.

## Docs

No changes. `docs/architecture/memory-system.md` already states the contracts this PR keeps: tombstone lookup inside the insert transaction, breaker scope limited to injection recall, commit boundary outside store transactions, supersede with a derivation edge, and the stop/drain/start maintenance sequence.

## Verification

- `pnpm run format:check`, `pnpm run lint`, `pnpm run i18n`, `pnpm run typecheck`
- `pnpm run test:memory`: 54 files, 928 tests. Six new tests pin the contracts the merge must keep (a forgotten claim never reaches recall, the decision model or the dispatch commit; the commit fires once before the first row write and propagates a journal failure; SUPERSEDE keeps the old claim and its derivation edge; a forgotten extraction candidate settles before recall; a CAS retry reuses its query vector). Two regression tests cover the review fixes and fail with their fix reverted.
- Tape layer boundaries, DeepChat memory coordinator, memory tools and contract suites: 8 files, 143 tests
- `pnpm run test:memory:eval`: 6 tests
- Native memory suites ran their non-gated tests (118 of 334); the SQLite/DuckDB-gated tests need a native rebuild and this PR does not touch the data layer.

## Summary by CodeRabbit

- **New Features**
  - Added resilient query processing with request deduplication, concurrency controls, cancellation handling, and recovery after provider failures.
  - Improved memory-claim handling for archived, duplicate, suppressed, challenged, and superseded content.
  - Added more consistent batch decision processing and ownership resolution.

- **Bug Fixes**
  - Improved handling of partial write failures so completed changes are preserved.
  - Preserved maintenance and reflection accounting when individual operations fail.
  - Improved stale or missing embedding recovery during vector-store maintenance.

- **Tests**
  - Expanded coverage for memory extraction, decision processing, embedding recovery, and maintenance failure scenarios.

### DIFF
--- a/src/main/memory/domain/stateModel.ts
+++ b/src/main/memory/domain/stateModel.ts
@@ -2,7 +2,8 @@ import type {
   AgentMemoryEmbeddingState,
   AgentMemoryKind,
   AgentMemoryLifecycleState,
-  AgentMemoryStatus
+  AgentMemoryStatus,
+  CanonicalAgentMemoryRow
 } from './types'
 import {
   AGENT_MEMORY_EMBEDDING_STATES,
@@ -148,6 +149,36 @@ export function isRecallableMemoryState(row: CanonicalMemoryStateSource): boolea
 
 export function isEmbeddingEligibleState(row: CanonicalMemoryStateSource): boolean {
   return isRecallableMemoryState(row) && row.embedding_state === 'pending'
+}
+
+// The only row shape a correction decision may UPDATE, SUPERSEDE, or CHALLENGE: owned by the
+// Agent, current head of its chain, active, and not already inside a conflict.
+export function isLiveDecisionTarget(
+  agentId: string,
+  row: CanonicalAgentMemoryRow | undefined
+): row is CanonicalAgentMemoryRow {
+  return (
+    !!row &&
+    row.agent_id === agentId &&
+    row.superseded_by === null &&
+    row.lifecycle_state === 'active' &&
+    row.conflict_state === null
+  )
+}
+
+// A chain head that is itself under challenge cannot absorb new evidence until the conflict
+// resolves; writes that land on it stay conservative no-ops.
+export function isChallengedDecisionHead(
+  agentId: string,
+  row: CanonicalAgentMemoryRow | undefined
+): row is CanonicalAgentMemoryRow {
+  return (
+    !!row &&
+    row.agent_id === agentId &&
+    row.lifecycle_state === 'active' &&
+    row.superseded_by === null &&
+    row.conflict_state === 'challenged'
+  )
 }
 
 export function assertValidMemoryInsertState(input: {
--- a/src/main/memory/domain/types.ts
+++ b/src/main/memory/domain/types.ts
@@ -615,10 +615,12 @@ export type ProvenanceHitResult =
   | { action: 'noop'; reason: string }
 
 /**
- * Side-effect-free classification of who currently owns a candidate's provenance key.
- * `archived` owners may be restored, `duplicate` owners may only absorb temporal metadata,
- * `suppressed` owners must stay untouched, `challenged` chains reject new evidence, and a
- * `superseded` owner exposes its live chain head (or null) for correction decisions.
+ * Who currently owns a candidate's provenance key. Classification itself never writes; the
+ * lookup that precedes it may lazily re-key a legacy owner, which is why resolvers accept the
+ * caller's dispatch-commit callback. `archived` owners may be restored, `duplicate` owners may
+ * only absorb temporal metadata, `suppressed` owners must stay untouched, `challenged` chains
+ * reject new evidence, and a `superseded` owner exposes its live chain head (or null) for
+ * correction decisions.
  */
 export type ClaimOwnership =
   | { state: 'unowned' }
--- a/src/main/memory/domain/types.ts
+++ b/src/main/memory/domain/types.ts
@@ -614,6 +614,20 @@ export type ProvenanceHitResult =
   | { action: 'continue' }
   | { action: 'noop'; reason: string }
 
+/**
+ * Side-effect-free classification of who currently owns a candidate's provenance key.
+ * `archived` owners may be restored, `duplicate` owners may only absorb temporal metadata,
+ * `suppressed` owners must stay untouched, `challenged` chains reject new evidence, and a
+ * `superseded` owner exposes its live chain head (or null) for correction decisions.
+ */
+export type ClaimOwnership =
+  | { state: 'unowned' }
+  | { state: 'archived'; owner: CanonicalAgentMemoryRow }
+  | { state: 'duplicate'; owner: CanonicalAgentMemoryRow }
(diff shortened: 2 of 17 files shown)


zerob13 (approved):
## Review verdict

**APPROVE**

Reviewed HEAD `a65e31d285a78d9cc5ac5bdd333d058540ea950` against `dev`. This is a large but cohesive internal memory refactor; responsibilities are split without changing the established write, recall, circuit-breaker, maintenance, fencing, cancellation, or dispatch-commit contracts. I found no blocking correctness issue, unjustified breaking change, or DeepChat style/design violation.

### Review focus

- **Correctness:** single and batch writes share the same coordination kernel; ownership/tombstone handling and CAS retry accounting remain coherent.
- **Concurrency and lifecycle:** recall fencing, read epochs, cancellation, and query-embedding circuit state are preserved.
- **Design:** extracted services and pass-table structure are proportionate to the existing complexity; no clear over-engineering beyond the scope of the refactor.
- **Tests:** added regression coverage targets write ownership, retry failure, audit failure, cancellation diagnostics, and workload bounds; no obviously excessive test-only surface.
- **Breaking changes:** no externally visible contract break identified.

### Verification

- TypeScript check passed.
- Oxfmt and Oxlint passed.
- Relevant memory tests passed: **242/242**.
- Full renderer/native integration suites were not run; no data-layer changes were identified that would make this a blocker.

No blocking findings.

PR #2277, opened 2026-09-09
feat(skills): discover workspace skills

zerob13:
Project-local SKILL.md files were absent from both composer suggestions and runtime resolution. Discover skills under the current workspace's `.agents/skills`, `.deepchat/skills`, `.claude/skills`, `.codex/skills`, and `.cursor/skills`, and carry that scope through selection, routing, skill views, and execution snapshots.

Project skills take precedence over shared names within their workspace. They stay outside global skill management and use default runtime settings, so same-named shared credentials and script overrides cannot leak into project execution. Workspace switching discards stale UI responses and pending selections; window focus refreshes discovery, with repeated focus events coalesced while a composer refresh is in flight. The catalog route rejects relative paths and null bytes before service dispatch. Existing layout and keyboard interactions are preserved.

```text
BEFORE                      AFTER
/project-a                  /project-a
/skill                      /skill
+ global-skill              + project-skill
                            + global-skill
```

Validation: format, i18n, lint, and main/renderer type checks passed. The full main suite passed 8533 tests (493 skipped); targeted composer suites passed 9 tests. Regression coverage includes filesystem discovery and unsafe paths, precedence, two-session content isolation, credential isolation, shared deletion, and late composer responses.

Discovery is limited to the selected workspace's conventional skill directories; it refreshes on workspace selection, window focus, and runtime resolution without adding filesystem watchers.

## Summary by CodeRabbit

* **New Features**
  * Added workspace-scoped Project Skills discovered from supported project directories.
  * Project Skills appear in chat skill lists and slash-command suggestions without importing them into the shared library.
  * Project Skills take precedence over same-named shared skills within the workspace.
  * Skill catalogs refresh when switching workspaces, focusing the app, or when catalog changes occur.
* **Behavior Changes**
  * Project Skills use default runtime settings and remain isolated from shared assignments, credentials, and management actions.
  * Sessions without a project and ACP agents retain their existing behavior.
  * Skill resolution now uses conversation context for more accurate session-specific results.
* **Bug Fixes**
  * Improved protection against catalog changes during skill execution-package capture.

### DIFF
--- a/docs/architecture/shared-skills/spec.md
+++ b/docs/architecture/shared-skills/spec.md
@@ -17,6 +17,12 @@ DeepChat stores each mutable Skill package once under the global Skills root. A 
 a logical binding to a Skill; no Agent-owned copy or generated link is created. Read-only bundled
 and Plugin-owned Skills stay in their provider roots but appear in the same global list.
 
+[Project Skills](../../features/project-skills/spec.md) are a separate, request-scoped overlay on this
+shared library. They are discovered from the selected workspace, can shadow shared names within
+that workspace, and have no shared assignment or extension state. They are excluded from global
+management, import/export, and deletion. Materialized project sources use default runtime settings
+and cannot inherit credentials from a same-named shared binding.
+
 The renderer exposes only:
 
 - `Skills` for the global list;
--- a/docs/features/project-skills/plan.md
+++ b/docs/features/project-skills/plan.md
@@ -8,7 +8,7 @@
 - [x] Add focused regression protection after implementation; run format, i18n, lint, typecheck,
       and relevant main/renderer suites.
 
-Validation: format, i18n, lint, and main/renderer type checks pass. The selected main suites pass
-677 tests (2 skipped); the selected renderer suites pass 17 tests. Regression coverage includes
+Validation: format, i18n, lint, and main/renderer type checks pass. The full main suite passes
+8531 tests (493 skipped); the selected renderer suites pass 17 tests. Regression coverage includes
 workspace discovery, precedence, unsafe paths, content and credential isolation, session selection,
 shared deletion, and stale composer responses.
--- a/docs/features/project-skills/plan.md
+++ b/docs/features/project-skills/plan.md
@@ -9,6 +9,7 @@
       and relevant main/renderer suites.
 
 Validation: format, i18n, lint, and main/renderer type checks pass. The full main suite passes
-8531 tests (493 skipped); the selected renderer suites pass 17 tests. Regression coverage includes
+8533 tests (493 skipped); the targeted composer suites pass 9 tests. Regression coverage includes
 workspace discovery, precedence, unsafe paths, content and credential isolation, session selection,
-shared deletion, and stale composer responses.
+shared deletion, stale composer responses, focus refresh coalescing, route path validation, and
+project catalog overflow degradation.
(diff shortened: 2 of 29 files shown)


zhangmo8 (commented):
## Review: Project Skills discovery

Overall this is a well-scoped and carefully isolated change. I like that project skills never touch the shared cache/management state, that symlink escapes are rejected via `resolvePhysicalSkillPath` at both root and per-skill level, and that credentials/extension overrides cannot be inherited by same-named project skills (enforced again in `resolveSkillRuntimeEnvironmentBinding`). The regression tests covering precedence, isolation, and per-session resolution are thorough.

A few non-duplicate points beyond the existing CodeRabbit comments (route-boundary absolute-path validation and operation-local discovery memoization):

### 1. Run tool universe loses its size cap for project sessions (src/main/agent/deepchat/runtime/toolResolver.ts:377)

The `projectDir?.trim()` branch builds the snapshot with `getMetadataList(...)` directly and skips both `MAX_RUN_TOOL_UNIVERSE_SKILLS` and the `overflow` degradation path. For a project session the catalog is shared skills + project skills, so a user with many shared skills now gets an unbounded universe exactly where the cap existed for a reason. Consider applying the same cap (or a project-specific bound) and emitting the existing degradation signal for consistency.

### 2. Window focus rescans are unbounded and per-instance (src/renderer/src/components/chat-input/composables/useSkillsData.ts:293)

Every window focus fires a full five-directory workspace scan over IPC, and each mounted composable instance registers its own `focus` listener, so multiple composers multiply the scans. Suggest debouncing (e.g. VueUse `useWindowFocus` + `watchDebounced`) or at least coalescing focus events while a scan is already in flight (the sequence guard prevents stale results but not redundant IPC).

### 3. Project skills skip catalog-stability assertions — intentional? (src/main/skill/index.ts, `resolveFreshEffectiveSkillContents` / `viewSkillForAgent`)

For project skills, `metadataCache` is a fresh scan, so the first `metadataCache.get(name) !== metadata` check is trivially true, and the second check against `getMetadataCacheForAgent` is explicitly skipped via `!metadata.projectRoot &&`. Net effect: a skill file modified between manifest load and execution-package snapshot is no longer detected through catalog identity. The file re-read will pick up new content, so it is not a correctness hole, but the shared-skill TOCTOU guarantee is silently weaker for project skills. If intentional, a short comment in the code (and spec) would help future readers.

### 4. Workspace trust consideration (product/security)

Opening any workspace now auto-discovers and activates skills from `.agents/.deepchat/.claude/.codex/.cursor/skills`, including shadowing shared skill names within that workspace. A cloned repository can therefore change agent behavior (system prompt routing, available skills) on first message with no user confirmation. The spec documents shadowing as intended, which is fine, but consider a first-discovery trust prompt for untrusted workspaces (similar to VS Code workspace trust) as a follow-up — this is a meaningful prompt-injection surface for repos cloned from the internet.

### 5. Minor typing nit (src/renderer/src/components/chat-input/composables/useSkillsData.ts:33)

`projectSkills` is typed `SkillMetadata[]` but holds `UnifiedSkillItem[]` from `getUnifiedSkillCatalog`. It works because `UnifiedSkillItem` structurally extends the metadata shape, but typing the ref as `UnifiedSkillItem[]` (or aliasing) would be more honest.

Non-blocking; 1 and 2 are the ones I would most like to see addressed. Skipped anything CodeRabbit already raised (relative `workspacePath` rejection at the route boundary, operation-local discovery memoization) to avoid duplicates.

PR #2279, opened 2026-09-09
fix(agent): restore live assistant on resume

zerob13:
Answering an ask-user interaction can fail with `Provider projection clone removed its bound Skill context.` because paused assistant blocks live in the mutable transcript until terminal settlement. An initialized Tape can omit the assistant message entirely, preventing the resume context builder from identifying its owning user and injecting the selected Skill. A stale Tape revision can also omit the latest question and answer.

Merge the exact target assistant message from the current transcript into Tape history before resume compaction and refresh it after compaction. Validate its session and role, replace any stale revision, and retain the ordering used by context assembly. Other history continues to come from Tape.

Validation:
- Add missing/stale Tape target regressions that materialize a Skill, assemble a real initial provider request, pause, answer a question, and reach the resumed provider request with the Skill, current assistant content, and answer intact.
- Against the unfixed code, the missing-target case reproduces the reported stack; the stale-target case sends the old reply without the answered tool call.
- All 554 tests across the harness, context builder, compaction service, input preparation coordinator, and Skill materializer pass.
- Format, i18n, lint, and Node/renderer typechecks pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resuming interrupted conversations when the saved assistant message is missing or outdated.
  * Ensured resumed sessions use the latest assistant response and preserve recovered skill content, live answers, and tool results.
  * Improved conversation history ordering and consistency during resume operations.

* **Tests**
  * Added coverage for resuming skill-based turns with missing or stale saved messages.

### DIFF
--- a/src/main/agent/deepchat/runtime/turnCoordinator.ts
+++ b/src/main/agent/deepchat/runtime/turnCoordinator.ts
@@ -1654,29 +1654,25 @@ export class TurnCoordinator {
       let basePromptAssembly = unguardedBasePromptAssembly
       let baseSystemPrompt = basePromptAssembly.prompt
       let resumeTargetOrderSeq: number | undefined
+      const readResumeHistory = () =>
+        runSynchronousPreStreamStep(sessionId, 'tape-ready', () => {
+          const { historyRecords } = this.ports.tapeReconciliation.ensureSessionTapeReady(
+            sessionId,
+            this.ports.messageStore
+          )
+          const target = this.ports.messageStore.getMessage(messageId)
+          if (!target || target.sessionId !== sessionId || target.role !== 'assistant') {
+            throw new Error('Resume target assistant message is no longer available.')
+          }
+          // Paused assistant blocks remain mutable until terminal settlement, so Tape may omit
+          // this message or hold an older revision without the latest interaction response.
+          return [...historyRecords.filter((record) => record.id !== messageId), target].sort(
+            (left, right) => left.orderSeq - right.orderSeq
+          )
+        })
       const preparedInput = await this.ports.inputPreparationCoordinator.prepareExisting({
-        ensureHistory: () =>
-          runSynchronousPreStreamStep(
-            sessionId,
-            'tape-ready',
-            () =>
-              this.ports.tapeReconciliation.ensureSessionTapeReady(
-                sessionId,
-                this.ports.messageStore
-              )
-                .historyRecords
-          ),
-        refreshHistory: () =>
-          runSynchronousPreStreamStep(
-            sessionId,
-            'tape-ready',
-            () =>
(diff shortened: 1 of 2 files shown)


src/main/agent/deepchat/runtime/turnCoordinator.ts
```diff
@@ -1654,29 +1654,25 @@ export class TurnCoordinator {
...
+            this.ports.messageStore
+          )
+          const target = this.ports.messageStore.getMessage(messageId)
+          if (!target || target.sessionId !== sessionId || target.role !== 'assistant') {
+            throw new Error('Resume target assistant message is no longer available.')
+          }
+          // Paused assistant blocks remain mutable until terminal settlement, so Tape may omit
+          // this message or hold an older revision without the latest interaction response.
+          return [...historyRecords.filter((record) => record.id !== messageId), target].sort(
+            (left, right) => left.orderSeq - right.orderSeq
```
zhangmo8, line 1670: nit (non-blocking): `historyRecords` should already come back ordered from the Tape projection, so `filter + append + sort` re-sorts the full history on every ensure/refresh call. Splicing the target at its `orderSeq` position would avoid the O(n log n) pass — likely negligible in practice, so fine to keep as-is for simplicity.

zhangmo8 (commented):
## Review

**Verdict: looks good — correct root-cause fix with solid regression coverage.** 👍

The diagnosis is right: paused assistant blocks stay mutable in the transcript until terminal settlement, so the Tape projection can omit the assistant message entirely or hold a stale revision without the latest interaction response. Merging the exact live target into the resume history (replacing any stale Tape revision and restoring `orderSeq` ordering) is the right fix, and re-reading the target inside `refreshHistory` guarantees compaction sees the latest answered blocks.

Things I verified:

- `getMessage` returns `ChatMessageRecord | null`, which always carries `orderSeq`, so the merge/sort is type-safe and the target keeps its position in context assembly.
- The fail-fast validation (session id + assistant role) is consistent with the neighboring Skill-context errors in this method, and surfacing a hard resume error beats silently resuming with a broken/partial context.
- Deduplicating `ensureHistory`/`refreshHistory` into one `readResumeHistory` removes the previous copy-paste — good cleanup.
- The added `it.each(['missing', 'stale'])` regressions assert the actual provider request (recovered Skill body, current assistant content, pending tool call + answer), which is the level of coverage this needs.
- CI is green (build / static / test-main / test-renderer / test-native-memory all pass).

One more minor nit outside the diff, take or leave: with the target now guaranteed present in `historyRecords`, the `?? this.ports.messageStore.getMessage(messageId)?.orderSeq` fallback when resolving `resumeTargetOrderSeq` in `prepareIntent` is unreachable and could be simplified to just the `historyRecords.find(...)` lookup.

PR #2284, opened 2026-09-10
chore(deps): upgrade dependencies and fix compatibility

zhangmo8:
## Summary

Upgrade project dependencies across the board and adapt the code to the resulting breaking changes. Electron is capped at 43.6.0 (44 introduces the renderer clipboard removal); toolchain majors that would change runtime behavior or protocol compatibility are deliberately deferred (see below).

## Upgrades

Runtime:
- `electron` 41.10.4 → 43.6.0 (bundles Node 24.20, matches `engines >=24.18 <25`)
- `better-sqlite3-multiple-ciphers` 12.9.0 → 13.0.3, `opendal` 0.49.5 → 0.49.9 (with `@opendal/lib-*`), `@duckdb/node-api` 1.5.4-r.1 → 1.5.5-r.4
- `ai` ^7.0.94 and all `@ai-sdk/*` to latest in-range, `@aws-sdk/*` ^3.1128.0, `zod` ^4.5.4
- `electron-store` ^8.2.0 → ^11.0.2, `diff` ^9, `undici` ^8.10.2, `nanoid` ^6.0.1, `level` ^10, `https-proxy-agent` ^9, `pdf-parse-new` ^2.1.0, `tokenx` 2.1.0, `@modelcontextprotocol/ext-apps` 2.0.0, `@ff-labs/fff-node` ^0.10.6, `vue-router` ^5.3.1
- `@tiptap/*` 2.27 → 3.31.3 (history shim kept, `setContent` options object, no-op `addOptions` overrides removed)
- Dev: `oxlint` ^1.82, `taze` ^21, `@playwright/test` ^1.63, `katex` ^0.18.7, `@vueuse/core` ^14.4, `@iconify-json/*` latest, and the usual patch/minor refreshes

## Compatibility fixes

- **Electron 43 preload clipboard → main-process IPC**: clipboard APIs are leaving renderer processes (fully removed in 44). Image/text copy and read now go through main-process IPC handlers (`src/shared/clipboardChannels.ts` + `src/main/app/clipboardIpc.ts`, registered in `composition.ts`), so the app is already compatible with the removal when Electron is bumped again. `readClipboardText` is async end-to-end (`Promise<string>` through `preload/index.d.ts`, `runtime.ts`, `DeviceClient`); `copyText` / `copyImage` keep their fire-and-forget signatures so MessageItem / `usePageCapture` call sites are unaffected
- **Electron 43**: `websql` removed from `clearStorageData` (also gone in 43)
- **Tiptap 3**: `commands.setContent` takes an options object (`{ emitUpdate: false }`); no-op `addOptions` overrides removed (v3 types `parent()` as partial)
- **ext-apps 2.0**: `Tool` type now comes from `@modelcontextprotocol/client` (`McpAppView.vue`)
- **tokenx 2.x**: `approximateTokenSize` → `estimateTokenCount` (src + test mocks); OCR artifact cache revision bumped to `tokenx=2.1.0`; token-derived test fixtures (Tool Surface V1 hashes, pressure-history sizes, steer rebudget size) regenerated
- **zod 4.5.4**: intersections now merge compatible objects and only emit nested `allOf` for conflicts — `toDeepChatJsonSchema` rejects any `allOf` recursively instead of only at the root
- **electron-store 11 in tests**: ESM-only package is inlined in `vitest.config.ts`; shared electron mock exposes a `default` export; `userData` mocks point to a real per-worker temp directory (conf 15 mkdirs eagerly), cleaned up on worker exit
- **CI**: regenerated stale icon collections (`@iconify-json/*` bump); dropped the obsolete "prepare native SQLite for the Node ABI" step (better-sqlite3-multiple-ciphers 13 ships N-API prebuilds and has no `install` script)
- `opendal` version pins in `afterPack` / `electron-builder` tests updated to 0.49.9
- `MemoryDatabase.agentMemoryTable` getter hardened against property-narrowing issues

## Deliberately not upgraded

- `electron` 44 (renderer clipboard removal — the IPC bridge in this PR already prepares for it)
- `vite` 8 (electron-vite 5 peer range caps vite at ^7)
- `@agentclientprotocol/sdk` 1.x (protocol migration, 20 call sites)
- `pinia` 4 / `@pinia/colada` 1.x / `vitest` 5 (deferred to keep this PR behavior-neutral)
- `typescript` stays on the pinned `typescript-native-bridge` 6.0.3 override (latest available bridge)

## Verification

- `typecheck:node`, `typecheck:web`, `lint`, `format:check` all green
- `test:renderer`: 274 files / 2458 tests green (with tiptap 3)
- `test:main`: 641 files / 9024 tests green on Electron 43.6.0

Note for local runs on Node 26: jsdom's `localStorage` is shadowed by Node's experimental global; run renderer tests with `NODE_OPTIONS="--localstorage-file=/tmp/x"` if you hit storage errors. CI (Node 24) is unaffected.


src/preload/index.ts
```diff
@@ -22,7 +23,8 @@ const api = Object.freeze({
   },
   copyImage: (image: string) => {
     const img = nativeImage.createFromDataURL(image)
-    clipboard.writeImage(img)
+    const pngBlob = new Blob([new Uint8Array(img.toPNG())], { type: 'image/png' })
+    void clipboard.write([new ClipboardItem({ 'image/png': pngBlob })])
```
zerob13: [P1] Move clipboard access out of the preload renderer process

Electron 44 removes the Electron clipboard module from renderer processes, including a non-sandboxed preload. In a real 44.3.0 BrowserWindow with `sandbox: false` and `contextIsolation: true`, `typeof require('electron').clipboard` and `typeof require('electron').ClipboardItem` both return `undefined`. Consequently this call throws, and the unchanged `copyText` / `readClipboardText` functions fail too. MessageItemUser/MessageItemAssistant copy actions and usePageCapture.captureAndCopy still use this bridge. Route these operations through typed main-process IPC (or the supported browser clipboard API), and propagate/await the asynchronous results through the client contracts. Official removal notice: https://www.electronjs.org/docs/latest/breaking-changes#removed-clipboard-module-is-no-longer-available-in-the-renderer-process

package.json
```diff
@@ -104,152 +104,152 @@
...
     "@modelcontextprotocol/server": "2.0.0",
-    "@parcel/watcher": "^2.5.6",
+    "@parcel/watcher": "^2.6.0",
     "@zerob13/nativekit": "0.6.3",
-    "ai": "^7.0.54",
-    "axios": "^1.18.1",
-    "better-sqlite3-multiple-ciphers": "12.9.0",
+    "ai": "^7.0.94",
+    "axios": "^1.20.0",
+    "better-sqlite3-multiple-ciphers": "13.0.3",
```
zerob13, line 132: [P2] Update the native-memory setup for the SQLite 13 package

The upgraded package no longer declares an `install` script, but `.github/workflows/prcheck.yml:160` still executes `pnpm --dir node_modules/better-sqlite3-multiple-ciphers run install`. This fails deterministically with `ERR_PNPM_NO_SCRIPT: Missing script: install`, before the native SQLite smoke test and memory suite can run. I reproduced the command against the frozen lockfile; the same failure is present in the required test-native-memory job. Adapt that preparation step to the new package's native loading/build mechanism and rerun the native checks as part of this upgrade.

package.json
```diff
@@ -104,152 +104,152 @@
...
-    "@commitlint/config-conventional": "^21.1.0",
-    "@electron-toolkit/tsconfig": "^1.0.1",
+    "@antv/infographic": "^0.2.20",
+    "@commitlint/cli": "^21.2.2",
+    "@commitlint/config-conventional": "^21.2.2",
+    "@electron-toolkit/tsconfig": "^2.0.0",
     "@electron/notarize": "^3.1.1",
-    "@iconify-json/lucide": "1.2.109",
-    "@iconify-json/vscode-icons": "1.2.50",
+    "@iconify-json/lucide": "1.2.130",
```
zerob13, line 171: [P2] Regenerate the committed icon collections for the new packages

The icon package upgrades change the generator inputs, but this PR leaves `src/renderer/src/lib/icons/icon-collections.generated.ts` unchanged. After a frozen-lockfile install, `pnpm run icons:check` fails with `Generated icon files are stale: src/renderer/src/lib/icons/icon-collections.generated.ts`; the required static job fails at exactly this step. Run the existing icon generator, commit its output, and confirm `icons:check` passes so the required checks remain usable.

zerob13 (changes requested):
The dependency migration has one confirmed runtime regression and two reproducible required-check failures, detailed inline. Please address these before merging.

Validation at this commit: format, i18n, lint and full typecheck passed; 28 targeted file/OCR/build/skill/tape test files passed (325 tests). A real Electron 44.3.0 hidden-window probe confirmed the unavailable preload clipboard exports. Both failing CI commands were reproduced locally with the frozen lockfile. The successful mocked tests and package builds do not cover the preload runtime regression.

zerob13:
Electron 不要升级到44，有break change
先最高维持在 43

zhangmo8:
Thanks for the thorough review — all three points are addressed in 659df6c9d:

**[P1] Preload clipboard → main-process IPC**: You're right — Electron 44 removed the clipboard module from renderer processes entirely (preload included), so the previous `ClipboardItem` approach in the preload was still broken. Clipboard operations now go through main-process IPC:

- New `src/shared/clipboardChannels.ts` channel contract + `src/main/app/clipboardIpc.ts` handlers, registered in `composition.ts` next to `registerDeepchatRoutes`
- `src/preload/index.ts` now bridges `copyText` / `copyImage` / `readClipboardText` via `ipcRenderer.invoke`; no clipboard imports remain in the preload
- `readClipboardText` is now async end-to-end (`Promise<string>` through `preload/index.d.ts`, `runtime.ts`, `DeviceClient`); it has no external callers today, and `copyText` / `copyImage` keep their fire-and-forget signatures so the MessageItem / `usePageCapture` call sites are unaffected
- The `file.copyImage` route (FileClient) already ran in main and needed no change

**[P2] sqlite `install` script**: Removed — better-sqlite3-multiple-ciphers 13 ships N-API prebuilds and has no `install` script. The `test-native-memory` job already validated the shipped prebuild (928 memory tests ran green before the old rebuild step). The obsolete "Prepare native SQLite for the Node ABI" step is deleted, and `prcheckWorkflow.test.ts` now asserts it stays deleted.

**[P2] Icon collections**: Regenerated with `pnpm run icons:generate` and committed; `icons:check` passes (the static job is green on the latest run).

Also addressed the CodeRabbit note: mocked `userData` directories are now per-Vitest-worker (`process.pid` suffix) with `exit` cleanup in `test/setup.ts`, `toolService.test.ts`, and the three provider test mocks.

Verification on 659df6c9d: format / lint / typecheck (node + web) green, `test:renderer` 274 files / 2458 tests green, targeted main suites (toolService 81, provider mocks 36) green.

zerob13 (changes requested):
## Review result: CHANGES_REQUESTED

I re-reviewed the new head `4fe2c2607d2eb5027ced3c29315a0a7d622dcf00` after the previous review. The earlier Electron clipboard regression is fixed, and the native-memory/static checks are currently green. However, the required PR check is still red:

### P1 — Required `test-main` check fails on the PR head

- **Evidence:** PR Check run `34443225925` reports `test-main` failed: `test/main/orchestration/liveDelegationRepository.test.ts:1228` — `rejects an evaluation reference from another parent Tape incarnation` expected `requireTurn(...)` to throw, but it returned normally. The job summary is `640 passed, 1 failed`; `pr-required` consequently fails as well.
- **Impact:** This is a required check failure, so the PR is not mergeable. The failure is in a security/integrity boundary that validates evaluation references are bound to the correct parent Tape incarnation.
- **Request:** Reproduce this failure on the PR branch, identify whether the dependency/API migration changed the repository behavior or exposed an existing incompatibility, and restore the expected rejection with a regression test that passes in the full `test-main` job. Please rerun all required checks after the fix.

### Verification notes

- The new main-process clipboard IPC path is directionally correct: preload no longer imports Electron clipboard APIs, and the handlers are registered in main.
- `static`, `build`, `test-renderer`, and `test-native-memory` are green on the current head; the blocking issue above is the observed `test-main` failure, not the earlier stale-generated/native concerns.
- No over-engineered test additions or unrelated design regressions were found in the reviewed delta.

Please address the required test failure before merge.

zhangmo8:
> Electron 不要升级到44，有break change 先最高维持在 43

已处理：`f9d6ca2ac` 已将 electron 固定在 `43.6.0`（package.json + lockfile），不再升级到 44；顺带把基于 44 的 `ClipboardItem` 写法回退为 43 的 `clipboard.writeImage`，preload 剪贴板 API 继续走 main IPC。当前分支 CI 全绿。

zerob13 (approved):
## Review result: APPROVED

本次只复审了上一轮之后的新提交 `f9d6ca2..415c760`。

此前阻塞合并的 `test-main` 失败已修复：测试现在会稳定生成一个不同的 Tape identity，不再因原 identity 以 `0` 开头而偶发生成相同值。生产代码中的跨 Tape identity 校验没有改动，仍会拒绝绑定错误的 evaluation reference。

本次改动仅修正测试数据构造：没有生产行为变化、没有额外设计复杂度，也没有过度增加测试。当前 required checks 全部通过。

PR #2286, opened 2026-09-10
feat(agent): reap orphaned child processes

xiao-text:
ACP agents, MCP stdio servers, and detached background shell sessions can survive an application crash, retaining ports and file locks. This change persists launch records and recovers attested orphan processes before each subsystem starts new work.

The shared registry stores records beneath the effective Electron userData directory, including isolated E2E runs and utility processes. Each record includes the child PID, owner PID, launch command, write time, and OS-observed process start time. Cleanup requires matching process identity, preserves records when observation or termination is inconclusive, skips live foreign owners, and never terminates the current process. Legacy records without an observed start time remain ineligible for termination until they expire or their process is gone.

POSIX observation reads the fixed start time with portable `ps lstart` in UTC and the C locale. Windows explicitly serializes the CIM creation time as an ISO timestamp. Failed process queries do not count as confirmed exit, and queries have a bounded timeout. Asynchronous identity capture cannot recreate a cleared record or overwrite a newer launch.

ACP and background execution retain records until confirmed exit or termination. MCP records are created when the transport starts, with a unique ID per transport so overlapping reconnects cannot delete one another's records. Startup waits for in-flight orphan recovery before launching replacement work.

Validation: format, i18n, lint, and node/web typechecks pass locally. The relevant process, ACP, and MCP suites cover 287 passing tests across 13 files, including a live OS process-observation check on macOS, PID reuse, failed observations, asynchronous record cleanup, overlapping MCP clients, unconfirmed termination, and startup ordering. GitHub CI validates the merge result.

## Summary by CodeRabbit

- **New Features**
  - Added automatic tracking for ACP agent, background execution, and MCP server processes.
  - Added stale-process cleanup with safeguards that verify process identity before termination.
  - Added recovery before starting related services, ensuring stale processes are addressed first.

- **Bug Fixes**
  - Process records are cleared when processes exit or terminate successfully.
  - Records remain available when termination cannot be confirmed, enabling later recovery.

- **Tests**
  - Added coverage for process tracking, cleanup, identity verification, stale-process recovery, and termination edge cases.


zerob13 (changes requested):
## Review: changes requested

I found four correctness issues in process lifecycle cleanup that can defeat the stated orphan-recovery guarantee:

### Blocking findings

1. **ACP launch records are cleared before process exit**
   - `src/main/agent/acp/runtime/acpProcessManager.ts:2429`
   - `killChild()` clears the registry immediately after sending termination. If termination is unconfirmed, or the child has not emitted `exit` yet, the record disappears while the process may still be alive. A later startup can no longer reap it. Keep the record until the child `exit` path; this must also work for children killed during initialization.

2. **Background-exec records are cleared on forced finalization even when termination fails**
   - `src/main/agent/shared/process/backgroundExecSessionManager.ts:1027`
   - `killInternal()` calls `finalizeSession()` after `terminateProcessTree()` returns `false`, and `completeSessionFinalization()` unconditionally clears the registry record. This loses the only recovery record for a still-running process. Preserve the record when termination is unconfirmed, and clear it only after a confirmed termination or the child close/exit path.

3. **MCP stdio cleanup ignores the termination result**
   - `src/main/mcp/mcpClient.ts:905-931`
   - Both `forceTerminateStdioProcessTree()` and `closeTransport()` clear the record regardless of whether `terminateProcessTreeByPid()` returns `false` (the function reports failure without throwing). This can report success / remove the record while the process remains alive. Use one helper that checks the boolean result, only clears on `true`, and preserves the PID/record on failure for startup reaping.

4. **The registry path bypasses Electron's configured userData directory**
   - `src/main/agent/shared/process/childProcessRegistry.ts:58-60`
   - The implementation hard-codes `os.homedir()/.deepchat/child-processes`. In E2E and other runs that call `app.setPath('userData', ...)`, records are written outside the app data root, so isolation and cleanup are broken and tests can affect the host user's persistent directory. Resolve the root from Electron's `app.getPath('userData')` after the override is applied, deferring resolution until first use if needed.

### Validation

The targeted regression suite passes at this commit: 5 files, 199 tests. The passing tests do not cover the unconfirmed-termination and configured-userData paths above.

Please address these findings and add focused regression coverage before merge.

zerob13:
### Additional blocking finding: MCP instance records collide

The initial review also needs to cover a separate P1:

- `src/main/mcp/mcpClient.ts:794-801, 908-932`

MCP stdio records use `recordId: this.serverName`, and cleanup also clears by server name. If an old transport overlaps a reconnect/new transport for the same server, the new PID overwrites the old record; when the old client closes, it deletes the new record. The new process can then survive an app crash without a recoverable record. Bind a unique record ID (for example server name + PID + UUID) to each transport/client instance and clear only that instance's record. Add a regression test for overlapping same-name clients.

This is in addition to the four lifecycle/path blockers in the CHANGES_REQUESTED review. The targeted suite passed, but it does not cover this overlap scenario.

zerob13 (approved):

PR #2288, opened 2026-09-10
fix(provider): refresh settings without lag

zerob13:
Provider settings can stall when switching providers, retain stale model switches after a successful update, and hide API key status failures such as DeepSeek 401 responses.

- Wait for native accessibility status before mounting model rows, avoiding a full catalog render followed by virtualization. Keep the full accessible list when required and retain the accessible fallback if the native lookup fails.
- Replace changed virtual row snapshots so model status updates and rollback reach the visible switches immediately.
- Read the provider configuration index before requesting individual model configurations, retaining user overrides and falling back to individual reads if the index fails.
- Show key status errors below the API key, redact the stored key, reload when credentials change, and ignore stale responses after navigation or credential replacement.

- Restrict provider initialization and model-toggle diagnostics to development builds. Performance tracing also requires the existing trace setting; production builds skip the timing and paint-wait probes.

### UI

```text
BEFORE                            AFTER
API Key [********9b2a]             API Key [********9b2a]
                                  ! 401: API key invalid

Model A [ON] --click--> [ON]       Model A [ON] --click--> [OFF]
  Refreshes after switching          Updates immediately;
  away and back                     rolls back on save failure
```

### Validation

- 76 tests pass across provider components, model store, and native accessibility support.
- Format, i18n, lint, and Node/renderer typechecks pass. The 36 tests covering the logging change also pass.
- Production Electron/Vite build passes; emitted renderer JavaScript contains no model-toggle or provider-initialization diagnostic markers.
- Four new model-switch/key-error regression cases fail against the original implementation.
- Real Chromium component fixture: mouse and keyboard toggles, virtual scrolling, and key error display checked at two viewport sizes and in light/dark themes.
- In the isolated 500-model browser fixture, provider-switch median decreased from 658 ms to 97 ms across three runs per condition, with 16 rendered rows. This measures component mounting with a simulated native accessibility lookup, rather than end-to-end Electron startup or provider network latency.

## Summary by CodeRabbit

* **New Features**
  * Added inline, sanitized error messages for provider API key status checks.
  * Added loading feedback while accessibility settings are initializing.
  * Improved model configuration refreshes to preserve user-defined overrides.

* **Bug Fixes**
  * Prevented outdated credential or provider responses from overwriting current status.
  * Improved model list updates and loading behavior when accessibility information is delayed.
  * Preserved model loading when configuration lookups fail.

### DIFF
--- a/src/renderer/settings/components/ModelProviderSettingsDetail.vue
+++ b/src/renderer/settings/components/ModelProviderSettingsDetail.vue
@@ -265,7 +265,7 @@ const validateApiKey = async () => {
   try {
     const resp = await providerStore.checkProvider(props.provider.id)
     if (resp.isOk) {
-      console.log('验证成功')
+      if (import.meta.env.DEV) console.debug('Provider verification succeeded')
       checkResult.value = true
       showCheckModelDialog.value = true
       // 验证成功后刷新当前provider的模型列表
@@ -274,7 +274,7 @@ const validateApiKey = async () => {
       await modelStore.applyInitialModelRecommendations(props.provider.id)
       return true
     } else {
-      console.log('验证失败', resp.errorMsg)
+      if (import.meta.env.DEV) console.debug('Provider verification failed:', resp.errorMsg)
       checkResult.value = false
       showCheckModelDialog.value = true
       return false
@@ -310,7 +310,7 @@ watch(
 )
 
 const initProviderSettings = async () => {
-  console.log('initData for provider:', props.provider.id)
+  if (import.meta.env.DEV) console.debug('initData for provider:', props.provider.id)
 
   await providerStore.ensureDefaultProvidersReady()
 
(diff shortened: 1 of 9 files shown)


zerob13:
## Review result: CHANGES_REQUESTED

### P1 — 回退了已修复的 DeepSeek 推理历史兼容性问题

- 位置：`src/main/provider/deepseekResponsesAdapter.ts:201`
- 这个 PR 在 provider settings 刷新改动之外，回退了此前用于保留 DeepSeek 推理历史兼容性的修复，并删除了覆盖该行为的 3 个回归测试。

现在只要历史消息带有非空 `reasoning_content`，`omitEmptyReasoning()` 就直接保留原始 `reasoning_provider_options`。其中的 `deepseek.reasoningContent` 若为 `null`、空数组，或不含有效 `reasoning_text`，会再次被传给上游 SDK。SDK 会生成无法重放的 reasoning item，DeepSeek 会拒绝后续请求。

影响是已有 DeepSeek Responses 会话可能在继续发送消息时直接失败，尤其是含有摘要推理历史的会话。

请恢复这层兼容性处理和对应回归测试；若确实要改变该行为，也必须在 PR 描述中明确说明，而不能作为设置刷新改动的附带回退。

### 已检查的其他部分

provider status 刷新、模型列表索引和虚拟列表状态更新的实现范围合理，新增测试也基本围绕实际行为，不存在额外阻塞问题。当前必需检查通过，但不能覆盖上述已删除的推理历史回归场景。

PR #2301, opened 2026-09-14
perf(sync): read the backup database asynchronously

xiao-text:
## Why

Backing up a large database froze the whole UI: `readFileSync` made the main thread
wait on disk, stalling every IPC, render and stream flush for seconds.

Electron's main process has a single event loop shared by IPC handlers, window
rendering and stream flushes. A synchronous read parks that loop, so on a large
database the app is unresponsive for the entire duration of the read.

## What changed

```diff
-      files[ZIP_PATHS.agentDb] = new Uint8Array(fs.readFileSync(this.DB_PATH))
+      files[ZIP_PATHS.agentDb] = new Uint8Array(await fs.promises.readFile(this.DB_PATH))
```

The read now goes through `fs.promises`, which hands the work to the libuv
threadpool and leaves the event loop free. The surrounding steps already used async
fs (`writeFile`, `unlink`, `rename`, `stat`, `addOptionalFile`), and `zipAsync`
(fflate callback API) already compresses on a worker thread, so this was the last
synchronous holdout in the backup path.

Both APIs return a `Buffer` and `Uint8Array(...)` copies it either way, so the
archive bytes are unchanged.

## Trade-off

The snapshot is now taken while other tasks can run, so commits landing during the
read may be absent from the backup — the worst case is losing the last few
milliseconds of writes. Two things bound this:

- The TRUNCATE checkpoint just before it keeps the copied file internally
  consistent (`checkpointDatabaseForBackup()` runs `wal_checkpoint(TRUNCATE)`).
- In WAL mode ordinary writes go to the `-wal` file, not the main DB file being
  copied, so they only land in that file after the next checkpoint.

SQLite's own backup API would be the strict fix if that window ever matters. Not
done here: it is a behavioural change on a path with no test coverage, and
better-sqlite3 implements it as `setImmediate` + batched `transfer(rate)` on the
main thread, so it yields rather than truly running off-thread.

## Risk

One line, byte-identical output, no schema or format change. If the read rejects,
the existing `catch` already removes the temp zip and emits the `error` status, so
the failure path is unchanged. No UI, API or config change.

## Test plan

- [ ] `pnpm test:main -- test/main/sync` — restore suite (32 cases) unchanged
- [ ] `pnpm oxfmt --check .` / `pnpm lint` / `pnpm typecheck`
- [ ] Manual: Settings → Sync → create backup on a large DB, confirm the UI stays
      responsive during `collecting`, then restore from the produced archive and
      confirm the app opens with data intact

## Follow-ups (out of scope)

1. The backup-**create** path has zero test coverage — the 32 sync tests only cover
   restore. A create → restore round-trip test (plus one against a sqlcipher
   encrypted DB) is the hard gate before touching this chain further.
2. Streaming compression (fflate `Zip` + `AsyncZipDeflate` writing straight to disk)
   to avoid materialising the DB twice in memory; ~60–80 lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup consistency by capturing the agent database and supporting files from the same point in time.
  * Improved backup reliability during ongoing database activity.
  * Backups remain usable when database checkpointing is temporarily blocked by other readers.
  * Restoring a backup now preserves required write-ahead log data and removes stale sidecar files.
  * Prevented incomplete backup archives when database changes occur during collection.

* **Performance**
  * Large backup archives are now written in a streaming manner, improving handling of files larger than 4 MB.


zerob13 (changes requested):
## Request changes

这次改动只改了一行，目标是避免在主进程同步读取大数据库时卡住 UI，这个方向是合理的。但当前实现把 `readFile` 放到了 `checkpointDatabaseForBackup()` 之后，数据库文件的复制过程没有被同步保护，备份可能生成损坏或不可恢复的 `agent.db`。

### P1 — 读取期间数据库文件仍可能被 checkpoint 修改

`src/main/sync/index.ts:577-579`

`wal_checkpoint(TRUNCATE)` 只保证 checkpoint 调用返回时 WAL 已处理完，并不能冻结后续数据库活动。`await fs.promises.readFile(this.DB_PATH)` 会让出主进程事件循环；在这段等待期间，其他写入仍然可以提交，WAL 也可能触发自动 checkpoint，或者被其他显式 checkpoint 处理。这样 SQLite 可能一边把 WAL 页写回并截断主数据库文件，一边被 `readFile` 复制，最终压缩包里的数据库不是某个一致时刻的快照。

这会直接破坏备份/恢复链路，尤其是数据库较大、备份耗时较长且应用仍在写入时。PR 描述中“TRUNCATE checkpoint just before it keeps the copied file internally consistent”并不成立；它只描述了 checkpoint 当下的状态，不能覆盖异步读取窗口。

请保留异步读取带来的 UI 响应性，同时使用项目已有的数据库同步机制，把 checkpoint 和数据库快照读取放进同一个可靠的保护边界；或者采用 SQLite backup API/等价的快照方案。需要明确保证读取期间不会有写入或 checkpoint 改变被复制的主数据库文件，并补充覆盖并发写入/自动 checkpoint 的回归测试。

### 参考分析

- 当前 diff：`src/main/sync/index.ts:577-579`
- 现有数据库配置启用了 WAL：`src/main/data/connectionConfig.ts:20`
- 现有恢复备份逻辑也特别处理了主库与 `-wal` 文件：`src/main/data/mainDatabase.ts:255-267`
- PR 当前只有恢复测试；本次 create-backup 路径没有覆盖并发写入场景。

zhangmo8 (commented):
Agree with the existing P1 on the WAL/auto-checkpoint race during the async `agent.db` read — no need to repeat it. Two related but distinct points that haven't been raised yet:

### Cross-file snapshot coherence (`src/main/sync/index.ts:576-582`)

Before this change, `readFileSync` blocked the main process, so collecting `agent.db`, `appSettings`, custom prompts, and system prompts was effectively atomic with respect to other main-process code. With the `await`, the event loop now yields between `checkpointDatabaseForBackup()` and each subsequent read, so IPC handlers can interleave mid-collection: DB writes, settings changes, or prompt edits can land between the reads. The archive then contains files captured at different instants — e.g. `appSettings` consistent with a *newer* DB state than the copied `agent.db`. Note that fixing only the DB read (per the fence / SQLite backup API suggestions above) does not close this gap: the fence or snapshot boundary should span the **entire collection phase** (checkpoint → all reads), not just the database read.

### Minor: main-thread copy cost remains (`src/main/sync/index.ts:579`)

`new Uint8Array(await fs.promises.readFile(...))` still copies the entire buffer synchronously on the main thread. The async read removes the I/O wait (threadpool), but for a very large `agent.db` the blocking `Uint8Array` copy — plus the later in-memory zip at level 6 — still causes jank. If the motivating issue was UI freezes on large databases, this alone may not fully resolve it; worth measuring.

zhangmo8 (commented):
Thanks for the follow-up — `b3026` properly addresses the previously raised WAL/auto-checkpoint race (the read mark held by the snapshot connection spans the whole collection phase, which also closes the cross-file coherence gap) and the main-thread copy cost (zero-copy views + streaming zip). I checked out the branch and ran `test/main/sync/backupConsistency.test.ts` + `syncService.test.ts` locally: all 27 tests pass.

### Fallback backup can silently drop un-checkpointed WAL transactions (`src/main/sync/index.ts:741-746`, `src/main/data/backupReadLock.ts:8-22`)

When `drainWal()` fails (long-lived reader, only ~60 ms of retry budget), the fallback copies just the main DB file. Any transactions still sitting in the WAL are silently absent from the backup — the restore path in `mainDatabase.ts` knows how to apply a `-wal` file, but the archive never contains one.

I verified this with a throwaway probe against this branch: with a blocked checkpoint, a commit of 1500 rows landing before `startBackup()`, the archived `agent.db` contains only the 120 seeded rows — a usable, integrity-clean image that is nonetheless missing ~93% of the data. The existing test "still produces a usable image when another reader blocks the checkpoint" masks this because it only asserts `count >= SEED_ROWS` and integrity, not completeness.

Suggestions:

1. In the fallback path, include the `-wal` (and `-shm`) file in the archive, so restore recovers the full state; or fail loudly / emit a warning instead of producing a quietly-stale backup.
2. Tighten the blocked-checkpoint test to assert the concurrent 1500-row commit is present in the archive — that would have caught this.

### Minor: fallback silently reintroduces the jank this PR fixes

The sync `readFileSync` fallback is exactly the pre-PR behavior, and the 3×20 ms drain budget makes it likely on any busy system. If the fallback becomes WAL-aware per the above, an async read of `agent.db` + `-wal` under the same retry budget might let the fallback stay responsive too. Not blocking on its own, but worth considering together with point 1.

The rest of the design (read-mark fencing, PASSIVE drain, streaming `writeZipToDisk` with backpressure) looks sound to me.

zhangmo8 (commented):
Independently checked `b3026` against the discussion so far — the WAL/auto-checkpoint race, cross-file coherence, and fallback-completeness points are all already covered above, so these are just two small robustness nits in the new lock helper that nobody has raised yet:

### Connection leak if `BEGIN` fails (`src/main/data/backupReadLock.ts:33-34`)

If `openDb()` succeeds but `db.exec('BEGIN')` throws (e.g. a cipher/decryption error on the snapshot connection), the connection is never closed — the `try` block only starts after `BEGIN`. Worth moving the connection acquisition and `BEGIN` inside the `try` (or adding an outer `finally`) so a failed snapshot does not leak a file handle to the (possibly encrypted) database.

### `ROLLBACK` failure masks the original error (`src/main/data/backupReadLock.ts:41`)

In the `catch` branch, `db.exec('ROLLBACK')` runs before `throw error`; if the rollback itself throws, the caller sees the rollback failure instead of the real cause. Wrapping the rollback in its own try/catch (swallowing or logging only the secondary error) would preserve the original failure for diagnostics.

Both are non-blocking. One behavioral note, not a request for change: since `work()` rethrows on read errors, a transient failure inside the lock path aborts the whole backup rather than falling back to the sync read — that arguably fails loudly, which is the right call, just flagging it in case silent aborts show up in crash reports.

zerob13 (changes requested):
## Request changes

这次增量会把 fallback 备份中的 `-wal` 一并归档，解决了“只复制主库必然漏掉 WAL 已提交数据”的问题。但备份仍不能保证可恢复。

### P1 — fallback 仍未得到一致的 SQLite 快照

`src/main/sync/index.ts:772-786`

fallback 先异步读取 `agent.db`，再异步读取 `agent.db-wal`，两次读取之间没有 SQLite 读事务或其他快照保护。

WAL drain 被长期 reader 阻塞时，其他连接仍可在主库读取完成后继续提交或 checkpoint；`-wal` 可能被重置、删除或复用。归档因此可能组合来自不同时间点的主库和 WAL，恢复时会缺少已提交数据，甚至无法打开数据库。

请改用 SQLite backup API 或其他能够产出单一一致镜像的方式。若 fallback 无法保证完整性，应明确失败，而不是生成可能无法恢复的备份。并补充一个回归测试：在主库与 WAL 两次读取之间发生提交或 checkpoint，验证生成的归档仍可完整恢复。

### P2 — `BEGIN` 失败会泄漏 snapshot connection

`src/main/data/backupReadLock.ts:33-46`

`db.exec('BEGIN')` 在 `try/finally` 外。连接创建成功但 `BEGIN` 失败时，`db.close()` 不会执行。

请把 `BEGIN` 放进现有的 `try`，确保所有异常路径都会关闭连接。也请单独保护 `ROLLBACK`，避免 rollback 失败覆盖原始错误，影响排查。

### 参考

- 本次增量：`b302678..ba234550`
- 新增 WAL round-trip 测试是必要的，不属于过度测试；但它始终维持 blocker，未覆盖两次文件读取之间的竞态。
- 未发现 UI、公共接口或项目风格方面的额外问题。

zhangmo8 (commented):
Checked `fddf957` — the `BEGIN` fix is correct: moving `db.exec('BEGIN')` inside the `try` means a failing `BEGIN` now falls through to `finally` (connection closed), and the `db.inTransaction` guard correctly skips the `ROLLBACK` on that path. CI is green on this commit.

Status of the previously raised items at head `fddf957` (not re-reviewing, just flagging they remain open):

- **Fallback snapshot consistency (P1)** — `readBackupFilesFallback()` still reads `agent.db` and `agent.db-wal` as two independent async reads with no snapshot boundary between them; the archive can still mix SQLite generations.
- **AsyncZipDeflate backpressure (Major)** — input slices are still pushed before the drain gate is checked, and `archive.ondata` writes regardless of `drain`.
- The `ROLLBACK` error-masking nit is also still pending.

One small new point not raised yet:

### Minor: `Buffer.from(chunk)` copies every compressed chunk (`src/main/sync/index.ts:719`)

In `archive.ondata`, `output.write(Buffer.from(chunk))` copies each compressed chunk (fflate emits fresh `Uint8Array`s, ~64 KB by default) onto the main thread before handing it to the stream. A byte-offset view avoids the extra copy:

```diff
-        if (!output.write(Buffer.from(chunk))) {
+        if (!output.write(Buffer.from(chunk.buffer, chunk.byteOffset, chunk.byteLength))) {
```

This is safe here because fflate allocates a new output buffer per `ondata` call (it does not reuse them). Non-blocking on its own — worth folding in if the backpressure rework touches this function anyway.

zhangmo8 (commented):
Reviewed `5fd89d2..e1b4a1d` (WAL snapshot pinning, zip backpressure rework, async support files). I verified the snapshot invariant end to end, and all previously raised blocking items look resolved at head:

- **Fallback snapshot consistency (zerob13's P1)**: pinning a read mark for the whole copy is sound — the WAL cannot reset while the mark is held, any backfilled frames are at or below the mark so the shipped `agent.db` + `-wal` pair replays to one generation, and appends beyond the mark are handled by WAL checksum validation on restore. The new mid-copy-reset and sneak-commit regression tests cover exactly these races.
- **`BEGIN` leak / `ROLLBACK` error masking**: fixed (`rollbackSilently` + the `inTransaction` guard) with unit tests.
- **Backpressure**: the always-`await drain` gate before each slice push now bounds in-flight input, and the `Buffer.from(chunk.buffer, ...)` view removes the per-chunk copy.
- **Support files** (`e1b4a1d`): reads are asynchronous now, closing the last sync-read holdout. Note this widens the window in which commits land during collection on the acquired path, but the read mark keeps the main db file frozen during `work()`, so the snapshot only gets staler — it never becomes inconsistent.

Two small new items, both non-blocking:

### Minor: pending `drain` promise is never resolved on the archive error path (`src/main/sync/index.ts:714-718`)

`output.destroy()` without an error argument does not emit `'error'`, so if a backpressure `drain` promise is pending at that moment, the producer coroutine stays suspended on `await drain` forever: the outer promise is already rejected, but the IIFE retains its closure — including the `files` record, which can hold the full `agent.db` image (potentially hundreds of MB) — for the lifetime of the process. The same applies to the `catch` branch at lines 746-749. Suggest `output.destroy(error)` (the existing `output.once('error', ...)` listener then releases the gate), or resolving `drain` via `output.once('close', ...)`.

### Trivial: fallback warning is inaccurate for the new bail-out cause (`src/main/sync/index.ts:760-765`)

`{ acquired: false }` now also occurs when the guard's re-drain finds frames it cannot backfill (a commit landed between the pre-drain and the snapshot mark) — in that case no reader is blocking the checkpoint. The message still says "a reader is blocking the checkpoint", which will mislead diagnostics. Suggest wording that covers both causes, e.g. "could not take a fully drained WAL snapshot (blocked checkpoint or a commit landed during the drain window)".

Everything else looks good from my side. CI is green on head (`test-renderer` was still pending at the time of writing), and the remaining CodeRabbit nitpick (redundant `new Uint8Array(data.slice(...))` before `entry.push`) is already tracked. Ready once zerob13 re-checks the fallback pinning.

zhangmo8 (approved):
Reviewed `626a484` (harden backup error paths and fs races). All three changes look correct:

1. **`output.destroy(error)` on both zip failure paths** (`src/main/sync/index.ts:723`, `:755`) — this properly releases the pending backpressure drain gate so the producer coroutine no longer stays suspended and retains the `files` record (including the full `agent.db` image) for the process lifetime. The double `reject` (once via the stream `'error'` handler, once explicit) is a harmless no-op.

2. **Broadened fallback warning** — accurately reflects both ways the drain window can lose a fully drained snapshot (blocked checkpoint *or* a commit landing during the window). Good diagnostic hygiene.

3. **`existsSync` check-then-act replaced with direct operation + `ENOENT` handling** across settings, prompt store, temp backup, WAL sidecar, and cleanup paths — semantics are preserved (each `ENOENT` maps to the exact same early-return/skip as the old existence check), so this closes the delete-between-check-and-use races without changing behavior on the happy path or on genuine I/O errors.

No regressions to the previously resolved blocking items (snapshot pinning, fallback WAL completeness, `BEGIN`-in-try, zip backpressure). CI is green on head. With the blocking concerns verified resolved earlier and this hardening commit checked, approving from my side.

PR #2304, opened 2026-09-16
fix(auth): harden OAuth credential storage

xiao-text:
## Summary

Harden OAuth credential persistence for OpenAI Codex and xAI Grok against
crash corruption, power loss, and silent credential loss.

- Target branch: `dev`
- No UI changes (main-process only, no BEFORE/AFTER layouts needed)

## Problem

Both `OpenAICodexCredentialStore` and `XaiGrokCredentialStore` wrote
`credentials.json` directly to its final path and swallowed all load errors in
a single `catch { return null }`:

- A crash or power loss mid-write left a truncated JSON file. On next launch
  the refresh token was unrecoverable and the user was silently signed out.
- `load()` could not distinguish "file missing" from "file corrupted" or
  "decryption failed" (e.g. a transiently locked keyring). Corruption was
  reported to the user as `signed-out`, and re-login then overwrote a file
  that might still have held valid credentials.
- Transient decryption failures were classified as corruption, which both
  misinformed the user and polluted the `.corrupt` backup with healthy
  ciphertext.
- Errors thrown downstream (`sign-in is required`) were misleading when the
  real cause was an unreadable credential file.
- `clear()` only removed the main file, leaving credential material behind
  after logout, and a single failed removal skipped the remaining artifacts.

## Changes

**`openaiCodex/credentialStore.ts` / `xaiGrok/credentialStore.ts`** (mirrored):

- Atomic writes with durability: write to a uniquely named temp file
  (`credentials.json.tmp-<pid>-<random>`, same pattern as
  `mcp/oauthCredentialStore.ts`) via an fd, `fsyncSync` before `renameSync`,
  and remove the temp file in a `finally` if the rename fails. The credential
  file is never observed half-written, even across power loss.
- `load()` classifies reads via a private `readEnvelope()` into
  `missing` / `ok` / `undecryptable(reason)` / `corrupt(reason)`. `ENOENT`
  maps to `missing`; decryption failures (possibly transient keyring locks)
  are separated from structural corruption.
- Corruption is surfaced instead of swallowed: `console.warn` plus a
  `getLoadError()` accessor that callers consume.
- Before overwriting a structurally corrupted file, `save()` copies it to
  `credentials.json.corrupt` (copy, not move, so the original path never
  disappears mid-save; fixed name, at most one backup). Undecryptable files
  do not pollute the backup.
- `clear()` now attempts every artifact independently (main file, `.corrupt`,
  legacy `.tmp`), so one failed removal never skips the others. Failures warn
  instead of being swallowed.
- `getStorageState()` warns when the encryption-availability check fails
  before falling back to plaintext file storage.

**`openaiCodex/index.ts` / `xaiGrok/index.ts`**:

- `getStatus()` reports `error` with the current load error (preferred over
  the stale `lastError`) instead of `signed-out` when credentials exist but
  are unreadable.
- `getBackendAuth` / `getAccessToken` / `forceRefresh*` throw the actual
  load-error reason instead of `sign-in is required`.


zhangmo8 (approved):
## Review

Verified on the PR head (`590de99e`): read the full diff and both baseline stores, cross-compared the two implementations line-by-line, checked every `getLoadError()` consumer for staleness, compared the atomic-write pattern against `mcp/oauthCredentialStore.ts`, and ran the 4 touched test files — 36/36 pass. The three earlier review comments (non-isolated artifact cleanup in both `clear()`/logout flows, and `getStatus` load-error precedence) are already fixed at head with dedicated regression tests, so those aren't repeated here.

The missing/corrupt classification is correct, `lastLoadError` can't go stale (every consumer calls `load()` immediately before reading it, and `save()`/`clear()` reset it), and the atomic-rename design is sound for the stated process-crash threat. Approving; the items below are non-blocking polish.

### Findings

1. **minor — `src/main/provider/auth/xaiGrok/credentialStore.ts:111` and `src/main/provider/auth/openaiCodex/credentialStore.ts:115` — fixed `.tmp` filename deviates from the repo's own atomic-write pattern.** `mcp/oauthCredentialStore.ts:581` uses a unique temp name (`${this.filePath}.tmp-${process.pid}-${randomBytes(6).toString('hex')}`), while both new stores use a fixed `credentials.json.tmp`. A crash between `writeFileSync` and `renameSync` leaves the tmp file behind indefinitely (under `file` storage it can contain partial plaintext tokens), self-healing only on the next `save()`/`clear()`. Concurrency is prevented only by the single-instance lock, not by the code itself. Suggested: use the unique-suffix pattern and remove the tmp file in a `finally` if `renameSync` throws.

2. **minor — `readEnvelope` catch (~L166-170) + backup check (~L103-110) — transient decryption failures are classified as corruption, which pollutes the backup.** A transiently locked keychain/keyring makes `readEnvelope()` return `corrupt: credential decryption failed`, so `getStatus()` shows `error` and `save()` copies the *still-valid* ciphertext to `credentials.json.corrupt` — clobbering any earlier backup with a file the store itself labels corrupt. Nothing is destroyed, but the backup's forensic value is undermined and the user is told a healthy file is corrupted. Suggested: distinguish `undecryptable` from structurally corrupt (only back up on parse/shape failures), or at minimum skip the backup when the reason starts with `credential decryption failed`.

3. **minor — `openaiCodex/credentialStore.ts:115-120` (same in xaiGrok) — no fsync before rename.** Process crash mid-write is handled, but on power loss the rename can be durable while the tmp data is not, leaving a truncated/empty `credentials.json` — the exact failure mode this PR aims to prevent. Suggested: write via an fd and `fh.sync()` before `renameSync`. The `mcp` store shares this gap, so a follow-up hardening both would also be acceptable.

4. **minor (test gap) — `openaiCodex/index.ts:309,326` and `xaiGrok/index.ts:317,326` — the new throw-site error propagation is untested.** No test asserts that `getBackendAuth` / `forceRefreshBackendAuth` / `getAccessToken` reject with the load-error reason; only the `getStatus` precedence path is covered. The store tests also assert `copyFileSync` was called but never that the backup content equals the corrupt file. Suggested: one test per provider — corrupt the file, expect `auth.getBackendAuth()` to reject with `/not valid JSON/`.

### Positives

- Stale-error hygiene is done right: every `getLoadError()` consumer calls `load()` immediately beforehand, so the surfaced reason is always fresh, and `save()`/`clear()` reset it.
- The rename-based save is structurally sound for its threat model: same-directory tmp (no `EXDEV`), `0o600` on creation, backup via `copyFileSync` so the final path never disappears mid-save, and the single-instance lock rules out concurrent writers.
- Tests assert real behavior, not just calls: the precedence tests plant a realistic prior auth error and prove the load error wins — exactly the semantics that matter.

zerob13 (approved):
## Review result

✅ **Approve**

I reviewed the current HEAD `590de99e`. The credential-store changes are reasonable and consistent with the existing DeepChat patterns: atomic temp-file replacement, corrupt/missing state distinction, corrupt-file backup, independent cleanup, and propagation of load errors to auth status/calls. I did not find a blocking correctness, security, compatibility, or over-design issue.

### Verification

- `git diff --check`: passed
- Focused Vitest could not start locally because the worktree dependencies are incomplete (`@vitejs/plugin-vue` is unavailable).
- Existing GitHub review evidence reports the relevant tests passing at this HEAD.

<details>
<summary>Detailed analysis</summary>

- The fixed `.tmp` path is acceptable here because DeepChat uses a single-instance lock; logout cleanup now attempts all credential artifacts independently.
- Missing files clear the transient load error, while malformed or undecryptable envelopes are surfaced as actionable load errors without weakening auth behavior.
- No unnecessary production abstraction or excessive test scope was introduced.
</details>

zerob13 (approved):
## Re-review: delta since `590de99e` — ✅ Approve

I approved this PR earlier at `590de99e`. Since then the author pushed two commits addressing review feedback: `075d82ad` (nits) and `fd15074f` (temp-file cleanup). I reviewed the full delta `590de99e..fd15074f` (+289/−48, 7 files) against the complete head sources of both credential stores. The xaiGrok store mirrors openaiCodex change-for-change; I verified every hunk on both sides.

### What the delta does, and why it's fine

**1. Randomized temp file + durable write (`save()`)**

The fixed `.tmp` path is now `credentials.json.tmp-<pid>-<12 hex chars>`, written through `openSync(..., 'w', 0o600)` → `writeFileSync(fd)` → `fsyncSync(fd)` → `closeSync` → `renameSync`, with a `finally` that removes the temp file even when the rename throws. This fixes two real problems at once:

- Two concurrent writers can no longer clobber each other's temp file and rename a torn file into place (the old fixed name made that possible).
- A failed rename no longer leaves a stale temp file with credential material sitting in the directory. The cleanup after a *successful* rename is a harmless no-op (`rmSync` with `force: true` on a path that no longer exists).

The random name also removes the fixed, predictable temp path, which is strictly better for a credentials directory.

**2. New `undecryptable` state, split from `corrupt`**

`safeStorage.decryptString` failing (locked or rotated OS keychain) is no longer classified as file corruption. I checked the consequences:

- `load()` still records the reason via `getLoadError()`, so the provider status keeps surfacing it — the "prefer credential load error in status" behavior from `f0f66a0d` is retained.
- `save()` only backs up genuinely `corrupt` files to `.corrupt`; an undecryptable envelope is now overwritten without a backup. That's the right call: a successful save means the user just re-authenticated, so the old wrapped tokens are obsolete, and backing up undecryptable envelopes would only accumulate files nobody can read.
- A JSON-parse failure *inside* the decrypted payload is still `corrupt` with an accurate message, so the two error paths are now distinguishable instead of collapsed into one misleading "decryption failed".

**3. `clear()` removes randomized temp files**

Because temp names are no longer fixed, `clear()` now scans the credential directory for the `<basename>.tmp-` prefix, while still removing the legacy `.tmp` and `.corrupt` artifacts. `ENOENT` on the directory scan is tolerated; every other failure warns and continues instead of aborting the cleanup. The prefix is anchored to the credential filename, so unrelated files in the directory can't be deleted (covered by tests on both stores).

### Criteria check

- **Over-engineering**: none. `fsync` + randomized names + cleanup-on-failure are proportionate for a credential store, not gold-plating.
- **Style consistency**: typed result states, `console.warn` prefixes, and the `node:crypto` import all match existing `src/main` conventions.
- **Breaking changes**: none. Public/IPC behavior is unchanged; the `.corrupt` backup semantics were tightened deliberately and the reasoning holds (see 2).
- **Tests**: proportionate to the new behavior — rename-failure cleanup, randomized-temp `clear()`, undecryptable-no-backup, fd-aware `fs` mock (with the missing `fsyncSync` added to `test/setup.ts`). The per-store duplication in tests mirrors the mirrored implementations, which the base review already accepted.
- **CI** on `fd15074f`: build / static / test-main / test-renderer / test-native-memory all green.

No blocking findings on the delta. Approving.

zhangmo8:
Synced with the latest `dev` (merged `6b57eb72`, which includes #2306, #2307, #2308, #2309, #2310, #2311) and pushed `eb1d2d3e`.

- No code changes to this PR's own diff; merge was conflict-free (disjoint file sets).
- All required CI checks pass on the new head: static, test-main, test-renderer, build, test-native-memory, package-required, pr-required.
- All prior review threads remain resolved.

Requesting a fresh review on the updated head. 🙏

zhangmo8 (approved):
## Re-review: delta since `590de99e` — ✅ Approve

Reviewed the changes since my last approval: `075d82ad` (review nits), `fd15074f` (clear randomized temp files), and the merge from `origin/dev` (verified it does not touch the credential stores).

**What I verified on the PR head (`fd15074f`):**

- Randomized `.tmp-<pid>-<random>` temp names eliminate concurrent-writer clobbering; the file is created with `0o600` via `openSync`, `fsync`ed before `renameSync`, and the `finally` cleanup handles both rename failure and stale temps. Good.
- Splitting `undecryptable` from `corrupt` keeps keyring/decryption failures diagnosable without changing load behavior (both still surface `lastLoadError` and return `null`). The "payload is not valid JSON" message is now accurate.
- `clear()` now sweeps `.tmp-*` leftovers in addition to the legacy fixed `.tmp` path, with ENOENT tolerated and other readdir failures logged. Prefix matching only removes the store's own namespace, and `rmSync` without `recursive` unlinks (not follows) any symlink, so no traversal concern.
- Test coverage is solid: temp-path pattern + fsync assertions, rename-failure cleanup, backup-snapshot correctness, undecryptable-file handling, and the randomized-sweep cases for both stores.

One non-blocking nit for a future pass: `fsync` covers the temp file but not the parent directory before the rename, so the rename itself is not durable across a crash. Fine to defer for credential files.

CI is green (`test-main`, `static`, `build`). The only previously open finding (CodeRabbit's `.tmp-*` cleanup) is addressed in `fd15074f`. Nothing blocking — happy to see this merged.

PR #2305, opened 2026-09-16
feat(plugin): remote distribution for plugins and OCR runtime

zerob13:
## Summary

Remote distribution foundation for heavyweight optional capabilities (spec: `docs/features/plugin-remote-distribution/spec.md`):

- **Distribution catalog** (`resources/plugin-catalog.json`): sha256-pinned per-platform artifacts with ghproxy-style mirror chains. Stable builds resolve `stable` channel entries only; dev builds can resolve `pre-release` entries via the `DEEPCHAT_PLUGIN_CATALOG` override hook (ignored in packaged builds).
- **Remote installers**: shared download core (probe all candidates in parallel, fastest wins, resume + stall watchdog + sha256 via the toolchains downloader) reused by the plugin installer and the OCR runtime asset installer.
- **Plugins**: `plugins.enable` transparently downloads and installs a catalog-declared plugin before enabling it; the plugins hub shows a downloadable section with progress/cancel/retry.
- **OCR runtime**: payload becomes a downloadable asset (`userData/runtimes/ocr/<version>/`) with the same identity verification as bundled payloads (resolver falls back bundled → installed); first use triggers a silent background download (the triggering turn skips OCR text extraction); 5-minute failure cooldown; `ocrRuntimeAutoDownload` settings toggle (default on).
- **Release tooling**: `plugin:catalog` generates catalog entries from built artifacts (incl. OCR payload packaging); `plugin:catalog:verify` fetches and verifies every target (release gate).

No runtime `npm install`: artifacts are prebuilt, lockfile-reproduced, checksummed packages.

## UI

```
BEFORE (bundled)                        AFTER (on-demand)
+-----------------------------+        +-----------------------------+
| Plugins Hub                 |        | Plugins Hub                 |
|  CUA    [Enabled]           |        |  CUA    [Install 96MB 62%]  |
|  OCR    [Built-in]          |        |  OCR    (first use: silent   |
|  Feishu [Enabled]           |        |         background download)|
|  all payloads ship in the   |        |  Feishu [Enabled]           |
|  installer                  |        |  not downloaded = absent    |
+-----------------------------+        +-----------------------------+

OCR settings: new "Runtime download" section
  [Download OCR engine] [Cancel] [Retry] + auto-download switch (default on)
```

## Commits

1. `docs(plugin)` — RFC + implementation plan (SDD)
2. `feat(plugin)` — shared catalog/route/event/settings contracts
3. `feat(plugin)` — catalog loader + remote installer + `installOfficialPluginPackage` + enable fallback (+ 23 tests, incl. full-chain L1 integration test)
4. `feat(ocr)` — runtime asset installer + coordinator + resolver fallback + settings/routes (+ 16 tests)
5. `feat(app)` — composition wiring
6. `feat(renderer)` — plugins hub + OCR settings UX, i18n for all 23 locales
7. `build(plugin)` — catalog generation/verification tooling + extraResources + npm scripts

## Verification

- `typecheck:node` / `typecheck:web` / `format` / `lint` / `i18n` (validate + i18n-check): pass
- `test:main`: 8573 passed (37 new tests; two full-chain tests cover catalog-override → download → `.dcplugin` checksum verification → PluginService install, and OCR fixture payload → installer → resolver identity verification → available)
- `test:renderer`: 2432 passed
- `plugin-catalog.mjs generate/verify` exercised against local fixtures (good catalog passes; corrupted sha256 rejected)

## Not in this PR (release flips, tracked in plan.md)

- CI publishing `.dcplugin`/OCR zip artifacts to a prerelease release and generating real catalog URLs
- Removing the cua bundle step and `runtime/ocr` from electron-builder once catalog entries point at real artifacts
- Self-hosted mirror domain (ops)

## Summary by CodeRabbit

* **New Features**
  * Added downloadable official plugins with progress, cancellation, retry, manual `.dcplugin` installation, and uninstall controls.
  * Added first-use OCR runtime downloads with progress, manual installation, version visibility, cancellation, and removal.
  * Added compatibility and availability indicators for plugins and OCR runtimes.
  * Successful plugin installations can be enabled immediately.

* **Bug Fixes**
  * Plugin enablement now installs missing runtime payloads before retrying.
  * Improved download verification and cleanup for failed or interrupted installations.

* **Documentation**
  * Added guidance for testing remote installations and building without bundled payloads.


zhangmo8 (changes requested):
## Review

I reviewed the full diff (+4887/−79, 65 files), with cross-checks against the main-process sources (`PluginService`, the toolchains downloader, the package-verification path). I'm omitting the `enablePlugin` result-vs-throw issue here since it's already been reported — it still needs fixing before merge.

Overall the security posture of the remote-distribution chain is solid: sha256-pinned artifacts verified before use, zip-slip-safe extraction, renderer input restricted to ids, and the catalog override disabled in packaged builds. The findings below are the ones I didn't see reported yet.

### Findings

1. **major — `src/main/plugin/remoteInstaller.ts:151-155` — plugin-id mismatch is detected after the foreign plugin is already installed and registered.**
   ```ts
   const installed = await this.deps.installPackage(staged.archivePath)
   if (installed.pluginId !== pluginId) {
     throw new Error(`Installed package declares a different plugin id: ${installed.pluginId}`)
   }
   ```
   By this point `installOfficialPluginPackage` has extracted the package into `userData/plugins/<id>` and persisted the installation record (disabled). The installer then reports `error`, leaving an installed-but-untracked-by-UI plugin and an error state on the UI entry — a retry re-downloads and re-errors. The sha256 pin makes this a build-mismatch hazard rather than an attack vector, but the check should run *before* installation (e.g. pass an `expectedPluginId` to `installOfficialPluginPackage` and validate `plugin.json` before extracting), or roll back the installation on mismatch.

2. **minor — `src/main/ocr/runtimeAssetInstaller.ts:211` — whole-archive synchronous unzip on the main process, no decompressed-size cap.**
   `const files = unzipSync(new Uint8Array(fs.readFileSync(archivePath)))` reads the entire OCR payload into memory and decompresses it synchronously. The OCR payload is large (node runtime + models — potentially hundreds of MB compressed, GBs decompressed), so this stalls the main-process event loop for seconds and spikes RSS; a pathological payload could OOM since fflate's `unzipSync` applies no decompressed-size limit (the catalog pins compressed size only). Suggest streaming/async extraction (fflate's async `unzip` + per-entry writes, or a worker) plus a decompressed-bytes sanity cap derived from `target.size`.

3. **minor — `src/main/ocr/runtimeAssetInstaller.ts:174-176` — non-atomic version-directory swap.**
   ```ts
   fs.rmSync(versionDir, { recursive: true, force: true })
   fs.renameSync(extractedDir, versionDir)
   ```
   A crash or force-quit between the `rmSync` and `renameSync` destroys a previously good install of that version (the resolver then reports assets missing). Also, reinstalling the *currently running* version deletes files out from under a live helper process. Suggest renaming the old dir aside (`versionDir` → `versionDir.old-<uuid>`), renaming the new payload into place, then removing the old.

4. **minor — `src/shared/contracts/routes/plugins.routes.ts:18,22` — artifact URLs and mirrors allow plain `http://`.**
   `url: z.url({ protocol: /^https?$/ })` and the mirrors schema accept http, yet the sibling `UserPluginSourceSchema` in the same file (line 162) is https-only, and `assertHttpUrl` in `scripts/plugin-catalog.mjs` also permits http. Integrity is protected by the sha256 pin, but these are executable payloads: on hostile networks plaintext http enables trivial blocking and lets mirror hosts observe every install. Recommend `/^https$/` for both canonical URLs and mirrors (the verify script gets this for free once the schema tightens).

5. **minor — staging roots accumulate orphans after a hard crash (`src/main/lib/remoteArtifactDownload.ts` and both installers).**
   Cleanup is handled in `catch`/`finally`, but a kill/crash mid-download leaves `userData/plugins/.staging/<uuid>/artifact.zip.partial` (or the OCR equivalent) behind forever — repeated crashes accumulate up to payload-size disk waste with no sweep. Cheap fix: at composition time, delete stale uuid dirs under each `.staging` root on startup.

### Positives

- Strong, layered integrity chain: sha256 pinned per target in the app-shipped catalog, verified by `downloadVerifiedFile` before use, staging-dir isolation with deterministic cleanup, and zip-slip-safe extraction — backed by tests including corrupted-checksum rejection and an end-to-end integration test asserting nothing is left behind on failure.
- Clean IPC boundary: the renderer can only pass `pluginId`/`assetId`; URL, mirror, and descriptor resolution stay entirely in main; every new route and event is a Zod-validated contract, and the `DEEPCHAT_PLUGIN_CATALOG` dev override is correctly ignored in packaged builds.
- Careful concurrency/UX wiring: busy/cancel/cooldown handling in the OCR coordinator is well tested, both new renderer components unsubscribe their progress listeners, and the new i18n keys are consistent across all 23 locales.

zerob13:
感谢 @zhangmo8 和 CodeRabbit 的 review，全部意见已处理（b8e5dfd77，6 个 fix commit），本地 typecheck×2 / format / lint / i18n / test:main (8578) / test:renderer (2432) 全绿。

## 人工 review（zhangmo8）

1. **plugin-id mismatch 在安装后才检测（major）** — 已修：`installOfficialPluginPackage` 新增 `expectedPluginId` 参数，在解压/落盘**之前**校验 manifest id；安装器传入并保留事后断言作为纵深防御。新增 L1 集成测试证明 ID 不符时零残留（无目录、无注册记录）。
2. **主进程同步解压 + 无解压尺寸上限（minor）** — 已修：改用 fflate 异步 `unzip`（worker 线程，不阻塞主进程事件循环），并以 zip 头 `originalSize` 在解压前过滤超过 `max(256MiB, 8×压缩尺寸)` 的条目。zip-bomb 测试：260MiB 零填充条目（压缩后 <4MiB）被拒绝且零落盘。
3. **版本目录非原子交换（minor）** — 已修：`swapVersionDirectory` 先把旧目录 rename 到 `<version>.old-<uuid>`，新目录 rename 入位后删除旧目录；中途失败回滚。重装运行中版本不再从活进程脚下删文件。
4. **artifact URL 允许 http（minor）** — 已修：schema 与生成脚本均收紧为 https-only，**唯一豁免是 loopback**（localhost/127.0.0.1/[::1]）——这是 spec §4.4 L1 本地 fixture e2e 流程的需要，且不削弱威胁模型（明文不出本机）。如有异议可再收紧。
5. **崩溃残留 staging 孤儿（minor）** — 已修：启动时（composition init，任何安装开始前）对两个 `.staging` 根执行 `sweepStagingRoot` 清扫。

## CodeRabbit

- **enable 失败走 result 而非 throw（major）** — 确认为真 bug（我的测试 mock 与生产契约不一致导致漏检）。已修：路由层检查 `result.ok`，仅当"未安装且有 catalog 条目"时走远程安装；已安装插件的其他 enable 失败原样透传（新增测试覆盖）。测试 mock 改为返回非 ok 结果。
- **生成 catalog 缺 platform/arch（major）** — 确认为真 bug（生成物会被 `PluginCatalogSchema` 拒绝）。已修：插件目标从文件名推导、OCR 目标取自 manifest，`validateTargets` 校验 platform/arch 枚举；新增回归测试直接用 `PluginCatalogSchema` 解析生成产物。
- **resolver 身份校验在回退循环外（major）** — 已修：`verifyIdentity` 移入每个候选根的循环内，bundled 根身份失败会继续尝试 installed 根；dev 解析保持单次校验。
- **destroy 不排空安装器（major）** — 已修：两个安装器新增 `cancelAll()`（abort 全部 + 等待在飞操作 settle），`destroy()` 第一步排空，之后才关闭 pluginService/ocrRuntimeService。
- **polled 状态较新时 live 优先（minor）** — 已修：computed 取 updatedAt 更新者。
- **页面重挂载丢失安装进度（minor）** — 已修：`loadDistributableCatalog` 用响应内 `installState` 按 pluginId 补水（新者胜出）。
- **优先级测试无法区分顺序（minor）** — 已修：installed fixture 改为完全有效，断言 `helperEntryPath` 落在 unpackedRoot。
- **vue-i18n `{{percent}}`（minor）** — 确认为真 bug（双花括号不是 vue-i18n 具名插值语法）。已修：23 个 locale 的 `installing`/`runtimeInstalling` 全部改为 `{percent}`。
- **dispatcher fixture 缺类型字段（major）** — 已修，另同步修复 pluginCatalogStore 测试的 `OcrRuntimeStatus` fixture 新字段。

zhangmo8 (commented):
## Re-review: all findings verified fixed on `b8e5dfd7`

I re-checked the five findings from my earlier review against the current head. All five are properly fixed:

1. **Plugin-id mismatch after install (major) — fixed.** `installOfficialPluginPackage` (`src/main/plugin/index.ts`) now takes `expectedPluginId` and rejects a mismatching manifest id *before* any extraction or persistence, and `PluginRemoteInstaller.runInstall` passes the expected id while keeping the post-install re-check as defense in depth. The failure path no longer leaves an installed-but-untracked plugin behind.

2. **Synchronous whole-archive unzip with no decompressed-size cap (minor) — fixed.** Extraction moved to fflate's async `unzip` (worker thread, event loop not blocked) with a decompressed-size cap derived from the pinned compressed size (8x ratio, 256 MiB floor), plus manifest/helper-entry structural validation before anything is written into the payload directory.

3. **Non-atomic version-directory swap (minor) — fixed.** `swapVersionDirectory` renames the previous install aside, renames the new payload into place, restores the old directory if the second rename fails, and only then removes the old copy. A crash mid-swap now leaves either the old or the new install intact.

4. **Plain `http://` artifact URLs (minor) — fixed.** `ArtifactUrlSchema` (and the mirrors schema) is now https-only with a documented loopback exception for the local-fixture e2e flow, and `scripts/plugin-catalog.mjs` enforces the same rule, so the producer and consumer sides agree.

5. **Staging roots accumulating orphans after a crash (minor) — fixed.** `sweepStagingRoot` runs at startup for both the plugin `.staging` root and the OCR runtime `.staging` root in `composition.ts`.

No new findings from me. The remaining open items on this head are the ones CodeRabbit posted in its 04:52 review (unguarded startup sweep, cancellation not observed during extraction, the cumulative-vs-per-entry decompressed-size cap, non-fatal backup cleanup, sync plugin install on the main process, the 260 MiB test fixture, and the i18n nits) — I won't duplicate them here, and I agree several of those are worth addressing before merge.

Thanks for the fast turnaround — the fix quality is good.

zhangmo8 (commented):
## Re-review of `7ac42a1c` (cua/ocr unbundle switches)

Checked the new head commit against the earlier findings. The unbundle switches and build-tooling changes are clean (the verify step correctly asserts the unbundled artifact is absent from the app), keying `installed` off runtime-payload presence instead of discovery is the right call, and extracting the shared `RuntimeInstallControls` component removes a lot of duplication. One new finding:

### major — OCR uninstall does not guard against an in-flight install (`src/main/app/composition.ts` uninstall → `OcrRuntimeAssetInstaller.removeInstalled`)

`uninstall()` calls `removeInstalled()` without checking whether an install is running (the shared `running`/`activeOperations` maps are untouched by `removeInstalled()`). If a download or manual install is mid-flight:

- `removeInstalled()` deletes the existing version directories and calls `this.states.clear()`, which wipes the active install's progress state while its `update()` calls keep repopulating it — progress events and UI state flicker back into existence after a successful uninstall response.
- When the install finishes, `swapVersionDirectory` moves the new payload into `installRoot` and reports success, so the runtime is installed again right after the user saw uninstall succeed. There is also a window where `rmSync` on the install root races the install's rename.

Suggested fix: in `uninstall()`, check `ocrRuntimeAssetInstaller.isRunning(LIGHT_OCR_RUNTIME_ASSET_ID)` first — either cancel and await the active operation (would need a public awaitable accessor for `activeOperations`), or return `{ ok: false, reason: 'busy' }` and let the user cancel explicitly. The current `states.clear()` behavior is fine for the quiet path, but it should not run under an active install.

### nit — `enableResolvedPlugin` throws the persisted `runtime.lastError` when the payload is missing

`runtime.lastError ?? "Runtime ... is not installed"` can surface a stale error from a previous failed activation attempt rather than the actual missing-payload condition, which makes the remote-install repair flow report a misleading reason. Prefer the "Runtime X is not installed" message (optionally appending `lastError` as detail).

The rest of the open items on this head are the ones CodeRabbit reported in its 08:52 review (sync I/O in extraction, lexicographic `listInstalledVersions()[0]`, the leftover `RuntimeDependencyRecord` after `uninstallOfficialPlugin`, unverified integrity of manually selected artifacts, and the stale `catalogInstallState` on the detail page) — I agree with those and won't duplicate them here.

zhangmo8 (commented):
## Incremental re-review of `86413bb9` → `2fd73ae9`

Checked the three commits added since my last review (`86413bb9` baseline refresh, the `origin/dev` merge `144e261f`, and `2fd73ae9` baseline re-sync).

- **Baseline refreshes — fine.** Both commits touch only `renderer-application-boundaries-baseline.json`; mechanical, nothing to flag.
- **Dev merge — clean.** Spot-checked the conflict resolution: the PR-side fixes are all still intact at head (`expectedPluginId` rejection before extraction in `plugin/index.ts`, atomic `swapVersionDirectory`, https-only artifact URLs, size-capped async extraction), and the incoming dev changes (clipboard IPC, accessibility event, provider OAuth helper removal) apply cleanly. `composition.ts` changes in the merge are purely dev-side additions.
- CodeRabbit's latest run already covered the baseline-only delta and found nothing actionable, so I have no new findings on the new commits.

### Still blocking: major from the `7ac42a1c` re-review is unaddressed

[The OCR uninstall vs. in-flight install race](https://github.com/ThinkInAIXYZ/deepchat/pull/2305#pullrequestreview-3154329675) remains open — no fix commit and no reply since. Re-confirmed at `2fd73ae9`: `composition.ts` `uninstall()` still calls `ocrRuntimeAssetInstaller.removeInstalled()` without checking `isRunning(LIGHT_OCR_RUNTIME_ASSET_ID)` first, so a mid-flight install can resurrect the runtime right after uninstall reports success (and the progress state flickers back via the still-running `update()` calls).

No new blockers beyond that one. Keeping **changes requested** until the uninstall guard lands (a one-line `isRunning` check returning an error or deferring the removal, plus ideally a test for uninstall-during-install).

zhangmo8 (changes requested):
## Incremental re-review of `e916c9e3` (build(ci): test unbundled cua and ocr distribution)

CodeRabbit has not run on this commit yet (review skipped for drafts), so the findings below are new, not repeats.

### 1. must-fix — CI is red: workflow contract test not updated

`test-main` fails on head:

```
FAIL test/main/build/electronBuilderConfig.test.ts > Linux ARM64 packaging >
  owns runner selection and x64-only CUA behavior in the Linux reusable workflow
AssertionError: expected 'env.DEEPCHAT_UNBUNDLE_OCR != \'1\'' to be undefined
```

The contract test asserts the `Verify packaged Light OCR offline` step has no `if:` condition, but this commit added `if: env.DEEPCHAT_UNBUNDLE_OCR != '1'` (and macOS gained a second gated step). The workflow contract test needs to be updated in the same commit that changes the workflows.

### 2. major (question) — unbundle switches are now hard-coded `'1'` in the shared package workflows

`DEEPCHAT_UNBUNDLE_CUA: '1'` and `DEEPCHAT_UNBUNDLE_OCR: '1'` are set as workflow-level `env` in all three `_package-*.yml` files. That makes unbundled the *default* for every packaging run of these reusable workflows, not just a one-off distribution test — the previous comment ("flip only once the published catalog carries real cua artifact URLs for this version") was removed along with the `'0'` values. Once this reaches `dev`, any release packaging would ship an app without cua/OCR unless someone remembers to flip it back, and a fresh install has no fallback if the published catalog lacks live artifact URLs for that version.

Two suggestions, either is fine:

- Make both switches workflow inputs (default `'0'`) so the unbundled run is opt-in via `workflow_dispatch`, and the verification-only run uses `DEEPCHAT_UNBUNDLE_*: '1'` explicitly; or
- Keep the hard-coded flip but state the revert plan in the PR description (flip back before the next release cut, gated on the catalog carrying real URLs).

Related note: CI now asserts *absence* of the bundled runtime, but nothing in CI exercises the first-run remote-fetch path on a clean machine with no network-independent runtime — worth confirming the L1 e2e flow covers that before this ships for real.

### 3. minor — manifest script reads env vars directly

`package-manifest.mjs` now derives `allowMissingLightOcrReports` / `cuaUnbundled` from `process.env.DEEPCHAT_UNBUNDLE_OCR` / `DEEPCHAT_UNBUNDLE_CUA`, while every other option is an explicit CLI flag. An explicit `--allow-missing-light-ocr-reports` / `--cua-unbundled` flag would be more auditable (visible in the workflow step and in any manifest debugging), and avoids the script silently tolerating missing smoke reports if the env var leaks into a different packaging context.

### Verified good

- `verifyMacZipDistribution` does honor the `verifyCuaMacHelper` override (`package-manifest.mjs:160-201`), so the unbundled ZIP check is real, not mocked-only; and `verifyDmgDistribution` does not inspect `Contents/Helpers`, so no DMG-side swap is needed.
- The new `packageContract.test.ts` cases cover the success path, the helper-still-bundled rejection, and the missing-smoke-report default — good coverage of the new switch surface.
- All `package-*` verification jobs pass on all three platforms with both flips enabled.

### Still blocking

The OCR uninstall vs. in-flight install race from the `7ac42a1c` re-review remains unaddressed — `e916c9e3` does not touch `composition.ts`. That one stays blocking, plus the CI failure above.

zerob13:
这个效果不好，尝试下来收益太多，revert

PR #2223, opened 2026-08-31
feat(provider): add RunInfra support

zerob13:
RunInfra users currently need a custom provider profile. This PR adds a built-in `runinfra` profile with the official workspace API endpoint and logo.

- Reuse `AiSdkProvider` with the OpenAI-compatible Chat Completions transport at `https://api.runinfra.ai/v1`; no new adapter or dependency.
- Discover models and check credentials through authenticated `GET /v1/models`, without generating billable inference. API keys use existing main-process provider storage.
- Add the static official SVG mark with the existing monochrome light/dark theme handling.
- Cover default-profile model discovery, Bearer authentication, and invalid-key errors with a focused contract test.

Validation passed:

- `pnpm run format` and `pnpm run format:check`
- `pnpm run i18n`
- `pnpm run lint`
- `pnpm run typecheck`
- 100 main tests across provider defaults, registration, runtime, factory, reasoning wire, and OpenAI compatibility.
- 37 renderer tests covering provider settings, API configuration, and model icons.
- SVG checked at small sizes in light and dark themes.

```text
BEFORE
Provider catalog -> RunInfra unavailable

AFTER
Provider catalog -> [R] RunInfra
                         |
                         v
Provider settings -> API Key [********]  [Refresh models]
```

API contract: [RunInfra quickstart](https://runinfra.ai/docs/api-reference/model-apis-quickstart) and [model discovery](https://runinfra.ai/docs/api-reference/models).

## Summary by CodeRabbit

* **New Features**
  * Added RunInfra as a supported AI provider.
  * Added RunInfra model discovery through its models endpoint.
  * Added API key authentication and provider connection validation.
  * Added RunInfra branding and icon support.

* **Tests**
  * Added coverage for model discovery, successful authentication, and invalid API key handling.

PR #2314, opened 2026-09-16
feat(provider): add Cheaper Inference

aiapienthusiast:
Adds Cheaper Inference as a built-in, disabled-by-default OpenAI-compatible
provider. Cheaper Inference is a gateway at `https://api.cheaperinference.com/v1`
that serves models from several upstream providers behind one endpoint, using the
model makers' own bare model ids.

It needs no special provider class: the registry entry is `OPENAI_BASE` with
`credentialStrategy: 'api-key'` and `embeddingStrategy: 'none'`, so it resolves
to the existing `openai-compatible` runtime with `modelSource: 'openai'` and
`checkStrategy: 'fetch-models'` — the same shape as API Route and RunInfra.

## Files changed

| File | Change |
|---|---|
| `src/main/provider/defaults.ts` | `DEFAULT_PROVIDERS` entry (`enable: false`) |
| `src/main/provider/providerRegistry.ts` | `PROVIDER_ID_REGISTRY` definition |
| `src/renderer/src/assets/llm-icons/cheaper-inference.svg` | new icon |
| `src/renderer/src/components/icons/modelIconRegistry.ts` | icon import + `modelIcons` key |
| `test/main/provider/defaultProviders.test.ts` | asserts the defaults entry |
| `test/main/provider/basicApiKeyProviders.test.ts` | asserts the resolved definition |

Insertions only, 68 lines, no existing line modified.

`resources/model-db/providers.json` is deliberately not touched — per AGENTS.md
("Keep provider and ACP registry refreshes produced by normal builds") it is
regenerated by `scripts/fetch-provider-db.mjs` from `PublicProviderConf` on every
`prebuild`. This provider uses live `/models` discovery, so it needs no
provider-db entry; if its display metadata should be shared, that belongs in
`PublicProviderConf`, not here.

## UI

The only visible change is one more row in Settings → Model Provider, off by
default:

```
BEFORE                          AFTER
┌────────────────────────┐      ┌────────────────────────┐
│ ▣ Ollama               │      │ ▣ Ollama               │
│ ▣ Deepseek             │      │ ▣ Deepseek             │
│ ▣ API Route            │      │ ▣ Cheaper Inference  ← │
│ ▣ RunInfra             │      │ ▣ API Route            │
│ ...                    │      │ ▣ RunInfra             │
└────────────────────────┘      │ ...                    │
                                └────────────────────────┘
```

## Precedent

Same six-file shape as `87b9fc5e` *feat(provider): add API Route (#2269)*, and
the same registry definition as that entry and `04451d75`
*feat(provider): add RunInfra support (#2223)*.

## Testing

Run locally on this branch:

| Command | Result |
|---|---|
| `pnpm vitest run --config vitest.config.ts test/main/provider/defaultProviders.test.ts test/main/provider/basicApiKeyProviders.test.ts` | 33 passed |
| `pnpm vitest run --config vitest.config.ts test/main/provider` | 706 tests passed |
| `pnpm vitest run --config vitest.config.renderer.ts test/renderer/components/ModelIcon.test.ts` | 6 passed |
| `pnpm run format:check` | clean, 3006 files |
| `pnpm run lint` | clean (both guards + oxlint) |
| `pnpm run typecheck:web` | clean |
| `pnpm run i18n` | clean, no new keys needed |
| `pnpm run icons:check` | clean |
| `pnpm run architecture:renderer-baseline:check` | current |

Not run, and why:

- `pnpm run typecheck:node` reports one error, `Cannot find module 'xlsx'` in
  `src/main/file/adapters/ExcelFileAdapter.ts`. `xlsx` is fetched from
  `cdn.sheetjs.com`, which my network could not reach, so the package is absent
  from my `node_modules`. The same missing package also stops four unrelated
  test files in `test/main/provider` from importing (via `src/main/file/mime.ts`).
  Both are install artifacts, not code changes — no file in this PR imports `xlsx`.
- The full `pnpm run test:renderer` did not finish: individual tests were taking
  10–110 s each on my machine and hitting their timeouts across unrelated suites
  (`McpServers`, `MarkdownRenderer`, `MemorySettings`, …). `ModelIcon.test.ts`,
  the only renderer test covering this change, passes in isolation.
- `pnpm run build` and the `test-native-memory` job were not run.

## Summary by CodeRabbit

- **New Providers**
  - Added Cheaper Inference as a built-in OpenAI-compatible provider.
  - The provider is disabled by default and supports API-key authentication and model discovery.
  - Embeddings are not available for this provider.

- **User Interface**
  - Added a dedicated Cheaper Inference icon for provider and model displays.

- **Tests**
  - Added coverage validating provider configuration, authentication, model discovery, and default settings.


src/main/provider/defaults.ts
```diff
@@ -31,6 +31,21 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
...
+  {
+    id: 'cheaper-inference',
+    name: 'Cheaper Inference',
+    apiType: 'openai-completions',
+    apiKey: '',
+    baseUrl: 'https://api.cheaperinference.com/v1',
+    enable: false,
+    websites: {
+      official: 'https://cheaperinference.com/',
+      apiKey: '',
```
zhangmo8, line 43: nit: every other third-party provider here links to a key management page (`deepseek`, `api-route`, `runinfra`, `greenpt` all have a non-empty `websites.apiKey`). If Cheaper Inference has a dashboard/keys page, it would be nice to link it for consistency; otherwise leaving it empty is fine.

src/main/provider/defaults.ts
```diff
@@ -31,6 +31,21 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
...
   },
+  {
+    id: 'cheaper-inference',
+    name: 'Cheaper Inference',
+    apiType: 'openai-completions',
+    apiKey: '',
+    baseUrl: 'https://api.cheaperinference.com/v1',
+    enable: false,
+    websites: {
+      official: 'https://cheaperinference.com/',
```
zhangmo8, line 42: nit: `https://cheaperinference.com/` (and the `/docs` link) 308-redirect to the canonical `https://www.cheaperinference.com/...` URLs. Consider using the final URLs directly to save a hop, though this is purely cosmetic.

zhangmo8 (commented):
LGTM overall — this follows the API Route / RunInfra precedent exactly (disabled-by-default, `OPENAI_BASE` + `credentialStrategy: 'api-key'`, live `/models` discovery), the tests mirror the existing ones, and the tests-in-PR claim checks out against the diff. I verified that `https://www.cheaperinference.com/` is live and `https://api.cheaperinference.com/v1/models` correctly returns 401 without a key, which is consistent with the `fetch-models` check strategy. Two non-blocking nits below; happy to approve once addressed (or as-is if you prefer).

src/main/provider/defaults.ts
```diff
@@ -31,6 +31,21 @@ export const DEFAULT_PROVIDERS: LLM_PROVIDER_BASE[] = [
...
+      apiKey: '',
+      docs: 'https://cheaperinference.com/docs',
+      models: 'https://api.cheaperinference.com/v1/models',
```
zhangmo8, line 45: nit: `websites.models` points at the raw API endpoint `/v1/models`, which returns authenticated JSON rather than a browsable page (API Route and RunInfra link to human-readable pages here). If the docs have a public model list page, that would be a friendlier target for the Settings link; fine as-is otherwise.

zhangmo8 (approved):
LGTM overall. This follows the established six-file shape from #2269 (API Route) and #2223 (RunInfra): registry entry is exactly the `OPENAI_BASE` + `credentialStrategy: 'api-key'` + `embeddingStrategy: 'none'` shape, disabled by default, insertions only, and all three registration points for a built-in provider (defaults, `PROVIDER_ID_REGISTRY`, `modelIconRegistry`) are covered with matching tests. Nothing else in the codebase needs to know about a new built-in provider id, so this is complete.

One small non-blocking nit inline below.

PR #2315, opened 2026-09-17
feat(agent): kernel extraction and contract freeze

zerob13:
## Summary

This branch delivers the repository and ownership groundwork for a portable DeepChat agent service. It is intentionally a **structural extraction and contract-freeze PR for stages 0–2**, not the standalone service product itself.

The current and future execution paths are deliberately separate:

```text
Current
CLI local-control client ───────> Desktop-owned local-control server

Desktop composition
  ├─> built-in host adapters ──> @deepchat/agent-kernel ──> @deepchat/shared
  │                              └─ provider / MCP portable cores
  └─> direct ACP backend ──────> external ACP agent (owns its own loop)

Future — not implemented by this PR
CLI service-host composition ──> kernel + domain cores + headless host adapters
Desktop / CLI clients ────────> standalone service binding
Direct ACP remains a separate peer path
```

The branch preserves the current Desktop compatibility path. It does not add a standalone service process, a new service transport or authentication implementation, a Desktop/CLI cutover, or a new UI. To be precise: the Stage 1 contract vocabulary and the CLI local-control client changes do exist in this PR; what does not exist is the future service's host process, transport, and authentication. Those belong to later standalone-service stages.

## Scope at a glance

- **86 commits** relative to `origin/dev` (at the time of writing)
- **4,720 files changed**: `+76,466 / -25,342`
- Branch: `architecture/standalone-agent-harness`
- PR target: `dev`
- Current HEAD: `777bbac5b` — includes periodic merges of `dev`; the extraction-specific content is under "What changed"
- Required toolchain: Node `>=24.18.0 <25`, pnpm `>=10.34.5 <11`

The large file count is primarily a deliberate monorepo relocation and ownership change: Desktop sources/tests/resources move under `packages/desktop`, while shared contracts, CLI, provider, MCP, and kernel code become explicit workspace package surfaces. With git's default rename detection, roughly 4,400 of the ~4,720 changed paths are pure renames; only about 314 paths add, modify, or delete content. This is not a product-feature diff of the same size — but it is not path-only either: provider failure semantics, MCP diagnostics redaction and process handling, ACP cancellation and bounded shutdown, and the CLI build/run/launcher loop all changed. The review-focus section below lists exactly where the new logic lives.

### 1. Standalone architecture and contract freeze — Stages 0–1

Added the normative standalone-agent architecture, baseline, compatibility mapping, and execution tracker under:

- `docs/architecture/standalone-agent-harness/spec.md`
- `docs/architecture/standalone-agent-harness/plan.md`
- `docs/architecture/standalone-agent-harness/baseline.md`
- `docs/architecture/standalone-agent-harness/compatibility.md`

The client-facing contract is deliberately serializable and host-neutral. It covers the initial vocabulary for:

- typed handshake and complete capability advertisement;
- session snapshot and bounded event subscription;
- epoch/sequence cursor replay and explicit resync behavior;
- submission identity, duplicate detection, receipt lookup, and retention semantics;
- interaction/approval request-response DTOs;
- separate cancellation layers for queued submission, running run, and active turn;
- owned artifact references;
- bounded structured errors;
- explicit `capability_unavailable` behavior and compatibility mapping to the maintained local-control CLI.

The public DTOs do not carry Electron objects, callbacks, `AbortSignal`, provider clients, database connections, principals, renderer handles, absolute paths, or runtime class instances. `typecheck:contracts` compiles the contract assertions so these boundaries are enforced rather than documented only.

The Stage 1 Agent Service DTO/client contracts currently remain under `packages/desktop/src/shared/contracts/agent-service`. Promoted local-control wire contracts live in `packages/shared`. Host-neutral contract design does not imply service wiring or publication through `@deepchat/shared`.

The current client contract is intentionally incomplete: live delivery, session lifecycle, steering, interruption, pending-input queue operations, and compatibility mapping are explicit hand-offs to later slices rather than claims that a final service API already exists. A missing or pruned receipt is also not evidence that execution never started; a binding that cannot retain receipts reports `receipt_not_retained`, and `not_found` does not authorize blind resubmission.

### 2. ACP ownership and lifecycle seams

The direct ACP backend is separated from the built-in Desktop agent ownership model:

- ACP session state is accessed through a neutral `SessionStatePort`/`AcpSessionStateAdapter` instead of hydrating built-in scope.
- ACP sessions own their status publication, transcript/skills routing, and destruction path.
- The compatibility factory no longer makes ACP look like a built-in session internally.
- Provider settings are sanitized before use and remain the update base.
- ACP prompt cancellation uses the **caller-owned active prompt signal** all the way into projection/runtime helpers; no replacement `AbortController` is created.
- `closeAll()` has bounded shutdown phases. An uncooperative peer cannot block shutdown forever; pending work is detached after the phase timeout and runtime maps are cleared.

The direct ACP backend continues to use its own `AcpAgentRuntime`. External ACP does not execute through the built-in `@deepchat/agent-kernel` loop. Existing Desktop and direct ACP flows retain their respective host seams; these changes close ownership and lifecycle leaks without merging the two runtimes.

### 3. Kernel extraction into `@deepchat/agent-kernel`

The portable built-in agent kernel is now a private workspace package:

- package: `packages/agent-kernel`
- package name: `@deepchat/agent-kernel`
- explicit public subpath exports;
- loop/runtime/memory/resources/instance/contracts/composition and neutral collaborators owned by the package;
- generated JavaScript and declarations emitted with plain per-file TypeScript build;
- forbidden-specifier scan prevents Electron, SQLite, PTY, repository aliases, and host paths from leaking into the artifact.

The old host re-export layer was removed rather than preserved as an invisible public API. Production callers now use supported package exports. The kernel depends on explicit structural ports instead of `Pick<HostClass>` or Desktop implementation types.

The package is embedded by Desktop through the workspace link, so the current Desktop path still runs the same kernel implementation. A future service host can consume the package without importing Electron, but that host composition is not implemented by this PR.

### 4. Shared package extraction — `@deepchat/shared`

The reviewed neutral closure is now owned by `packages/shared`:

- package name: `@deepchat/shared`;
- exact exported subpaths with manifest-derived aliases;
- shared schemas, value modules, event contracts, route contracts, model/provider types, and portable utilities;
- browser-safe closure for promoted modules;
- no shared → kernel/domain/Desktop dependency;
- compiler-symbol identity regression prevents accidental `any`/`unknown` substitutions, including the `MessageFile` barrel case.

The package boundary is based on actual consumers, not a speculative extraction of all historical `src/shared` files.

### 5. Desktop relocation into a normal workspace package

Desktop is now a workspace member at `packages/desktop` while retaining its product identity (`name: DeepChat`):

- Desktop source, tests, resources, build output, Electron/Vite/Builder configuration and app-local scripts move together;
- repository root becomes a private workspace/tooling package;
- root lockfile and pnpm policy remain authoritative;
- config-relative paths replace caller-cwd assumptions;
- Desktop outputs remain app-local (`packages/desktop/out/{main,preload,renderer,cli}`) and Builder staging remains under Desktop;
- installed artifact layout and app identity are preserved;
- root and filtered commands use the same configuration/resources/output contract.

No UI behavior change is intended. Desktop still owns real application composition, ACP lifecycle, built-in host adapters, local-control server behavior, and the Desktop-owned Agent Service contract gate. The relocation is therefore not only a path/config move; those runtime boundaries remain part of the behavior review.

### 6. CLI package and launcher boundaries

Added `packages/cli` as the explicit CLI client/launcher package boundary:

- exported launcher artifacts and global `--help`/`--version` behavior;
- explicit Desktop version input rather than incidental root-manifest discovery;
- package-local build and declaration boundaries;
- installed launchers require the bundled Electron runtime and fail closed without a system-Node fallback;
- current local-control protocol remains mapped and compatibility-preserved.

The current CLI bundle itself can be executed with system Node during development after it has been built. The installed-launcher runtime-discovery rule must not be confused with a general Electron-only restriction on the bundle:

The current local development loop is deliberately explicit:

- `pnpm run cli:build` bundles `packages/desktop/out/cli/deepchat.mjs`, using the Desktop version;
- `pnpm run cli:run -- --help` and `pnpm run cli:run -- profile list` execute that bundle with plain Node and forward its arguments;
- `packages/desktop/scripts/run-cli.mjs` fails fast when the bundle is missing, strips the nested `--` forwarding marker, and propagates the child exit code;
- system Node can execute the development bundle, but local-control operations still require a running Desktop with its local control plane available.

The preceding watch-mode implementation was replaced by this build-then-run loop in the latest commit. This branch does **not** turn the CLI into the standalone service. Stage 3 service-host composition, profile ownership, headless credential/OAuth ownership, and Desktop/CLI cutover remain future work.

### 7. Provider and MCP portable cores

Extracted provider and MCP runtime code behind explicit ports:

- provider runtime receives config, credentials, OAuth, and related host capabilities through ports;
- standalone completion now fails fast by default instead of converting provider/auth/network failures into an indistinguishable empty string;
- swallowing is explicit opt-in and request diagnostics contain provider/model metadata rather than prompt contents or raw request messages;
- MCP transport/session/process code is isolated behind host capabilities and bounded stdio/process behavior;
- MCP package is explicitly Node-only and has a real isolated stdio/HTTP artifact-consumer gate;
- regular root build now includes MCP in the topological build chain.

MCP diagnostics are intentionally split by boundary:

- default logs, renderer-facing diagnostics state, and failed status events carry a fixed error category;
- raw errors still propagate to callers and protocol responses by contract, so callers can decide their own handling.

### 8. Build, test, CI, and artifact gates

Added or hardened the package and repository gates:

- manifest-derived package/source aliases;
- cold source typechecks for provider and MCP;
- shared → provider/MCP → kernel/consumer build ordering;
- serialized artifact writers to avoid shared `dist` races;
- isolated declaration/runtime consumer checks;
- real staged-artifact MCP consumers over both stdio and HTTP;
- kernel port structure type gate;
- agent-service contract type gate;
- forbidden-import and declaration-closure checks;
- package-impact, package-size, release-preflight, native path, and relocated CI checks;
- config-relative plugin bundle/package paths and app-root clean action.

These gates are intended to prove both sides of each package boundary:

1. clean source development/typecheck without relying on stale build output;
2. isolated declaration/runtime consumption from staged artifacts without repository aliases or workspace hoisting assumptions.

`typecheck:contracts` is static boundary evidence. It does not by itself prove that a service transport, authentication layer, or standalone host has been implemented.

### 9. Provider/ACP registry refresh and provenance

The branch retains the generated provider and ACP registry refresh required by the normal build flow.

The provider refresh was audited against:

- upstream: `ThinkInAIXYZ/PublicProviderConf`
- branch: `dev`
- immutable audit commit: `e67c827216525474...`

The ten model removals from the carried refresh were independently verified as absent from upstream at that commit. The sanitizer now preserves a future upstream `updated_at` as `source_updated_at` so subsequent generated snapshots can carry source provenance.

The current checked-in `providers.json` predates that metadata change and does not yet contain `source_updated_at`. This is a non-blocking P2 follow-up for the next registry refresh, not a runtime compatibility issue.

The registry refresh is a generated-asset change carried by the Desktop `prebuild` flow. Review it as generated output with provenance, separately from the architecture changes.

### Current Desktop built-in path

```text
Before: Desktop -> root-owned src/main/src/shared implementation graph
After:  Desktop composition -> Desktop host adapters
                              -> @deepchat/agent-kernel
                              -> @deepchat/shared
```

- Same Desktop product identity and installed layout.
- Same main/preload/renderer ownership.
- No new service process.
- No new UI.
- No migration of credentials, database schema, profile layout, or encryption keys.

### Direct ACP path

```text
Before/after: Desktop direct ACP backend -> AcpAgentRuntime -> external ACP agent
```

The direct ACP runtime remains a separate peer path; it is not routed through the built-in kernel loop.

### ACP cancellation

```text
Before: active prompt signal -> compatibility projection creates a different signal
After:  active prompt signal -> ACP projection -> runtime helpers observe the same signal
```

Cancellation, timeout, and peer-exit now reach the projection that owns the work.

### Provider failure

```text
Before: provider/auth/network failure -> empty completion string by default
After:  provider/auth/network failure -> propagated error by default
        explicit swallowErrors:true -> intentional fallback, redacted diagnostics
```

### MCP diagnostics

```text
Remote/local Error
      ├── renderer-facing diagnostics state -> error category only
      ├── failed status event             -> error category only
      └── default console diagnostics    -> operation/server/category only

Caller/protocol response -> existing propagation semantics remain unchanged
```

### Shutdown

```text
Before: closeAll waits indefinitely for an unresponsive ACP peer
After:  bounded phase timeout -> detach uncooperative work -> clear runtime maps
```

## What this PR does not do

These are intentionally outside this branch:

- no standalone service process or daemon;
- no service-host composition in CLI;
- no new transport/authentication implementation;
- no Desktop/CLI cutover;
- no headless credential/OAuth/profile owner or `safeStorage` compatibility strategy;
- no TUI, scrollback UI, interactive shell, or Desktop-dependent plugin support;
- no credential-format, database-schema, profile-layout, or encryption-key migration;
- no new UI/product feature.

The structural implementations and contract slices are present, with local and independently reviewed evidence recorded below. This does **not** close all monorepo acceptance gates: applicable cross-platform packaging/native and launcher checks remain pending or environment-dependent. Standalone Stages 3–7 are separately unimplemented. In particular, the recorded final package evidence is unsigned macOS arm64; non-macOS targets, DMG/ZIP publication, signing, and update flows are not proven by that evidence.

## Verification

> **Revision scope:** the checks below were run on the reviewed pre-CLI-workflow revision. Everything pushed after it — the CLI workflow commits (`ced528c5b` watch-mode dev build, superseded by `dab4b3b15` local run command), the CI packaged-smoke path anchoring (`3ed57f925`), the periodic `dev` merges (`252d972db`, `777bbac5b`), and the native-containment build fixes that followed — is covered by the scope statistics above but **not** by that evidence; rely on CI or a fresh local run for those paths.

### Previously reviewed revision evidence

On the reviewed pre-watch revision, using Node `v24.18.0` / pnpm `v10.34.5`:

- `pnpm run typecheck`
  - shared, provider, MCP, CLI, Desktop node/web/contracts/kernel-port gates;
- `pnpm run build:mcp`;
- `pnpm run test:mcp:artifact`
  - isolated declaration/runtime gate plus real stdio and HTTP transport fixtures;
- targeted MCP/Desktop tests: **79 passed**;
- MCP runtime tests: **3 passed**;
- `pnpm run format:check`;
- `pnpm run i18n:validate`;
- `pnpm run lint` — 0 errors, existing warnings only;
- `git diff --check` — clean for that reviewed revision.

### Historical whole-branch evidence — not current-HEAD claims

The architecture plans record broader evidence from the migration work, including main-process and renderer suites, the root build, and a specified unsigned macOS arm64 packaged smoke (**658 main files / 9,252 tests**; **278 renderer files / 2,516 tests**). Treat these as historical, path-scoped records; they were not re-run at the current HEAD.

### Suggested checks for the current HEAD

```sh
pnpm run typecheck          # shared/provider/mcp/cli + desktop node/web/contracts/kernel-ports
pnpm run build:mcp
pnpm run test:mcp:artifact  # staged-artifact MCP gate
pnpm run test:mcp           # + @deepchat/mcp runtime tests
pnpm run build:kernel
pnpm --filter @deepchat/agent-kernel run test   # clean-consumer kernel gate (real Node child process)
pnpm run format:check
pnpm run i18n:validate
pnpm run lint
pnpm run cli:build && pnpm run cli:run -- --help   # current-HEAD CLI loop
```

Within the aggregate `pnpm run test:main` (projects `main`/`kernel`/`shared`/`mcp`/`artifact`), prioritize `packages/agent-kernel/test/kernelPackage/agentKernelPackageRuntime.test.ts`, `packages/desktop/test/main/contracts/agentServiceClientContract.test.ts`, `packages/desktop/test/main/agent/acp/instance/*`, and `packages/desktop/test/main/provider/{providerRuntime,providerPackage,providerFailure}.test.ts`. `pnpm run cli:run -- profile list` requires a running Desktop local-control server; it is not a standalone-service test.

The per-stage acceptance evidence, exact command history, known limitations, and future blockers remain recorded in:

- `docs/architecture/standalone-agent-harness/plan.md`
- `docs/architecture/monorepo-migration/plan.md`
- `docs/architecture/standalone-agent-harness/baseline.md`
- `docs/architecture/standalone-agent-harness/compatibility.md`

## Review focus: where the new logic lives

Review the non-rename paths first (`git diff --find-renames origin/dev...HEAD`); do not walk moved test/resource files one by one unless a path-sensitive failure appears. The groups below are where new logic, new contracts, or changed behavior actually live, in priority order.

### P0 — architecture invariants and package ownership

- `docs/architecture/standalone-agent-harness/spec.md` — the two execution paths; direct ACP stays a peer runtime and is never routed through the built-in kernel; no capability or approval bypass for headless clients.
- `docs/architecture/monorepo-migration/spec.md` — dependency direction `desktop → kernel → shared`; no reverse or cross-package `../../src` imports; relocation vs installed-artifact compatibility.
- `packages/shared/package.json`, `packages/agent-kernel/package.json`, `packages/provider/package.json`, `packages/mcp/package.json`, `packages/cli/package.json` — exports, `files`, dependency edges. The kernel's wildcard subpath exports make its surface broad; nothing host-only (Electron, SQLite, PTY) may enter the kernel/MCP artifacts.

### P0 — kernel boundary and clean-consumer proof

- `packages/agent-kernel/src/index.ts`
- `packages/agent-kernel/src/composition/createDeepChatRuntimeServices.ts`
- `packages/agent-kernel/src/loop/ports.ts`
- `packages/agent-kernel/src/runtime/deepChatLoopRunner.ts`
- `packages/agent-kernel/src/runtime/dispatch.ts`
- `packages/agent-kernel/src/tape/ports/application.ts`

Check: the kernel depends on structural ports, not Desktop implementation types; a tool continuation completes without a client callback between provider rounds; transcript/Tape/terminal outcomes commit exactly once; authorization, cancellation, and pending input stay host-neutral.

Primary executable evidence: `packages/agent-kernel/test/kernelPackage/agentKernelPackageRuntime.test.ts` — it stages the built package, installs a clean consumer, runs a real Node child process through a two-round tool continuation, and compiles the declaration closure externally. This is the single most valuable regression gate in the PR.

### P0 — Agent Service contract boundary

- `packages/desktop/src/shared/contracts/agent-service/common.ts`
- `packages/desktop/src/shared/contracts/agent-service/client.ts`
- `packages/desktop/src/shared/contracts/agent-service/events.ts`
- `packages/desktop/src/shared/contracts/agent-service/interactions.ts`
- `packages/desktop/test/main/contracts/agentServiceClientContract.test.ts`
- `scripts/typecheck-agent-service-contracts.mjs`

Check: strict DTOs with bounded text/error fields; no Electron objects, callbacks, `AbortSignal`, provider clients, database handles, principals, renderer identity, or absolute paths cross the boundary; capability advertisement cannot hand a headless client Desktop authority; receipt/idempotency semantics do not invite blind resubmission. The type gate compiles one test plus the contract directory — deliberate, but not a repository-wide proof.

### P0 — Desktop composition and direct ACP lifecycle

- `packages/desktop/src/main/app/composition.ts` — Desktop still supplies every real host adapter; watch for an accidental second provider/session/approval authority and for lifecycle ordering (shutdown, database reopen, plugin/MCP init, route registration).
- `packages/desktop/src/main/agent/deepchat/harness/createDeepChatAgentHarness.ts`
- `packages/desktop/src/main/agent/deepchat/harness/runtimeServices.ts`
- `packages/desktop/test/main/app/compositionBoundaries.test.ts`
- `packages/desktop/src/main/agent/manager/directAcpAgentBackend.ts`
- `packages/desktop/src/main/agent/acp/instance/acpAgentRuntime.ts`
- `packages/desktop/src/main/agent/acp/instance/acpSessionStateAdapter.ts`
- `packages/desktop/src/main/agent/acp/instance/ports.ts`
- `packages/desktop/src/main/agent/acp/compatibility/dependencies.ts`
- `packages/desktop/test/main/agent/acp/instance/acpAgentRuntime.test.ts`
- `packages/desktop/test/main/agent/acp/instance/acpSessionStateAdapter.test.ts`

Check: the caller-owned prompt `AbortSignal` reaches projection; `closeAll()` phases are bounded and runtime maps are cleared; ACP state never hydrates built-in scope; provider settings are sanitized before use.

### P1 — provider and MCP behavior changes (behavioral, not relocation)

- `packages/provider/src/runtime.ts`
- `packages/provider/src/providers/aiSdkProvider.ts`
- `packages/provider/src/managers/providerInstanceManager.ts`
- `packages/provider/src/ports.ts`
- `packages/desktop/test/main/provider/providerRuntime.test.ts`
- `packages/desktop/test/main/provider/providerPackage.test.ts`
- `packages/desktop/test/main/provider/providerFailure.test.ts`
- `packages/mcp/src/client.ts`
- `packages/mcp/src/manager.ts`
- `packages/mcp/src/ports.ts`
- `packages/mcp/src/stdio.ts`
- `packages/mcp/src/diagnostics.ts`
- `packages/mcp/test/runtime.test.ts`
- `scripts/build-mcp.mjs`
- `scripts/check-mcp-artifact.mjs`
- `scripts/mcp-artifact-consumer.mjs`

Check: standalone completion fails fast by default and the empty-string fallback requires explicit `swallowErrors: true`; request diagnostics carry provider/model metadata but never prompt contents; MCP stdio abort/termination cleanup, HTTP header handling, and the category-vs-raw-error diagnostic split.

### P1 — CLI build/run and launcher

- `packages/cli/src/index.ts`
- `packages/cli/src/args.ts`
- `packages/cli/src/run.ts`
- `packages/cli/src/transport.ts`
- `packages/cli/src/discovery.ts`
- `packages/cli/src/launcher.mjs`
- `packages/cli/scripts/build.mjs`
- `packages/desktop/scripts/build-cli.mjs`
- `packages/desktop/scripts/run-cli.mjs`
- `docs/guides/cli.md`

Check: build-then-run development loop (no watch mode on the current HEAD); installed launchers require the bundled Electron runtime and fail closed without a system-Node fallback; the version comes explicitly from the Desktop package; argument forwarding, `--` stripping, exit codes, and platform quoting.

### P1 — relocation, packaging, CI

- `pnpm-workspace.yaml`
- root `package.json` and `pnpm-lock.yaml`
- `packages/desktop/electron.vite.config.ts`
- `packages/desktop/electron-builder.yml`
- `packages/desktop/tsconfig.node.json`
- `packages/desktop/tsconfig.app.json`
- `packages/desktop/vitest.config.ts`
- `.github/workflows/prcheck.yml`
- `.github/workflows/package-check.yml`
- `.github/workflows/_package-linux.yml`
- `.github/workflows/_package-macos.yml`
- `.github/workflows/_package-windows.yml`

Check: root scripts delegate to the owning package; build order is shared → provider/MCP → kernel → Desktop/CLI; config-relative paths work from both root and package directories; package-impact classification reacts conservatively to the mass relocation; packaged smoke paths are anchored at the repo root; native module resolution stays contained in the workspace tree (the recent `fix(build)` chain).

### Generated assets (review separately from architecture)

- `packages/desktop/resources/model-db/providers.json` and the ACP registry refresh — generated output; see the provenance section above.

## Reviewer reading order

1. Start with `docs/architecture/standalone-agent-harness/spec.md` (the two execution paths and the prohibition on merging direct ACP into the built-in loop), then `docs/architecture/monorepo-migration/spec.md` (package ownership and dependency constraints).
2. Then work through the review-focus groups above in order; they encode the same reading order with concrete files.
3. Use the latest evidence and exit-gate notes in both plans instead of treating the 4,720-file relocation as one undifferentiated runtime change.

## Review status

This PR has completed the intended structural extraction and contract slices for stages 0–2, with the boundaries and known observations recorded above. It should not be read as claiming that the standalone service host, all cross-platform packaging/native gates, or the final Desktop/CLI cutover already exist. The remaining standalone host work and any unclosed monorepo exit gates are explicitly tracked as separate follow-up work.


zhangmo8 (commented):
Reviewed at `777bbac5bc950f4ff5b29d518e0e03e4433a090d` against `origin/dev` (`074890266`). I did not read the 2,891 renamed files. I reviewed the parts that change behavior — the new package boundaries and the gates that are supposed to enforce them, the ACP ownership/lifecycle seams, the provider and MCP portable cores, the Desktop relocation's path and packaging consequences, and the CLI boundary — and checked each claim in the description against the code.

Four findings below are worth acting on. The two most severe are inline.

## 1. Deleting an ACP conversation no longer deletes its durable rows (inline)

`state.destroySession` now early-returns for ACP sessions (`packages/desktop/src/main/app/composition.ts:2159`), but nothing else in the deletion path removes the shared durable state.

In `dev` this same call ran `SessionLifecycleCoordinator.destroy`, which unconditionally deletes transcript, pending inputs, the session settings row and tape, releases compaction and programmatic-parent state, clears the conversation tool mapping and finishes memory destroy (`packages/agent-kernel/src/runtime/sessionLifecycleCoordinator.ts:133-157`). What is left for ACP at this head is:

- `cleanupSessionBackends` → `harness.cleanupSession` → the kernel `cleanup`, which is documented as "Releases runtime state without deleting durable session facts" (`sessionLifecycleCoordinator.ts:107-112`);
- `directAcpAgentBackend.cleanupSession` → `deleteDurableSession` → `acpSessionPersistence.deleteSessions` → `DELETE FROM acp_sessions` (`packages/desktop/src/main/agent/data/tables/acpSessions.ts:217`) — the ACP agent DB only;
- `appSessionService.delete`, which removes only the metadata row, search documents and the `new_sessions` row (`packages/desktop/src/main/agent/shared/appSessionService.ts:219-223`).

There is no FK cascade to fall back on — no `REFERENCES`/`ON DELETE CASCADE` exists on the session tables — and both backends write to the same `sessionData.transcript`. So the conversation disappears from the UI while its messages, assistant blocks, user-message files and links, pending inputs, settings row, active-skill and disabled-tool rows and tape entry remain permanently. This needs an ACP-side durable teardown equivalent to the built-in `destroy`, rather than a skipped stage.

## 2. A generating ACP session is now assessed as idle before a transfer (inline)

`handle.snapshot({lightweight: true})` for ACP returns `AcpSessionStateAdapter.getSessionListState`, a hardcoded `status: 'idle'` placeholder (`packages/desktop/src/main/agent/acp/instance/acpSessionStateAdapter.ts:62-67`), while the non-lightweight path does read the live instance status (`packages/desktop/src/main/agent/manager/directAcpAgentBackend.ts:53-70`).

`SessionAssignment.assessTransferSession` uses the lightweight form specifically for ACP (`packages/desktop/src/main/session/assignment.ts:508-510`), and `prepareSessionForTransfer` returns early unless the status is `generating` (:554). In `dev` the ACP backend received `sessionState: deepChatAgentHarness` (`origin/dev:src/main/app/composition.ts:1940`), whose `getSessionListState` reads the live summary — so the same call site used to return a real status. Now the pre-transfer cancel is skipped for a running ACP turn and the stop-poll at :563-566 can never observe `generating`, so a session can be reassigned while its prompt is still executing.

## 3. The sharp/platform step now writes the wrong workspace file

`packages/desktop/scripts/install-sharp-for-platform.js:56` resolves `pnpm-workspace.yaml` relative to `cwd`, and this file is a content-identical rename of the old root `scripts/install-sharp-for-platform.js`. The invocation changed, though: root `package.json:53` now delegates with `pnpm --filter DeepChat run install:sharp`, and pnpm runs filtered package scripts with `cwd` set to the package directory.

`packages/desktop/` has no `pnpm-workspace.yaml`, so `existsSync` is false and the step creates a fresh file containing only `supportedArchitectures` inside `packages/desktop/`, leaving the real root workspace file untouched. CI then runs `pnpm install` immediately afterwards (`.github/workflows/_package-macos.yml:128`, `_package-linux.yml:89`, `_package-windows.yml:94`), so the target-platform optional deps are not installed for the build target. Runners currently mask this because runner arch equals target arch on all three platforms; a local cross-arch build (`build:mac:x64` on Apple Silicon, `build:linux:arm64` on x64) will package host-arch natives. Two secondary effects: an untracked `packages/desktop/pnpm-workspace.yaml` appears inside the app directory and is not excluded by `packages/desktop/electron-builder.yml:22`, and the workaround at `packages/desktop/scripts/afterPack.js:126-137` exists only to tolerate it. Anchor the path to the workspace root and fail loudly when it is missing.

## 4. MCP start failures are now unactionable, and the replacement channel the comment promises does not exist

`packages/mcp/src/manager.ts:84` stores `errorCategory(error)` as the last error; on `dev` the same setter stored `error.message` (`src/main/mcp/serverManager.ts:78-81`). That value is the only source of `McpServerDiagnostics.lastError` (`packages/desktop/src/main/mcp/index.ts:1158-1179`) and reaches the server-card tooltip, so a failed start now displays literally `Error` / `TypeError` / `SpawnError` — the user loses the reason (missing env var, spawn ENOENT) with nothing to replace it.

The rationale in `packages/mcp/src/diagnostics.ts:6` says "detailed diagnostics belong to an explicitly injected host capability", but no such capability exists: `McpManagerHost.notifications.occur` carries only `{ code: 'mcp.connectionFailed', serverName }` and neither host interface in `packages/mcp/src/ports.ts` has a detail sink. Keeping raw text out of default console logging is right; please either add the bounded, explicitly injected detail sink that the comment describes, or correct the comment and the description so the boundary is stated as it is.

Related, same area: the stdio read buffer is 10 MB (`packages/mcp/src/client.ts:93`) while `MCP_TOOL_RESULT_MAX_BYTES` and `MCP_CONTROL_RESULT_MAX_BYTES` are 32 MB and are enforced only after deserialization (:102-103, :1496, :1842). The SDK's `ReadBuffer` throws once the buffer would exceed `maxBufferSize` (`@modelcontextprotocol/sdk/dist/esm/shared/stdio.js:12-14`), so a result between 10 MB and 32 MB tears down the stdio transport instead of producing the intended bounded error, and the 32 MB cap is unreachable on stdio. The explicit `maxBufferSize` also equals the SDK default, so it adds no bound. Either raise the buffer to match the caps or lower the caps to match the buffer and say so.

## Gates that are claimed but not enforced

- **The `@deepchat/shared` browser-safety invariant has no gate.** `docs/architecture/monorepo-migration/spec.md:92,173` states it as an invariant, but nothing checks it (`git grep -i browser -- scripts` finds only renderer-baseline and browser-overlay hits) and `packages/shared/package.json` declares no `browser` condition. Because the renderer aliases shared *source* (`packages/desktop/electron.vite.config.ts:95`), a `node:` builtin added to a renderer-reachable shared module yields a Vite externalization warning rather than a failure. The closure is clean today; nothing keeps it clean.
- **The forbidden-specifier scan is specifier-shaped.** `scripts/build-agent-kernel.mjs:17-22` matches only `@/`, `@shared/`, `electron`, `better-sqlite3`, `node-pty` and `src/main`, so a kernel edge onto an allowed workspace package that itself pulls Node or Electron passes. `scripts/package-artifact.mjs:293-297` scans the staged closure for electron/sqlite/pty but explicitly allows `node:` builtins, and the artifact/consumer gate drops the `src/main` pattern (`packages/agent-kernel/test/kernelPackage/agentKernelPackageRuntime.test.ts:51-55`), so "host paths" is enforced only by the build scan.
- **`--project kernel` selects no files.** The project includes `test/kernelPackage/**/*.test.ts` (`packages/agent-kernel/test/vitest.config.ts:12`) and its only match is excluded (`vitest.config.ts:22-28`), so the kernel segment of `test:main` (`package.json:13`) runs nothing — and the no-op string is pinned by `packages/desktop/test/main/scripts/testEntrypoints.test.ts:21`.
- **`scripts/build-shared.mjs` emits `packages/shared/dist` with no forbidden-specifier scan**, unlike `build-agent-kernel.mjs` and `build-mcp.mjs`; only the `artifact` vitest project catches a bad shared artifact.
- **`closeAll` is hardened but its neighbours are not.** The phase timeouts are a real improvement, yet `hydrations` and `operations` are still never cleared (`packages/desktop/src/main/agent/acp/instance/acpAgentRuntime.ts:250-273`) and `closeByAgent` (:242-248) still awaits hydration and `instance.close()` with no bound, so `AcpRuntimeOwner.refreshAgents` can hang exactly as `closeAll` no longer can. Both pre-existing, but now inconsistent enough to trip over.
- **`providers.json` drift is structural, not just a pending refresh.** The file has no `source_updated_at` and the sanitizer exists (`packages/desktop/scripts/fetch-provider-db.mjs:145`, run from `prebuild`), but the success path always rewrites the file and there is no `--check` counterpart to `icons:check`, so a normal networked build dirties the tree and nothing detects it.
- **Minor:** the root `preinstall: npx only-allow pnpm` guard was dropped and nothing re-adds it; `packages/desktop/scripts/build-cli.mjs:9` names the Desktop package dir `repositoryRoot`; `@deepchat/agent-kernel` has no `typecheck` script, so kernel types are checked only transitively through the Desktop program.

## Claims I verified as holding

- **ACP cancellation signal.** The replacement controller is gone (`adapters.ts:111` passes `abortSignal: input.abortSignal`), there is no `new AbortController`/`AbortSignal.any` in the ACP projection or runtime helpers, and every `abort` listener is removed in a `finally`. Abort, timeout and peer-exit all reach `send`. Wording caveat rather than a code caveat: the helpers ACP actually calls (`accumulate`, `startEcho`, `finalize`, `finalizeError`, `resolveProviderTerminalDecision`) never read `io.abortSignal`, so "all the way into projection helpers" overstates what the signal does today.
- **ACP ownership.** `AcpSessionStateAdapter` touches only `SessionSettingsStore` and `ProviderModelResolutionPort` — no built-in scope hydration — and the compatibility factory builds an unregistered instance fenced by `AcpSessionOwnershipPort`, publishing status itself instead of going through `runLifecycle.transitionCurrentStatus`. The destruction half is finding 1.
- **Settings sanitization.** `updateGenerationSettings` persists only keys present in the request with the current row as base (`acpSessionStateAdapter.ts:99-113`), matching `SessionSettingsCoordinator`, and the store's per-column update is `hasOwnProperty`-guarded. No field is dropped.
- **Kernel boundary.** `private: true`; all 162 distinct consumer specifiers resolve against the exports map; no kernel re-export shim survives in `packages/desktop/src/main/agent/deepchat/**`; zero `electron`/`better-sqlite3`/`node-pty`/`@deepchat/desktop`/`@/`/`require(` hits in `packages/agent-kernel/src`; the 57 `Pick<…>` usages are all over kernel-internal or shared port types rather than host classes.
- **The scan is real and wired.** It covers emitted `.js` *and* `.d.ts`, fails the build, handles `import type`, dynamic `import()` and `require()`, and runs from the package script through the root build into `prcheck.yml:129`. Remaining bypasses are narrow (computed specifiers, `/// <reference types="electron" />`).
- **Shared has no upward dependency** (`ollama`/`tokenx`/`zod`, with `ollama` type-only), and the compiler-symbol identity regression really runs: `sharedSourceAliases.test.ts:58-82` asserts host/shared `MessageFile` identity and zero `TypeFlags.Any` inside `test:main`.
- **The serialized artifact writers close the `dist` race.** The `artifact` project is the only reader of `shared/dist` (`fileParallelism: false`, `maxWorkers: 1`) and every other project aliases to source. `package-impact`, `package-size`, `release-preflight` and the native-helper check are all present in CI, and every `pnpm run …` command quoted in the migration docs is a CI step.
- **Provider fail-fast flip.** The default is `options?.swallowErrors === true` (`packages/provider/src/runtime.ts:456`), with a regression test at `packages/desktop/test/main/provider/providerRuntime.test.ts:479-509`, and the other callers opt in explicitly. One caller lost the old behavior and now embeds raw provider error text in model-visible tool output: `packages/desktop/src/main/tool/agentTools/agentToolManager.ts:2589,2611-2621`. Worth a fixed, non-echoing message there.
- **Provider/MCP diagnostics redaction.** Only two log lines were added across the provider scope and both are metadata-only; `dev`'s `logger.info('generateCompletion', …, messages)` became metadata plus `messageCount`, and no credential or token is logged in either package. Note the boundary is the diagnostics and status channels, not "the renderer": the OAuth auth path still forwards a raw (token-redacted) error via `McpServerAuthStatus.error` (`packages/desktop/src/main/mcp/mcpOAuthManager.ts:65-75`), and that is pre-existing.
- **MCP bounds.** 45 s soft / 5 min hard connect, 8 s/20 s probes, stderr drained, SIGTERM→SIGKILL with a 2 s grace and 500 ms settle, launch record retained until termination is confirmed, abort listeners removed in `finally`, and pending server→client requests are not orphaned because the SDK aborts in-flight handler signals on close.
- **MCP artifact gate is real and runs in CI**, not just in docs: `scripts/check-mcp-artifact.mjs` plus a strict NodeNext consumer exercising both stdio and HTTP, wired into `test:mcp:artifact` and therefore into `test:main`.
- **Desktop relocation.** `electron-builder.yml` is a pure rename; `appId`, `productName`, `win.executableName` and the package `name: DeepChat` (so `userData` is unchanged) all hold; every `files`/`extraResources`/`afterSign`/`buildResources` target resolves inside `packages/desktop`; `resources/`, `build/` and `plugins/` are otherwise byte-identical to `dev`; runtime lookups are `app.getAppPath()`/`process.resourcesPath`-relative and followed the move. Every root `build|dev|test*|format|lint|i18n*` script still exists and delegates, and all `node <path>` targets in both manifests exist at this head.
- **CLI boundary.** `packages/cli/src/launcher.mjs:9-45,47-92` checks host and module, prints a clear stderr message and exits 127 with no system-Node fallback; dev execution goes through `packages/desktop/scripts/run-cli.mjs` with a clear error and exit 1 when unbuilt; the bundle externalizes only `node:` builtins. Watch mode is gone, and `docs/guides/cli.md:256-264` documents the build-then-run loop.

## Not reviewed

- The 2,891 renamed files as such; the relocation was sampled for path and packaging consequences.
- I ran nothing — no build, no tests, no gates — so every conclusion above comes from reading the code at this head and comparing it with `origin/dev`.
- Cross-platform packaging, signing, DMG/ZIP publication and the update flow, which the description already scopes out.
- Residual `cwd`-coupled dev fallbacks that only resolve because pnpm sets `cwd` to `packages/desktop` (`backgroundExecSessionManager.ts:1605`, `watcherHostClient.ts:212`, `schedulerProcessManager.ts:239`, `runCodeRuntimeManager.ts:1171`, `plugin/index.ts:1372,1422`). Not defects today, but they are the same class as finding 3.

packages/desktop/src/main/app/composition.ts
```diff
@@ -2048,7 +2152,14 @@ export async function createMainProcessControl(dependencies: {
       destroySessionBrowser: async (sessionId) =>
         await yoBrowserPresenter.destroySessionBrowser(sessionId)
     },
-    state: deepChatAgentHarness,
+    // Backend-routed lifecycle destroy: ACP sessions never hydrate built-in scope, so there is
+    // no built-in owner to tear down; backend cleanup already ran through cleanupSessionBackends.
+    state: {
+      destroySession: async (sessionId) => {
+        if (isAcpBackendSession(sessionId)) return
```
zhangmo8, line 2159: `destroySession` now early-returns for ACP sessions, but no other stage in the deletion path removes the shared durable rows. `cleanupSessionBackends` → `harness.cleanupSession` is runtime-only by design ("Releases runtime state without deleting durable session facts", `sessionLifecycleCoordinator.ts:107-112`); `directAcpAgentBackend.cleanupSession` → `deleteDurableSession` only runs `DELETE FROM acp_sessions` (`acpSessions.ts:217`); `appSessionService.delete` removes the metadata row, search documents and the `new_sessions` row (`appSessionService.ts:219-223`).

On `dev` this call site passed the harness itself (`state: deepChatAgentHarness`), so `SessionLifecycleCoordinator.destroy` deleted the transcript, pending inputs, settings row and tape for ACP sessions too. There is no FK cascade on the session tables and both backends share `sessionData.transcript`, so a deleted ACP conversation now keeps every message, assistant block, user-message file/link, pending input, settings row, active-skill and disabled-tool row and tape entry permanently — invisible to the UI and to any later cleanup. Please route ACP sessions to an equivalent durable teardown instead of skipping the stage.

packages/desktop/src/main/agent/acp/instance/acpSessionStateAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+    return {
+      status: 'idle',
+      providerId: row.provider_id,
+      modelId: row.model_id,
+      permissionMode: row.permission_mode
+    }
+  }
+
+  async getSessionListState(sessionId: string): Promise<DeepChatSessionState | null> {
+    // List projections only need a settled placeholder; this is not a status truth source.
```
zhangmo8, line 63: `getSessionListState` returns a constant `idle` and the comment says it is not a status truth source, but `SessionAssignment.assessTransferSession` calls `handle.snapshot({ lightweight: true })` specifically for ACP sessions (`assignment.ts:508-510`) and treats the result as live status: `prepareSessionForTransfer` returns early unless it is `generating` (`assignment.ts:554`), and the stop-poll at `:563-566` can never observe `generating`.

On `dev` the ACP backend was constructed with `sessionState: deepChatAgentHarness` (`origin/dev:src/main/app/composition.ts:1940`), whose `getSessionListState` reads the live summary, so this path used to see a real status. As it stands, transferring a generating ACP session no longer cancels the in-flight prompt before reassignment. Either expose live ACP status through the adapter for this read, or stop using the lightweight snapshot in the transfer path.

PR #2319, opened 2026-09-17
feat(sync): add tunnel host sync endpoint

zhangmo8:
## Summary

Host-side sync endpoint for a user-operated Cloudflare Tunnel: one DeepChat instance becomes a host, exposes an authenticated sync endpoint that its own tunnel forwards to, and other devices pair with it and pull a backup snapshot.

Self-contained: the endpoint, pairing, device tokens, snapshot serving and their tests land here. The tunnel process itself (binary provisioning, transport selection, lifecycle) is out of scope and is intended to ship separately as its own plugin-backed helper, so nothing in this PR depends on it.

### What lands

- **Loopback-only endpoint** (`127.0.0.1`, ephemeral port) serving `handshake`, `pair`, `status`, `snapshot`. Binding is hard-coded to loopback and never `0.0.0.0`; a test asserts it is unreachable on the machine's LAN address *and* on `127.0.0.2`, so the check can never silently skip.
- **Pairing**: short-TTL single-use code carrying the host identity; `POST /pair` exchanges it for a per-device token. Failed attempts never invalidate or globally block the code — an anonymous caller who knows the hostname must not be able to deny pairing. Failures are charged to the calling source, and there is no global attempt counter or `attemptsRemaining` value for a caller to drive.
- **Device tokens**: hash-only at rest, constant-time comparison, immediate revocation (durable across restart). Phase 1 issues unscoped, non-expiring tokens — the store supports expiry but pairing does not set one, and there is no scope model, so a leaked token stays valid on every route until a human revokes it.
- **Snapshot pull**: streams the newest **backup package** (`backup-<epochMs>.zip`; any other zip in the sync folder is ignored) with `Content-Length` + SHA-256 headers, supports `Range`/206 with `Content-Range`, 416 on unsatisfiable ranges, 409 on an empty package, 404 when there is no snapshot. Resumed transfers are byte-exact, and a storage failure degrades to "no snapshot" rather than a 500.
- **Private machine-local state** (`<userData>/sync-host/host-state.json`, `0600`, atomic temp+rename) for the enabled flag, host identity and device records — deliberately **not** the settings store, because settings travel inside backup packages and are merged on import.
- **Audit ring buffer** (device, method, bytes, result, client IP; never tokens or payloads). Anonymous traffic is coalesced into one entry per source with a `suppressed` count, so unauthenticated requests cannot flush the ring.
- **Renderer-facing `syncHost.*` IPC surface** — seven routes, each asserting a renderer caller, since enabling host mode opens a network listener and pairing mints tokens.
- `docs/features/cloudflare-tunnel-sync/{spec.md,plan.md}` — the design notes (with transport validation evidence and a review record) and the implementation notes.

### Security posture

- Every unauthenticated request other than `handshake`/`pair` receives one uniform 401; authentication runs before path and method handling so the route surface cannot be mapped.
- Per-request *receive* timeout (60 s, independent of response streaming) with a per-socket guard, and a 32-connection ceiling that **evicts the oldest connection which is not streaming a response** instead of refusing the newcomer — a caller holding stalled sockets cannot deny the legitimate device or the user's own pairing.
- Rate limiting per device and per source, including `handshake`, applied before path and method handling; limiter state is pruned **and** hard-capped, so it cannot grow without bound.
- Archive errors and stream errors are contained: a corrupt package degrades to a null format version, a mid-stream failure aborts the connection instead of hanging the client, and an aborted request always settles and is audited (499).
- Host state never reaches the synced settings keyspace, and a failed state read or write fails closed instead of replacing or diverging from what is on disk.

### Known limitations (stated, not silently missing)

- **Provider credentials are inside `agent.db`**, so the served package currently carries them. Serving a package is what the existing S3/R2 flow already does; the difference is that this path hands it to every paired device, so "credentials never leave the machine" would be false until the export redacts them or the feature ships explicit consent. **Needs a product decision** before any device-visible UI.
- The host serves the newest existing package; it does not yet produce one on demand, and `startBackup` refuses while the legacy S3 sync toggle is off.
- Tokens are unscoped and non-expiring in practice, as stated above.

## Verification

- `pnpm run typecheck` (node + web), `pnpm run lint`, `pnpm run format`, `pnpm run i18n` — all clean.
- `test/main/sync/host/hostEndpoint.test.ts` — **30 tests against a real listener**: uniform pre-auth 401s, authenticated 404/405/501, pairing single-use and per-source failure accounting, revocation across a state reload, token-hash containment on disk, byte-exact `Range` resume, abort-then-resume, abort *during* snapshot resolution (settles and audits 499), corrupt archive, unreadable backup list, loopback-only reachability, stalled-request reaping, connection-ceiling eviction, keep-alive sockets staying evictable after a download, per-device limiting of unknown paths, anonymous-audit coalescing, package-name filtering, host-identity stability and lazy persistence, pairing-code restoration, oversized-body 413, pre-`initialize()` state preservation, interleaved enable/disable consistency, teardown.
- `test/main/sync/host/state.test.ts` — **3 unit tests**: concurrent failing updates roll back, a failed mutation stays out of the next successful write, and a read/parse failure fails closed instead of resetting to defaults (then recovers).
- `test/main/sync/host/routes.test.ts` — **12 route-level tests** for the seven `syncHost.*` handlers: coverage, status/pairing shapes, `setEnabled` pass-through and failure propagation, pairing creation and fallback, device-list and audit redaction, revoke/rename pass-through, input validation, renderer-caller enforcement.
- Broader suites: `test/main/{sync,contracts,logging,routes}` + `test/main/cli/surface.test.ts` — **403 tests pass**.

No UI changes in this PR, so no before/after layout.

## Reviewer notes

Three rounds of review (security, correctness/lifecycle, spec-conformance, plus a maintainer pass) are recorded in the `Review record` section of `docs/features/cloudflare-tunnel-sync/plan.md`; every actionable item is fixed here. The high-severity ones: device token hashes would have travelled inside backup packages, a mutation arriving before the initial state read could wipe the state file, an aborted snapshot request hung forever and was never audited, and a failed state read silently replaced the real file with defaults. Each fix carries a test that was verified to fail against the unfixed code.

## Summary by CodeRabbit

* **New Features**
  * Added host-side sync support through a secure, loopback-only endpoint.
  * Added pairing codes and device management, including listing, renaming, revoking, expiration, and authentication.
  * Added sync status, enablement controls, audit history, and snapshot availability reporting.
  * Added resumable snapshot downloads with range support and validation.
  * Host sync state is stored locally and excluded from backup data.

* **Bug Fixes**
  * Improved startup, shutdown, rate limiting, pairing recovery, and connection handling.


zerob13 (approved):
## Review: tunnel host sync endpoint

**Verdict: approved.** This is a well-built phase-1 data plane. The security posture is consistent (auth before routing so the route surface can't be mapped, hash-only tokens with constant-time comparison, loopback-only binding asserted by a test against the LAN address), lifecycle handling is careful (atomic 0600 state writes with rollback, listener/flag transitions serialized, abort settling audited as 499), and every abuse-surface bound (rate windows, pairing failure budgets, connection ceiling with non-streaming eviction, audit coalescing) is both pruned and hard-capped. Keeping host state out of the settings store is the right call — token hashes travelling inside backup packages would have been the worst bug in this feature, and the PR fixes and regression-tests exactly that class of issue.

Verified locally on this branch: `test/main/sync/host` 38/38 pass, full `test/main/sync` 77/77 pass, `typecheck`, `oxfmt --check`, `oxlint` all clean.

Nothing below blocks merge; treat them as follow-ups or nits for the next stage.

### Non-blocking findings

- **P3 — authenticated 404/405 bypass the per-device rate limit and can flush the audit ring** (`src/main/sync/host/endpoint.ts`, `handle()`). The `HANDLED_PATHS` 404 branch and the global 405 branch both run *before* `consumeRateLimit(device.deviceId)`, so a paired device spamming unknown paths never hits the limiter — and every response writes an audit entry. With `SYNC_HOST_AUDIT_LIMIT = 500`, that evicts legitimate audit history, which is exactly the eviction problem the anonymous coalescing exists to prevent. Moving the rate-limit consumption above the 404/405 branches would close it. Impact is low: it requires a device the user explicitly paired, and such a device has far more direct powers (it can download the whole snapshot), so this is a hardening nit, not a hole.
- **Nit — pairing code modulo bias** (`pairing.ts` `createCode()`): `bytes[index] % 31` over 256 values biases the first 8 alphabet entries ~1.4%. Irrelevant against the 20-failures/5-min per-source budget and ~40 bits of entropy, but rejection sampling is free if you ever revisit this file.
- **Nit — pairing code burned before device issuance** (`endpoint.ts` `handlePair()`): `pairing.consume()` succeeds, then `devices.issue()` runs; a state-write failure at that point spends the code and the user must generate a new one. Fail-safe direction, just slightly annoying for the user.
- **Nit — dead constant**: `SYNC_HOST_MAX_PUSH_PART_BYTES` is declared but unused until push ships. Acceptable as a spec forward-declaration; noting for completeness.
- **Nit — `state.update()` rollback under interleaving** (`state.ts`): the `if (this.state === next)` rollback only fires when no other update interleaved; in a contrived interleaving a caller can be told its write failed while the mutation later lands via a subsequent successful write. Direction is safe for revocation (safer to persist), not worth restructuring.

### Scope and conformance notes

- Purely additive on the product surface: new main-process service + endpoint, new `syncHost.*` route contracts, composition wiring, docs, tests. No existing behavior changes, no breaking contract changes — `routes.ts` only adds a new catalog part.
- Routes follow the existing registry pattern (`createRouteMap`, `defineRouteContract`, zod in/out), and `requireRendererCaller` on all seven handlers is correct for a surface that opens a listener and mints tokens.
- Test volume (38 tests, many against a real listener) is proportionate to the behavior surface: each test pins a distinct guarantee (loopback-only, uniform 401, single-use pairing, byte-exact resume, abort settling, eviction policy, state preservation). Not over-tested; nothing I'd cut.
- The known gaps (provider credentials inside the served package, unscoped non-expiring tokens, no on-demand package production) are honestly documented in the spec and PR description, with the credentials question explicitly flagged for a product decision before the slave side ships. The right place to hold that line is the Settings UI / push stages.

### Detailed analysis

<details>
<summary>What I checked and how</summary>

- `endpoint.ts`: request flow (handshake → pair → auth → 404/405 → rate limit → route), `Range` parsing incl. suffix ranges and multi-range rejection, snapshot streaming with abort-before-first-await guard, connection eviction policy (`oldestNonStreamingSocket`), per-socket receive guards vs Node's 30 s connection-checking interval, bounded window maps (prune + hard cap), anonymous audit coalescing, `readBody` overflow drain for a real 413. The P3 above is the only asymmetry I found in the ordering.
- `pairing.ts`: single-use codes, in-memory only (restart invalidates — intended direction), failed attempts never penalize the code, ~40-bit entropy. Modulo bias noted above.
- `devices.ts` / `state.ts`: token hashing (SHA-256, constant-time), revocation/expiry semantics, atomic temp+rename with 0600 and fsync, fail-closed on read errors (a corrupted read never degrades to defaults), pre-`initialize()` mutation safety, write-chain rollback semantics.
- `snapshot.ts`: `backup-<epochMs>.zip` name filtering (matches the other call sites in the sync pipeline), streaming manifest extraction with a 1 MiB bomb ceiling, digest cache keyed on size+mtime+ino+ctime with re-validation and a retry loop for mid-read replacement, single-flight for concurrent digest passes.
- `index.ts` / `composition.ts`: lifecycle serialization, start-before-persist with rollback, descriptor file publication, `startIfEnabled` at boot with startup failure reporting, teardown stop step. Wiring uses existing `syncService.listBackups()` / `syncSettings.getFolderPath()`.
- Contracts (`syncHost.ts`, `syncHost.routes.ts`): route catalog addition is append-only; schemas match what the handlers produce.
- Tests: read both suites end to end; ran them plus the full `test/main/sync` tree, typecheck, format and lint on this branch.
</details>

zhangmo8:
Addressed in `010f705d3` + `95b3e691e` + `89d208d1e` — all six CodeRabbit findings and @zerob13's P3/nits.

Note on the anchors below: they refer to the revision CodeRabbit reviewed. The doc line numbers have since shifted (the tracker-issue linkage was dropped from `spec.md`/`plan.md`), so the doc items are described by section instead of line.

**CodeRabbit**

- **`plan.md`, slice 2 completion note** — `handshake` **and** `pair` are now named as the unauthenticated routes, with why `pair` cannot require a token (it is what issues one). ✅
- **`spec.md`, Resolved Questions → "Default sync scope"** — the row now states that provider credentials are **currently included** (plaintext columns in the `agent.db` every package carries) and marks the exclusion unresolved, matching the Excluded Data section instead of contradicting it. ✅
- **`endpoint.ts`, snapshot streaming** — the streaming flag is cleared when the stream settles, aborts included. Without it a finished keep-alive download stayed unevictable, and 32 of them turned the ceiling into a wall; the test fills the ceiling with completed ranged downloads and still requires a newcomer to be admitted. ✅
- **`index.ts`, `getHostId()`** — no longer writes fire-and-forget: the identity stays in memory until `initialize()` persists it through its awaited update, so a failed write surfaces to its caller instead of leaving the cache and the file disagreeing. ✅
- **`state.ts`, `update()`** — the whole transaction (load → mutate → write → rollback) is serialized and the rollback is unconditional. A state-store test runs two concurrent failing updates and asserts the cache is unchanged. ✅
- **`hostEndpoint.test.ts`, pre-`initialize()` test** — `seed.stop()` before cleanup, so the listener it started is released. ✅

**zerob13**

- **P3** (authenticated 404/405 bypassing the per-device limiter, and every 404 evicting a legitimate audit entry) — the limiter now runs before path and method handling; a test probes unknown paths until it is throttled. ✅
- Modulo bias — rejection sampling (`248 = 31 × 8`). ✅
- Pairing code burned before device issuance — the code is restored when `devices.issue()` fails, so a storage failure no longer costs the user a code. ✅
- Dead constant `SYNC_HOST_MAX_PUSH_PART_BYTES` — kept as the forward declaration for the push stage. ⏭️
- Rollback under interleaving — the serialized transaction above removes the interleaving entirely. ✅

**Verification** (at `89d208d1e`): `test/main/sync/host` 45/45 (30 endpoint, 3 state, 12 route); `test/main/{sync,contracts,logging,routes}` + `test/main/cli/surface.test.ts` 403/403; `typecheck` (node + web), `lint`, `format`, `i18n` clean.

---

Correction to this comment as first posted: it claimed every fix carried a regression test that fails against the unfixed code, and that was not yet true for two of them. The host-identity test passed with the fire-and-forget write restored, and `restore()` had no test at all. Both are pinned now in `95b3e691e` — reading the identity must not write the state file (fails against the old code), and `restore()` returns a spent code without clobbering a newer or expired one. Every fix above except the modulo-bias change now has a test verified to fail against the unfixed code; the bias change is statistical and is not test-asserted.

zerob13 (approved):
## Re-review: third-round fixes (3dc7b5a → 010f705)

**Verdict: approved.** Everything from the previous round — my P3/nits and the six CodeRabbit findings — is fixed, each fix carries a regression test, and the fixes match what the tests claim to pin.

What changed and why it holds up:

1. **Rate limiting now runs before path/method handling** (`endpoint.ts`). A paired device probing unknown paths is finally charged for it, and its 404 spam can no longer flush the audit ring. Covered by a test that probes unknown paths until throttled.
2. **Finished keep-alive downloads no longer leave their socket permanently unevictable** (`endpoint.ts`). The streaming flag is cleared in a `finally` when the stream settles, aborts included. Covered by a test that fills the entire connection ceiling with finished one-byte downloads and still admits a newcomer.
3. **A failed device issuance no longer burns the user's pairing code** (`endpoint.ts` + `pairing.ts`). The outstanding code is captured before consumption and restored on failure; `restore()` correctly refuses to clobber a newer code or revive an expired one.
4. **Pairing code generation uses rejection sampling** (`pairing.ts`). The alphabet is 31 characters, so bytes ≥ 248 are discarded and the modulo bias is gone. One trivia nit, non-blocking: the comment says the old bias was ~1.4%, but a favored character appeared 9/256 times against a fair 8.26/256 — roughly +9% relative. Moot now that it's fixed.
5. **State updates serialize the whole transaction, not just the writes** (`state.ts`). This closes the interleaving where a second update could snapshot the first one's mutation before its write failed — which would have both blocked the rollback and persisted a change whose caller was told it failed. The rollback is now unconditional, which is safe precisely because updates are serialized. Covered by concurrent-failure and failed-mutation-isolation tests.
6. **`getHostId()` no longer persists the pre-initialize identity fire-and-forget** (`index.ts`). `initialize()` owns the awaited write, so a failed write can't leave the in-memory cache and the file disagreeing about who the host is.

Verification on this head: `test/main/sync` 82/82 pass (77 before this round), typecheck node+web clean, oxlint 0/0, oxfmt clean.

Test additions are proportionate — five tests, each pinning exactly one fixed defect or durability contract, no filler. Docs (the `plan.md` findings table and the `spec.md` scope row) now match the implementation.

This is ready to merge from my side.

zerob13 (approved):
## Re-review: delta since `010f705d` (last approved head)

Scope: `95b3e69` (test) + `89d208d` (docs). Verdict: **both are fine, no findings.**

**What changed and why it's OK**

1. **`95b3e69` — two regression tests, not new behavior.** These pin the two review fixes from the previous round that had no coverage:
   - *Host identity*: `getHostId()` must not touch the state file before `initialize()` runs. The test proves it by asserting `host-state.json` is absent after a read plus a settle window, and that `initialize()` then persists the exact identity that was already handed out. This fails against the old fire-and-forget write — a genuine regression guard for a persistence/lifecycle invariant.
   - *Pairing `restore()`*: verifies a consumed code is returned, a newer user-created code is never clobbered, and an expired code is never revived. The implementation guard (`this.code || expiresAt <= now`) matches each case.
   Both fall under the durable-test categories the project keeps (persistence, lifecycle, recovery). No over-testing.

2. **`89d208d` — docs only.** Drops the `#2302` issue linkage from plan/spec. The stated rationale (host endpoint is a self-contained data plane; the tunnel helper ships separately) is coherent and leaves no stale references behind.

**Verification**

- `vitest run test/main/sync/host` — 45/45 passing across all 3 files, including the 2 new cases.
- Implementation cross-checked: [src/main/sync/host/index.ts](../blob/feat/sync-host-endpoint/src/main/sync/host/index.ts) (`getHostId`/`initialize`), [src/main/sync/host/pairing.ts](../blob/feat/sync-host-endpoint/src/main/sync/host/pairing.ts) (`restore`).

The delta closes out the review findings cleanly; the earlier rounds' approval stands.

PR #2320, opened 2026-09-17
feat(chat): scroll-to-latest pill and message map

zhangmo8:
## Summary

Implements #2313: getting around a long conversation without losing your place.

- **Scroll-to-latest pill** — appears whenever the viewport leaves the bottom, using the scroll controller's own 80 px `nearBottom` proximity flag, so the indicator and auto-follow can never disagree about where "the bottom" is. Activation issues `reason: 'user-return-to-bottom'` with a `{ kind: 'bottom' }` target and the **controller performs the write**. The badge counts the loaded messages that reach below the fold.
- **Message map** — a rail in the right gutter of the message viewport with one hairline mark per message, on a constant 16 px pitch so the spacing never depends on how tall a single message is. A mark's width encodes how long that message is relative to the longest one in the conversation. Hovering a mark highlights it and previews the message in a popover; clicking one jumps there, and the arrow keys walk the marks with Home/End at either end. Once the marks no longer fit, the map scrolls on its own instead of squeezing them together.
- Both derive from the same logical layout table the windowing uses, so the marks and the badge cannot describe different geometries.

## Design notes

- **New scroll reason `indicator-navigation`** rather than reusing `spotlight-navigation`: the scroll-ownership contract requires every programmatic scroll to carry an explicit reason, and reusing the Spotlight reason would misreport the cause in the audit trail. It is wired through the priority table, the acceptance gate and the `navigating` transition.
- **Shared jump path**: the retry-plus-highlight routine was extracted from the Spotlight deep link into `jumpToMessage`, so the pill and the map reuse it instead of adding a second one.
- **One derivation**: `useMessageVirtualization` exposes the below-viewport *messages* rather than a bare count, so the badge cannot disagree with the list. Deliberately not `messageWindowRange.end`, which includes 2400 px of overscan.
- **Count boundary** — a message counts when its *bottom* crosses the fold, not when it starts below it. A long answer, or a reply still streaming, fills the bottom of the viewport while its end is off screen; the earlier "entirely below" rule reported nothing there, which left the pill showing no badge in exactly the case where the user had just triggered a message. A useful invariant falls out: while the pill is visible the count is at least 1.
- **Mark width is compressed** — width is `sqrt(length / longest)`, not the raw ratio. A real session measured 55 to 7837 characters between its messages, i.e. 140:1, and raw proportions put every question on the 2 px floor while only the longest answer was visible. The compression keeps the ordering and keeps the longest message at full width.
- **Map layout** — marks are laid out by index on a constant pitch, not by their position in the conversation, so one enormous message cannot stretch the gaps around it. The stack is centred while it fits the rail and scrolls once it does not. The rail sits 20 px in from the viewport's right edge so its hit area clears a classic ~17 px Windows scrollbar, not just the app's 8 px one.
- **Hover preview** — a `DcPopover` anchored to an invisible marker that follows the hovered mark, so the card points at the message under the pointer. The text is projected from the display content on hover only, which keeps the preview at one message regardless of history length, and it is clamped to four lines because it is a thumbnail rather than a second copy of the message.
- **Rail interaction** — the rail is a single `role="slider"` reporting the message at the top of the viewport; positions resolve to a slot, so the space between two marks belongs to one of them rather than being a dead zone. It never scrolls the conversation by itself: it emits a message id and ChatPage routes the request through the controller.
- **Reading position** — the mark at the top of the viewport is drawn brighter than the rest (resting < current < hover), and the rail scrolls itself to keep that mark on screen. It only moves when the mark is actually outside the visible slice, so browsing the rail by hand is not undone the next time the conversation scrolls. The rail is present from the conversation's first message and never toggles, which keeps the message text from reflowing under it.
- **Jumps are settled before the user's own navigation** — `returnToLatest` cancels pending jump retries and moves the jump generation on, because a retry would otherwise re-issue an explicit navigation (accepted unconditionally by the controller) and pull the viewport away from the bottom the user just asked for.
- **Retries report their outcome** — each scheduled retry carries the resolver of the promise its caller awaits and resolves it with the chain's result; every cancellation path resolves `false`. Without that, a jump that succeeded on a later attempt left the Spotlight path's pending state behind forever.
- **Role is shown by width and position, not by colour or label** — the repository has no localized "User"/"Assistant" copy (the message list uses data-provided names), and inventing 23×2 translations for this was not worth it. The rail's own accessible name is a single new key, `chat.messages.minimap`.
- Scroll-ownership spec updated with the new reason, and two drifted spots repaired (missing `history-navigation` in the reason list; missing `resumeUserOwnedAfterNavigation` / `hasExplicitNavigation` in the state shape).

## Verification

- `pnpm run format`, `pnpm run lint`, `pnpm run i18n`, `pnpm run typecheck` (node + web) — all clean.
- **2505 tests passed** across the full renderer suite (278 files) on the commit before the map restyle; since then ChatPage's 98 tests and the map, pill and virtualization suites pass.
- New coverage: pill activation, count and accessible name; tick derivation (fractions, widths against the longest message, missing geometry), viewport window (clipping, unknown geometry), reading-position lookup, rail follow offset (visible, above, below, unmeasured); component: one mark per message, constant pitch, scrolling content height, hidden while the list fits, click-to-jump, slot resolution for hover, three-way mark styling, arrow/Home/End jumps, slider aria values, four-line preview.
- The scroll-architecture guard still passes, i.e. no new direct scroll writes outside the controller.
- The rail adds no timers or observers beyond one click, one keydown and one pointer handler, so it has no late-firing or cleanup obligations of its own.

## Known gaps

- **The hover preview has no real-window verification.** Electron windows reject synthesised background hover events, so the popover, its anchoring and the highlight are covered by component tests only.
- **One heavy test times out intermittently** in the full suite under load — a different unrelated file each run (`ChatPage > exposes trace diagnostics…`, then `BrowserDataImportDialog > prevents closing or leaving…`). The first reproduces with this feature's source files reverted to `646d44b02` and the second passes in isolation in 3 s, so these are pre-existing load flakes in that 10 s default, not regressions here.
- No real-Chromium coverage: the repo's sanctioned place for layout-dependent scroll behaviour is the opt-in e2e spec (`test/e2e/specs/31-chat-scroll-ownership.smoke.spec.ts`, requires `RUN_PROVIDER_INTEGRATION=true` and a real model). A unit-level ChatPage case was attempted and reverted rather than committed flaky — the harness's mocked rAF plus gesture and idle-debounce plumbing produced nondeterministic hangs for this path.
- The rail has no drag/scrub and no hover tooltip beyond the preview card: click-to-jump plus keyboard, per the agreed scope.
- The dead zero-inbound `MessageActionButtons.vue` arrow-down button and its `common.scrollToBottom` key are untouched; retiring them is a separate decision.

Translation note: `bo-CN`, `ug-CN` and `mn-Mong-CN` copy is my best effort and would benefit from a native review.


zerob13 (changes requested):
# Review — scroll-to-latest pill & message map

**Verdict:** High-quality PR. No blocking/P1 issues found. Requesting changes for the three items below — two are small fixes, one is a sign-off question.

## What this adds (for readers who haven't followed #2313)

- A floating pill above the composer that appears whenever you scroll away from the bottom. Clicking it returns to the latest messages; a small badge reports how many loaded messages currently sit below the fold.
- A thin vertical "message map" rail on the right edge of the message viewport: one colored mark per message (user vs assistant) plus a window box showing the slice currently on screen. Click a spot on the rail to jump to that message; arrow keys walk the marks one by one, Home/End jump to either end.

## Why the design holds up

1. **One source of geometry.** Both features derive from the same layout table the virtualization uses (`useMessageWindow` entries), so the pill count, the map marks, and the windowed list cannot disagree about where anything is. Index alignment between entries and `displayMessages` holds because entries are built by a 1:1 order-preserving `messages.map` — verified.
2. **Scroll ownership is respected.** Every jump goes through the chat scroll controller with the new explicit reason `indicator-navigation`; the rail itself never writes to the viewport. The spec is updated to match, and the diff also repairs two pre-existing doc drifts (`history-navigation` missing from the reason list, two fields missing from the state shape) — good housekeeping.
3. **The supersede guard fixes a real race.** Before this, a retry timer belonging to an older jump could re-issue an explicit navigation (accepted unconditionally by the controller) and drag the viewport back to a message the user had already scrolled past. The generation counter plus gesture-sequence guard in `messageJumpRetry.ts` closes this for both entry points (Spotlight deep link and the indicator), and the Spotlight path keeps its existing behavior.
4. **Tests are proportionate.** Pure-logic predicates (retry/supersede/tick math) plus light component smoke tests. No implementation-coupled noise. CI is green including `test-renderer`, and the two new i18n keys are present in all 23 locales.

## Items to address

### 1. Rail can overlap the scrollbar on Windows (P3, needs fix or confirmation)

`ChatMinimap.vue` places the rail at `right-3` (12px from the edge) with a 10px width, so it occupies the 12–22px band from the viewport's right edge. The message container uses `scrollbar-gutter: stable both-edges` (ChatPage.vue style block), and on Windows a classic scrollbar is ~17px wide — so the rail overlaps the top ~5px of the scrollbar track. The rail is `pointer-events-auto` with `z-index: var(--dc-z-sticky)`, so in that strip it will steal clicks meant for dragging the scrollbar. The "clear of the 8px scrollbar" note in the PR description reflects the macOS overlay scrollbar, not Windows. Please verify on Windows, or shift the rail past the gutter (or make it gutter-aware).

### 2. Pill's leave transition is skipped in one path (P3)

The float-layer parent in ChatPage.vue (~line 201) now has `v-if="latestPlanSnapshot || activePendingInteraction || showScrollToLatest"`. When the plan snapshot / pending interaction conditions drop while the pill is still visible, the parent unmounts synchronously and the pill's `<Transition name="scroll-to-latest">` never gets to play its leave animation — the pill pops out. Either keep the parent mounted while the pill is visible, or move the pill out of the shared `v-if`.

### 3. Badge semantics — conscious sign-off please (question)

The badge counts **loaded messages below the fold**, not "new/unread messages". Scroll a 500-message history to the top and it will read ~497. The PR notes explain this as a deliberate navigation-oriented choice fitting #2313, and the invariant "while the pill is visible, count ≥ 1" is nice — but users coming from Telegram/Slack style UIs will likely read it as an unread counter. Also note the count only covers *loaded* messages, so it under-reports while older history is unloaded. If that trade-off is intended, say so and this item is closed.

## Minor observations (no action needed)

- `role="slider"` on the rail is a slight stretch semantically (it reports a message position, not a numeric value), but the keyboard contract (arrows/Home/End + preventDefault) is coherent and focus-visible styling is present.
- `minimapTicks` recomputes on every measurement flush (entries rebuild), which adds per-frame allocations in long conversations — acceptable since entries rebuild is a pre-existing cost of the windowing pipeline.

## References

- `src/renderer/src/components/chat/ChatMinimap.vue` — rail placement, click/keyboard handling
- `src/renderer/src/components/chat/ScrollToLatestPill.vue` — pill + badge + transition
- `src/renderer/src/features/chat-page/model/minimapTicks.ts` — tick/window math, nearest-mark resolution
- `src/renderer/src/features/chat-page/model/messageJumpRetry.ts` — retry/gesture/supersede predicates
- `src/renderer/src/features/chat-page/ChatPage.vue` — float-layer `v-if` (~L201), jump wiring (~L890–995)
- `src/renderer/src/features/chat-page/composables/useMessageVirtualization.ts` — below-viewport derivation
- `docs/architecture/chat-scroll-ownership/spec.md` — new reason + repaired drifts

zerob13 (changes requested):
# Re-review — delta `58549a2d..1630ff5c` (hairline restyle + hover preview)

**Verdict: the response commit is good quality and fixes item 1. Requesting changes again because item 2 is untouched and item 3 still has no reply.** No new blocking issues.

## What this commit does (for readers who haven't followed the thread)

- Moves the message map from the right edge of the chat to the left gutter, and redraws each message as a horizontal hairline whose width is proportional to the message's text length — the longest message in the conversation fills the rail's full width.
- Adds a hover preview: hovering a mark highlights it and pops a card next to the rail showing that message's text.
- The "window box" that used to show which slice of the conversation is on screen is gone; the current position is now only reported to screen readers.

## Status of the previous review items

1. **Windows scrollbar overlap — fixed.** The rail now sits on the left gutter (`ChatMinimap.vue:121`, `left-3`), away from the scrollbar entirely. Thanks.
2. **Pill leave transition skipped — still open, unchanged.** `ChatPage.vue:203` still has the shared `v-if="latestPlanSnapshot || activePendingInteraction || showScrollToLatest"`. When the plan/pending conditions drop while the pill is visible, the parent unmounts synchronously and `ScrollToLatestPill`'s `<Transition name="scroll-to-latest">` (`ScrollToLatestPill.vue:32`) never plays its leave animation — the pill just pops out. This commit didn't touch that path.
3. **Badge counts loaded messages below the fold — still no reply.** If reading it as "new messages" (Telegram-style) and under-reporting while older history is unloaded is an accepted trade-off, please say so in this PR and the item is closed.

## New code — what I verified

- Hover and click share the same resolution function (`resolveMinimapTickIndex`), so the preview a pointer shows is always the message a click would jump to. Good design.
- Message length is cached per content object (`WeakMap` in `minimapTicks.ts:29`). A streaming reply replaces its content object as it grows, which is exactly when the length must be recomputed; settled messages are measured once. Correct and cheap.
- The preview text reuses the chat-search projector (`extractDisplayContentText` in `chatSearch.ts:507`), so the map, the widths, and search all see the same text — no duplicated text logic.
- I specifically checked whether the hover card follows its anchor: the invisible trigger span moves via inline `top` while the card is open. Verified in the shipped `reka-ui`/`@floating-ui` that `autoUpdate` runs with `layoutShift` enabled (IntersectionObserver-based `observeMove`), so anchor movement does reposition the card. Not a bug.
- `Math.max(...lengths)` follows an existing pattern over the same message list (`useDisplayMessages.ts:314`), so no new spread risk.
- Tests are proportionate: hover emit dedupe, preview portal, width scaling, empty-text fallback, click jumps. I ran the two touched files locally: **22/22 pass**. (CI `test-*` jobs were still pending at review time; `static` is green.)

## New notes (small, not blocking)

- `resolveMinimapTickIndex` (`minimapTicks.ts:104`) is a linear scan run on every `pointermove`. Ticks are sorted by `top` and the same file already contains a binary search (`findMinimapTickIndexAt`). A binary search plus neighbor comparison would do; fine at realistic conversation sizes, just wasteful on very long ones.
- With the window box gone, the map shows no visual cue for where you are — marks only highlight on hover. Keyboard users walking marks with the arrow keys get neither a position cue nor a preview. Consider dim-highlighting the `activeIndex` mark so the keyboard contract isn't blind.

## Also outstanding from CodeRabbit's pass at `58549a2d` (not addressed by this commit)

- Tick gap resolution still measures distance to tick centers rather than to interval edges (`minimapTicks.ts:119`).
- `returnToLatest` still doesn't bump `messageJumpSeq` / clear pending jump timers (`ChatPage.vue:641`).
- `jumpToMessage`'s retry branch still returns `false` immediately instead of a cancellable deferred result (`ChatPage.vue:902`).

## References

- `src/renderer/src/components/chat/ChatMinimap.vue` — hairline marks, hover handling, preview popover
- `src/renderer/src/features/chat-page/model/minimapTicks.ts` — width fractions, length cache, tick resolution
- `src/renderer/src/features/chat-page/ChatPage.vue` — hover wiring (~L1135–1147), float-layer `v-if` (L203)
- `src/renderer/src/lib/chatSearch.ts` — `extractDisplayContentText` (L507)
- `src/renderer/src/components/chat/ScrollToLatestPill.vue` — transition (L32)
- `test/renderer/components/ChatMinimap.test.ts`, `test/renderer/features/chat-page/model/minimapTicks.test.ts` — 22/22 pass locally

Reviewed at `1630ff5c`.

zerob13 (changes requested):
# Re-review — follow-up at `873d2bd0d` (two commits past my previous review at `1630ff5c`)

**Verdict: requesting changes again.** One real blocker: **`test-renderer` CI is red on this head** — two `ChatPage.test.ts` tests fail, reproduced locally. Also one correction to my own previous review, and items 2 and 3 from before are still open.

## First, a correction to my own previous review

I asked you to move the rail off the right edge because I believed it would overlap a classic ~17px Windows scrollbar. **That was wrong.** This app styles all scrollbars globally to 8px on every platform (`*::-webkit-scrollbar { width: 8px }`, `src/renderer/src/assets/style.css:1135`), so a rail at `right-3` (12px) clears the scrollbar by 4px everywhere. Moving the map back to the right gutter in `829a0d383` is fine — the left-gutter detour in `1630ff5c` wasn't needed for that reason. Sorry for the churn; the "clear of the 8px scrollbar" note in the PR description was accurate all along.

## The blocker: two ChatPage tests fail on this head

CI `test-renderer` fails at `873d2bd0d`, and I reproduced it locally (`pnpm vitest run test/renderer/components/ChatPage.test.ts -t 'geometry'` → 2 failed):

- `ChatPage > follows a restored session after the observed message geometry grows`
- `ChatPage > does not follow observed geometry when auto-scroll is disabled`

Both throw `TypeError: Cannot read properties of undefined (reading 'contentBoxSize')` inside `@vueuse/core`. The chain:

1. `1630ff5c..873d2bd0d` added `useElementSize(railRef)` to `ChatMinimap.vue` (for the even-pitch layout). `useElementSize` registers its own ResizeObserver callback.
2. The ChatPage test harness stubs ResizeObserver and keeps every registered callback in a `resizeCallbacks` array (`ChatPage.test.ts:3258`).
3. When the test fires them — `resizeCallbacks.forEach((callback) => callback([], {} as ResizeObserver))` at `ChatPage.test.ts:3315` and `:3388` — it passes an **empty entries array**. The VueUse callback reads `entries[0].contentBoxSize` and crashes.

The component itself is fine in a real browser; the test harness calls the ResizeObserver contract incorrectly (entries must be an array of `ResizeObserverEntry`, never empty when an observation fired). The smallest fix is in the harness: pass a realistic entry, e.g. build one `{ target: <the observed element>, contentBoxSize: ..., contentRect: ... } as ResizeObserverEntry` instead of `[]`. Please fix it there (or in whatever way you prefer) so the head is green — a red `test-renderer` blocks merge regardless of everything below.

## Status of the previous review items

2. **Pill leave transition skipped — still open, untouched.** `ChatPage.vue:203` still unmounts the shared float-layer parent synchronously when the plan/pending conditions drop while the pill is visible, so the pill's leave animation (`ScrollToLatestPill.vue:32`) never plays.
3. **Badge counts loaded messages below the fold — still no reply.** A one-line "intended" closes it.

## The two new commits, otherwise: good

For readers who haven't followed: past my previous review, `829a0d383` + `873d2bd0d` changed the map again — marks are now evenly spaced down a scrollable right gutter instead of positioned proportionally to conversation geometry, the rail itself scrolls once marks no longer fit at a 14px minimum pitch, and the mark widths are square-root compressed (a real session measured a 140:1 length ratio; raw proportions would have pinned every question to the 2px floor).

What I checked and liked:

- Hover/click resolution is now O(1) arithmetic on the fixed pitch (`indexAt` in `ChatMinimap.vue`), which replaces the linear scan I flagged last round and also makes CodeRabbit's tick-distance comment moot — `resolveMinimapTickIndex` is deleted entirely.
- The pitch limits are a clean pure function (`resolveMarkPitch`, min 14px / max 44px) with the max cap preventing a two-message conversation from pinning its marks to opposite ends. Sensible.
- The scrollable rail keeps pointer targets comfortable regardless of conversation length, and `dc-overscroll-contain` stops wheel events over the rail from chaining into the conversation scroll.
- The sqrt compression is documented with real measured numbers (`minimapTicks.ts:54-58`) — this is how magic constants should be justified.
- Tests track the redesign (pitch spacing, overflow-scroll behavior, gap ownership, clamped preview) — 27/27 pass locally in the minimap files.
- The popover anchor now sits inside the scrollable content box, so the preview card follows the rail's scroll — verified against the shipped floating-ui `autoUpdate` (scroll listeners + IntersectionObserver-based `observeMove`).

## Minor, not blocking

- With the rail scrollable, arrow-key walking (and jumping) doesn't reveal the target mark — nothing scrolls the rail to the mark the keyboard landed on. A `scrollIntoView`-style adjustment on key navigation would close the gap.
- The map still shows no visual cue for the current position (marks highlight on hover only; `aria-valuenow` reports it to screen readers only). If that's a deliberate aesthetic, fine — just confirming it's a choice, not an omission.

## References

- `src/renderer/src/components/chat/ChatMinimap.vue` — even pitch, scrolling rail, `useElementSize`, `indexAt`
- `src/renderer/src/features/chat-page/model/minimapTicks.ts` — `resolveMarkPitch` (L97), sqrt compression (L54, L83)
- `test/renderer/components/ChatPage.test.ts:3258,3315,3388` — ResizeObserver mock that breaks
- `src/renderer/src/assets/style.css:1135` — global 8px scrollbars (basis of my retraction)
- `src/renderer/src/features/chat-page/ChatPage.vue:203` — float-layer `v-if` (item 2)
- `src/renderer/src/components/chat/ScrollToLatestPill.vue:32` — skipped leave transition (item 2)

Reviewed at `873d2bd0d` (CI `test-renderer` red on this head; `static`, `test-main`, `test-native-memory`, `build` green).

zhangmo8:
## Review follow-ups

@zerob13 — all three addressed:

1. **Rail vs scrollbar** — fixed in `499f3ead1`: the rail now sits 20 px in from the viewport's right edge, so its hit area clears a classic ~17 px Windows scrollbar rather than only the app's 8 px one. Note the map was restyled after your review: marks are hairline bars on a constant 16 px pitch, width encodes message length relative to the longest message, and the map scrolls on its own once the marks no longer fit.
2. **Pill leave transition** — fixed in `5619dc7f2`: the float layer is now always mounted with only its children conditional, so hiding the pill plays its leave animation instead of being unmounted together with its parent.
3. **Badge semantics** — intended, and signing off on it: the badge counts loaded messages that reach below the fold as a "what is under me" navigation affordance, not an unread counter, and it does under-report while older history is unloaded. The PR description states this explicitly so it is a documented choice rather than a surprise.

@coderabbitai — two fixed, one stale:

- **`returnToLatest`** — fixed in `5619dc7f2`: it now calls `cancelPendingMessageJumps()`, which moves the jump generation on and cancels every scheduled retry (resolving `false`) *before* the bottom request, so a pending map or Spotlight jump can no longer override the user's own navigation.
- **Retry returning a deferred result** — fixed in the same commit: `messageJumpTimers` entries carry the resolver of the promise their caller awaits, the retry branch returns that promise and resolves it with the outcome of the chain, and every cancellation path (a newer jump, return-to-latest, unmount) resolves `false`. `focusPendingSpotlightMessageJump` therefore always receives a definitive outcome and drops its pending jump.
- **`minimapTicks.ts:85` (tick-interval distance)** — stale: that resolver was removed when the rail moved to constant-pitch slots. Positions now resolve with `floor(offset / pitch)`, so there is no per-tick interval distance calculation left to adjust, and the unequal-height case it asked a regression test for no longer exists by construction.

Verification: `format` / `lint` / `i18n` / `typecheck` clean; ChatPage's 98 tests plus the map, pill and virtualization suites pass. One regression I introduced in `873d2bd0d` — VueUse's `useElementSize` added a second ResizeObserver that the ChatPage harness drives with no entries, which broke two geometry tests — is fixed by the constant-pitch change in `499f3ead1`.

zerob13 (approved):
**Verdict: Approve.** The blocking test failure and the pill transition issue are both fixed, the badge semantics question is settled with the author's sign-off, and the delta is a net simplification. Two comment-only nits below — feel free to address in a follow-up.

## What changed since last review (`873d2bd0d` → `5619dc7f2`)

- `499f3ead1` — The message map switches from adaptive pitch (14–44px, rail-height-dependent) to a constant 16px pitch. This deletes `useElementSize`/ResizeObserver from ChatMinimap entirely, which is what fixed the red `test-renderer` job: nothing in the chat tree reads `entries[0].contentBoxSize` anymore, so the test harness firing RO callbacks with empty entries is harmless again. Also moves the rail from `right-3` to `right-5` (20px inset).
- `5619dc7f2` — (a) The composer float layer is now always mounted with only its children conditional, so the scroll-to-latest pill gets to play its leave transition instead of being unmounted with its parent. (b) `messageJumpTimers` entries now carry the awaiting caller's resolver; a new `cancelPendingMessageJumps()` cancels retries, resolves their callers with `false`, and bumps the jump generation — wired into `returnToLatest`, `startMessageJump`, and `onUnmounted`.

## Previous findings

1. **P1 test-renderer red — fixed.** `ChatMinimap.vue` no longer uses `useElementSize`; marks sit on a constant `MARK_PITCH = 16` (src/renderer/src/components/chat/ChatMinimap.vue:47-48). I re-ran the two previously failing geometry tests locally plus the full ChatPage suite: all green, and `gh pr checks` shows `test-renderer` passing at this head.
2. **P2 pill leave transition skipped — fixed.** The float-layer container's `v-if` is gone (src/renderer/src/features/chat-page/ChatPage.vue:202-206); only the pill's own inner `v-if` toggles inside its `Transition`, so the leave animation plays. Checked the always-mounted fallout: the dock renders an empty guarded div when nothing is docked (ChatInteractionDock.vue:6, :61), the container is `pointer-events-none`, and no test asserted the layer's absence.
3. **Badge semantics — resolved.** Thanks for the explicit sign-off; the "messages below the fold, not unread" semantics documented in the PR description is a sound navigation affordance.
4. **Keyboard walk rail-scroll / visual position cue — not addressed.** Still no rail scroll-to-mark on arrow-key walk and still no visual current-position styling (aria-valuenow only, ChatMinimap.vue:147). Fine as a follow-up; not blocking.
5. **CodeRabbit's two comments — both fixed** (returnToLatest now cancels pending jumps before the bottom request, ChatPage.vue:662-665; retries now settle the caller's promise with the chain outcome, ChatPage.vue:979-998, and every cancel path resolves `false`). The tick-interval comment is moot now that pitch is constant.

## New findings

1. **P3 (nit):** The new doc comment in src/renderer/src/components/chat/ChatMinimap.vue:18-22 says Windows "still draws a classic ~17 px" scrollbar — but this app styles all scrollbars to 8px globally (`*::-webkit-scrollbar { width: 8px }`, src/renderer/src/assets/style.css:1135), which applies on Windows in Electron too. The 20px inset is harmless; just the comment's justification doesn't hold for this app.
2. **P3 (nit):** The comment at src/renderer/src/features/chat-page/ChatPage.vue:946-948 ("Clearing the timers above…") still refers to the attempt-0 clearing block this commit removed from `jumpToMessage`; the clearing now lives in `cancelPendingMessageJumps`. The re-check it justifies remains correct — only the "above" pointer is stale.

## Details / references

- Delta reviewed: `git diff 873d2bd0d..5619dc7f2` — ChatMinimap.vue, ChatPage.vue, minimapTicks.ts (+ its two test files); −87/+67 lines, `resolveMarkPitch` and its 4 tests deleted.
- Deferred-promise chain traced end-to-end: every path (success, max retries, superseded after `nextTick`/rAF, cancel during the post-timer yield) settles the awaited promise exactly once; unmount clears all timers via `cancelPendingMessageJumps` (ChatPage.vue:1704-1707). No timer leaks found.
- Local test runs at this head (throwaway worktree, pnpm):
  - `pnpm vitest run test/renderer/components/ChatPage.test.ts -t geometry` → 3 passed (the 2 previously failing tests included)
  - `pnpm vitest run ChatPage.test.ts ChatMinimap.test.ts minimapTicks.test.ts` → 121/121 passed
  - `pnpm vitest run ScrollToLatestPill.test.ts` → 4/4 passed
- CI at `5619dc7f2`: `test-renderer`, `test-main`, `test-native-memory`, `build`, `static` (lint/format/i18n/typecheck), `package-required`, `package-impact`, CodeRabbit — all pass.

zerob13 (approved):
## Re-review: `b286a2179` — "give the message map its own column" (delta from `5619dc7f2`)

**Verdict: this structural fix is the right one.** Moving the rail out of the scroll container's overlay and into its own flex column (ChatPage.vue:44) resolves the scrollbar-gutter overlap at the root: the marks no longer float over the scroll area on any platform, the message list keeps its own scrollbar gutter, and the old "Windows draws a ~17px scrollbar" workaround comment is gone (ChatMinimap.vue:24). This is simpler than the inset-tuning it replaces — good call.

Verified locally at this head: ChatPage 98, ChatMinimap 11, minimapTicks 12, ScrollToLatestPill 4 — all pass. CI green.

Two non-blocking notes:

1. **The rail now appears with nothing to scroll.** `showChatMinimap` dropped the overflow condition (ChatPage.vue:1160), so a conversation with two short messages that fully fit the viewport still gets a permanent 64px column, and jumps in that state are no-ops. The tradeoff is real — re-adding the column on overflow would reflow message text mid-scroll, which is worse — but consider hiding it until the conversation actually overflows and accepting the single one-time width change, or some middle ground. Author's call; happy either way.

2. Carried from the previous round: the "timers above" comment at ChatPage.vue:942 reads stale against the current code flow — worth a rewrite next time you're in the file.

---

<details>
<summary>Detailed analysis</summary>

- **Layout**: `chat-viewport-region` becomes `flex flex-row`; the scroll container gets `flex-1` and the minimap column `w-16 shrink-0` (ChatMinimap.vue root). With `ChatMinimap` outside the scroll container, the rail keeps its position while the list moves — the load-more/jump flows are unaffected.
- **Visibility gate**: `ticks.length > 1` vs the old `> 1 && totalHeight > viewportHeight + 1`. The rail scrolls itself when marks no longer fit (`overflow-y-auto`), so the overflow gate is not needed for correctness — only for the "useless when nothing scrolls" concern in note 1.
- **Marks**: anchored `right-0` (was `left-0`), rail gains `pr-2` so the hover card keeps its spacing. Pointer handling unchanged — clicks resolve by index, marks stay `pointer-events-none`.
- **Tests**: no test changes needed — the visibility gate is a one-line computed and the component/model tests already cover the rail behavior (27 passing across the three touched suites, plus 98 ChatPage tests).
</details>

zhangmo8:
Thanks — both notes addressed.

1. **Stale comment** — rewritten in `ba4930d5f`: the "timers above" pointer now names `cancelPendingMessageJumps`, which is where the cancellation actually lives.

2. **Rail visible with nothing to scroll** — decided to keep it always on (`6bffa122b`), and the reasoning changed slightly with the follow-up below: the map now carries a reading-position marker, so even a conversation that fits the viewport answers "where am I" rather than only offering jumps. Combined with the column being reserved anyway, the alternative — gating on overflow — would reflow message text mid-scroll for a map that is still informative. `showChatMinimap` is now `ticks.length > 0`, i.e. it appears with the conversation's first message and never toggles again.

3. **Follow-up taken, not deferred** — the same commit adds both halves of what you flagged:
   - **Visual position cue**: the mark at the top of the viewport is drawn at `bg-foreground/70` against `bg-muted-foreground/50` for the rest, with hover staying brightest at `bg-foreground`. The hierarchy is resting < current < hover.
   - **Rail follows the walk**: a `watch` on the reading-position index scrolls the rail to the mark when it is off screen. It only acts when the mark is actually outside the visible slice (`resolveRailScrollTop` returns `null` otherwise), so scrolling the rail by hand to browse is not undone the next time the conversation moves.

`resolveRailScrollTop` is a pure function in `minimapTicks.ts` with its own cases (visible → no-op, above → scroll up, below → the smallest offset that shows it, unmeasured rail → no-op); the component cases cover the three-way styling. Gates clean, ChatPage 97/98 (the one failure is the known load-dependent timeout in the trace-diagnostics test, which passes in isolation), ChatMinimap + minimapTicks 28/28.

zerob13 (approved):
## Re-review: \`6bffa122b\` — reading-position marker on the map (delta from \`b286a2179\`)

**Verdict: Approve.** This delta closes both notes from the last round, the reasoning for keeping the rail always on is sound, and the new rail-following logic is correct. CI was still pending at review time; all three touched suites pass locally.

## What this delta does

Two commits since the last approval:

- \`ba4930d5\` — rewrites the stale "timers above" comment in \`jumpToMessage\`; it now points at \`cancelPendingMessageJumps\`, which is where the cancellation actually lives.
- \`6bffa122b\` — the map now shows where you are: the mark at the top of the viewport is drawn brighter than the rest, and the rail scrolls itself to keep that mark visible without undoing manual browsing of the rail.

## Verified

1. **Stale comment — closed.** The rewritten comment (ChatPage.vue:942) accurately describes the re-check it justifies: a jump that awaited the DOM while a newer one started re-checks its generation before acting, because \`cancelPendingMessageJumps\` only cancels the timers that existed at that moment.
2. **"Rail with nothing to scroll" — settled.** Keeping it always on (\`showChatMinimap\` now \`ticks.length > 0\`) is a deliberate, documented choice: with a reading-position marker the map answers "where am I" even in a conversation that fully fits, and an overflow gate would reflow message text mid-scroll. Accepting per that reasoning.
3. **Position marker — correct.** Three-way \`markClass\` hierarchy (resting \`bg-muted-foreground/50\` < current \`bg-foreground/70\` < hover \`bg-foreground\`) uses existing design tokens; the current position is distinguishable at a glance and hover stays the brightest state.
4. **Rail auto-scroll — correct and minimal.** \`resolveRailScrollTop\` returns \`null\` when the mark is already visible, so hand-scrolling the rail is preserved; otherwise it scrolls the smallest distance (mark above → align to top; below → smallest offset that reveals it), clamped at 0, and no-ops before the rail is measured. The \`watch\` runs \`flush: 'post'\` so measurements are read after the DOM settles.
5. **Tests — proportionate and passing locally.** Four pure-function cases for \`resolveRailScrollTop\` (visible no-op / above / below / unmeasured rail) plus two updated component cases for the styling hierarchy. Local runs at this head: ChatPage 98/98, ChatMinimap 12/12, minimapTicks 16/16.

## Notes (non-blocking)

- CI \`test-*\` / \`build\` / \`static\` were still pending at this head when I reviewed; the local equivalents pass, so nothing further to wait on from my side.

## References

- \`src/renderer/src/components/chat/ChatMinimap.vue\` — markClass hierarchy, rail-following watch
- \`src/renderer/src/features/chat-page/model/minimapTicks.ts\` — resolveRailScrollTop (L106-128)
- \`src/renderer/src/features/chat-page/ChatPage.vue\` — supersede comment rewrite (L942), showChatMinimap gate (L1161)
- \`test/renderer/components/ChatMinimap.test.ts\`, \`test/renderer/features/chat-page/model/minimapTicks.test.ts\` — 28 new/updated cases

Reviewed at \`6bffa122b\`.

PR #2323, opened 2026-09-17
fix(cheaper-inference): point provider website links at browsable pages

aiapienthusiast:
Follow-up to #2314, addressing the review notes on the `websites` block.

| field | before | after |
| --- | --- | --- |
| `apiKey` | *(empty)* | `https://www.cheaperinference.com/dashboard/keys` |
| `official` | `https://cheaperinference.com/` | `https://www.cheaperinference.com/` |
| `docs` | `https://cheaperinference.com/docs` | `https://www.cheaperinference.com/docs` |
| `models` | `https://api.cheaperinference.com/v1/models` | `https://www.cheaperinference.com/markets` |

- the key field was empty, so the "get an API key" link had nowhere to point; `/dashboard/keys` is the key management page, and it redirects to the login page when signed out
- the bare domain redirects to the `www` host, so the two links now point straight at the final URL
- `models` pointed at the raw JSON endpoint, which is not a page to send a user to in a browser; `/markets` is the human-readable catalogue

`baseUrl` and `defaultBaseUrl` are unchanged - those are API endpoints and stay on `api.cheaperinference.com`.

All four URLs returned 200 when checked on 2026-09-17. `test/main/provider/defaultProviders.test.ts` is updated to match.

## Summary by CodeRabbit

- **Updates**
  - Updated Cheaper Inference provider links to use the current official website and documentation pages.
  - Added a direct link to the provider’s API key dashboard.
  - Updated the models link to open the provider’s markets page for browsing available offerings.


zerob13 (approved):
**Review: LGTM — approving.** A small, correct follow-up to #2314 that fixes all three review notes on the `websites` block. I re-verified every claim in the PR description independently.

## What changed

Four link values in the `cheaper-inference` entry of `DEFAULT_PROVIDERS` (`src/main/provider/defaults.ts`), plus the matching assertions in `test/main/provider/defaultProviders.test.ts` — 2 files, +8/−8. `baseUrl` / `defaultBaseUrl` are untouched and stay on `api.cheaperinference.com`, which is correct since those are API endpoints, not pages.

## Why each change is right

- **`apiKey`**: was empty, so the Settings page's "How to get API key → Cheaper Inference" link (rendered from `websites.apiKey` in `ProviderApiConfig.vue`) had nowhere to point. `/dashboard/keys` is the key management page; it redirects to the login page when signed out, which is exactly what a key dashboard should do. This also brings the provider in line with Deepseek / API Route / RunInfra / GreenPT, which all link a key page here.
- **`official` / `docs`**: the bare domain 308-redirects to `www`, so pointing straight at the final URL saves a hop.
- **`models`**: previously pointed at the raw `/v1/models` JSON endpoint, which returns 401 without a key — nothing to show a user in a browser. `/markets` is the human-readable catalogue, matching the pattern of API Route (`/pricing`) and RunInfra (`/inference-api`).

## Independent verification (2026-09-17)

| URL | Result |
| --- | --- |
| `https://www.cheaperinference.com/` | 200 |
| `https://www.cheaperinference.com/dashboard/keys` | 307 → `/login?callbackUrl=%2Fdashboard%2Fkeys` (expected signed-out behavior) |
| `https://www.cheaperinference.com/docs` | 200 |
| `https://www.cheaperinference.com/markets` | 200 |

`pnpm vitest run --config vitest.config.ts test/main/provider/defaultProviders.test.ts` → 10/10 passed on the PR head (`fa48d5d`).

No over-engineering, no test bloat (existing assertions updated to match, nothing new), no breaking changes — these are pure display strings; nothing in the codebase fetches them programmatically, they are only opened in a browser.

zerob13 (approved):
**Verdict: Approve.** Clean, minimal fix — all four links verified live, no behavior change, test kept in lockstep.

## What this PR does

`Cheaper Inference` is one of the built-in (disabled-by-default) providers. Three of its four website links were wrong or missing:

- `apiKey` was **empty** — every other built-in provider links to its dashboard key page here, so the "get a key" entry point was simply absent for this one.
- `models` pointed to `https://api.cheaperinference.com/v1/models` — that's the raw OpenAI-compatible API endpoint, which returns **401 Unauthorized** in a browser. A user clicking it sees an auth error, not a model list.
- `official`/`docs` used the bare apex domain; the site canonicalizes to `www.`.

## Verification (all done against the live site)

| Link | New value | Result |
|---|---|---|
| official | `/` on `www.` | 200, real page |
| apiKey | `/dashboard/keys` | redirects to sign-in, then the keys page — same pattern as DeepSeek / API Route / RunInfra |
| docs | `/docs` | 200, "Documentation · CheaperInference" |
| models | `/markets` | 200, "Markets · CheaperInference" |

The site is not an SPA that 200s everything (a random path returns a 404 page), so these are genuine pages, not catch-all routes.

## Review criteria

- **No over-engineering**: 4 URL strings + the mirrored test assertion. Nothing else.
- **No breaking change**: `baseUrl` and `defaultBaseUrl` (the actual API host) are untouched — this only affects user-facing metadata.
- **No excessive tests**: the existing assertion in `test/main/provider/defaultProviders.test.ts` is updated in lockstep; nothing new was bolted on.
- **Consistency**: the `apiKey` URL now follows the exact pattern used by every sibling provider (`dashboard/keys`-style page).
- **Ran the test**: `vitest run test/main/provider/defaultProviders.test.ts` — 10/10 passed.

## One non-blocking observation

In the current tree, `websites.official`, `websites.docs` and `websites.models` are passed through `providerSettings` (`src/main/provider/settings.ts:81`) but the renderer only consumes `websites.apiKey` and `websites.defaultBaseUrl` today (`ProviderApiConfig.vue`). These fields are still worth keeping accurate since they're part of the exported provider shape and any future UI / registry flow will consume them — just noting they don't currently render anywhere.

---
*Sources: diff of `fa48d5d7` vs `origin/dev` (2 files, +8/-8); live curl checks of all six old/new URLs; `src/main/provider/defaults.ts:39-47`, `test/main/provider/defaultProviders.test.ts:7-21`, `src/renderer/settings/components/ProviderApiConfig.vue:334-369`, `src/main/provider/settings.ts:73-83`.*

Issue #2321, opened 2026-09-17
[Feature] 百智云 Agent Toolkit 接入与原生 Key 配置提案


ct-jaryn:
### What do you need? / 你需要什么？

DeepChat 维护团队好，我是百智云 Agent Toolkit 的维护及运营人员，使用 `ct-jaryn` 代表百智云提出接入建议。希望与社区合作维护一个可选的服务接入入口，让用户完成“选择服务 → 阅读说明并确认安装 → 按客户端引导显式启用并配置自己的 API Key → 确认连接状态”。

Agent Toolkit 是托管的商业工具服务，不是模型供应商。用户需要自己的百智云账号和 Key，调用可能消耗服务额度；公开的[集成仓库](https://github.com/chaitin/baizhi-agent-toolkit)提供配置、文档和测试，不包含托管后端源码。

#### 建议复用现有能力

DeepChat 已支持 Streamable HTTP，我们不是申请新增 MCP 协议实现。服务使用固定端点 `https://agent-toolkit.app.baizhi.cloud/mcp` 和用户自己的 Bearer Key，不需要本地命令、安装脚本、OAuth 代登录或自动批准规则。

我们注意到已合并的 [Serply 预设 #2299](https://github.com/ThinkInAIXYZ/deepchat/pull/2299)，也阅读了当前 `dev` 的[用户插件创作说明](https://github.com/ThinkInAIXYZ/deepchat/blob/c66b36d65b59251ef1cf109e2f608387cabaeea3/docs/features/user-plugins/authoring.md)。后者可用 `bearer_token_env_var` 声明凭据，再由插件的 MCP 设置表单收集输入，复用现有 SecretStore；这比要求用户在普通 Header 编辑器里填写完整凭据更符合我们希望提供的体验。

想请维护者确认推荐路线：是否适合先维护一个只含远程 MCP 声明的用户插件，并按项目接受的方式提供可发现的入口或接入文档？如果更推荐默认禁用的内置预设，我们愿意配合，但希望先确认它应如何复用上述凭据输入机制，不把静态 Header 占位符当成已有的隐藏 Key 引导。

[...]

PR #2324, opened 2026-09-18
feat(plugin): add Baizhi MCP example

ct-jaryn:
## Summary

Following the invitation in [#2321](https://github.com/ThinkInAIXYZ/deepchat/issues/2321#issuecomment-5724151875), this PR adds an installable, MCP-only Baizhi Cloud Agent Toolkit user-plugin example. I maintain Agent Toolkit integrations and operations on behalf of Baizhi Cloud through `ct-jaryn` and would be glad to work with the DeepChat community on this integration and its ongoing maintenance.

The proposal discussed both a default preset and a user plugin. This PR takes the smaller user-plugin route: it reuses the existing password input and SecretStore binding without adding a vendor-specific UI or extending core authentication. It does **not** add a bundled plugin, default MCP preset, MCPRouter listing, or automatic installation. If a different location or distribution approach is preferred, we are happy to adjust.

## Changes

- Add `examples/user-plugins/baizhi-agent-toolkit/` with a declarative manifest, `.mcp.json`, English/Chinese instructions, and a copy of the project's Apache-2.0 license.
- Declare one Streamable HTTP server at `https://agent-toolkit.app.baizhi.cloud/mcp`, with `bearer_token_env_var: "BAIZHI_API_KEY"`. The existing importer generates the Bearer template and required-variable form.
- Link the example from the user-plugin authoring guide.
- Add one regression test to the existing lifecycle suite. It packages the actual example as ZIP, inspects and installs it disabled, enables it, rejects an empty credential, configures and replaces synthetic credentials, replaces the installed snapshot, reinitializes the service, and uninstalls without changing unrelated MCP servers.

No runtime/UI code, dependencies, provider configuration, default enablement, or permission policy is changed.

## Installation and credential flow

The documented path is a ZIP containing **only the example directory**, including its hidden manifest and MCP file:

```text
Plugins → Install from ZIP → Review and select MCP → Install (disabled)
→ Explicitly enable → Enter own Key in the existing password field
→ Save and reconnect → Check the plugin detail page's MCP status
→ Built-in-agent conversation: Advanced Settings → Plugins → Confirm discovered tools
```

This is intentionally not documented as Git installation from the full DeepChat repository. At the reviewed `dev` revision, Git import archives and checks the entire repository before selecting its subdirectory; that repository exceeds the 4,096-entry plugin archive limit. The small ZIP uses the existing checks without changing their limits. No new UI layout is introduced.

## Security, service and licensing boundaries

- No API keys are included in the package, tests, or this PR. Users provide their own credentials locally. The MCP configuration keeps `Bearer ${BAIZHI_API_KEY}`; the value is separately bound by the existing SecretStore. User plugins do not implicitly read the host process environment.
- The package has no local commands, Skills, hooks, setup scripts, auto-approval rules, or automatic enablement. The instructions explain review, rotation, update, disablement and removal.
- Tools are dynamically discovered. The three suggested starting tools in the instructions are **not** a configured allowlist and do not override DeepChat's tool-selection or permission controls.
- Agent Toolkit is a separately operated commercial service. Calls may consume credits and send the corresponding tool inputs to Baizhi Cloud. The docs disclose account, fee and data-use considerations without making unverified retention or residency claims.
- The example files follow DeepChat's Apache-2.0 license. The hosted backend is not included, and this contribution does not imply DeepChat certification or endorsement.

## Validation

Reviewed base: `3313dc69a597bb428afcffda861eb9bd2e92b337` (`dev`, package version `1.1.2-beta.5`). Local environment: macOS, Node `24.19.0`, pnpm `10.34.5`; dependencies installed from the frozen lockfile with lifecycle scripts disabled.

- `pnpm run plugin:validate -- --plugin-root examples/user-plugins/baizhi-agent-toolkit` — passed, one HTTP server, no compatibility findings.
- `pnpm exec vitest run --config vitest.config.ts test/main/plugin/userPluginLifecycle.test.ts test/main/plugin/userPlugins.test.ts test/main/mcp/settings.test.ts test/renderer/components/OfficialPluginDetailPage.test.ts test/renderer/components/PluginsCatalogPage.test.ts` — **70 tests passed across 5 files**.
- A subsequent repeat hit the default 10-second timeout in two unchanged Feishu renderer tests (68 passed, 2 timed out; no assertion failures). The 3 main-process suites were rerun separately and passed **43/43**. Rerunning the 2 renderer files with `--project renderer --maxWorkers 1 --testTimeout 30000` passed **27/27**. No test or application timeout configuration was changed in this PR.
- `pnpm run format:check` — passed for all matched files; documentation is excluded by the existing formatter configuration and was reviewed separately.
- `pnpm run lint` — passed.
- `pnpm run typecheck` — both node and web checks passed.
- `pnpm run i18n` — passed; no locale strings were added or changed.

The new lifecycle test uses the real package inspector, installer, MCP settings and SecretStore code with the existing mocked Electron storage/encryption and connection supervisor. It checks configuration boundaries and service-instance reinitialization, **not** native OS encryption, disk/process restart, a real network connection, or interactive password-field operation. The two renderer suites are existing regression coverage, not an end-to-end test of this plugin's credential UI.

Not performed: packaged Electron build/start, full desktop GUI acceptance, Windows/Linux execution, real Baizhi API-key connection, production tool/model calls, or the entire repository test suite. No production credentials or billable operations were used. We will cooperate on further acceptance testing and maintainer feedback.

Related issue: #2321.

## Summary by CodeRabbit

- **New Features**
  - Added an optional Baizhi Cloud Agent Toolkit user-plugin example for hosted search, webpage, and document tools through an MCP connection.
  - Added English and Simplified Chinese setup guides covering ZIP installation, API-key management, permissions, troubleshooting, licensing, and potential usage charges.
  - Added guidance for credential rotation, disabling, updating, and uninstalling the plugin. The example is not bundled or enabled automatically.

- **Tests**
  - Added coverage for plugin installation, credential validation and protection, persistence, updates, and complete removal.


zerob13 (approved):
## Review: Baizhi MCP example

**Verdict: no blocking issues.** Purely additive (+513 / −0): an MCP-only example package, a docs pointer, and one lifecycle test. No product code changes.

### What I verified against the codebase

- `bearer_token_env_var` in `.mcp.json` is an existing supported field (`src/main/plugin/userPluginPackage.ts:240`); it maps to `Authorization: Bearer ${BAIZHI_API_KEY}` plus `requiredVariables` — exactly what the manifest declares and the test asserts.
- The docs claim "uses the existing password field" is accurate: plugin MCP setup renders each required variable as a password input (`src/renderer/src/pages/plugins/UserPluginDetails.vue:192`).
- All manifest keys used here (`author`, `license`, `mcpServers`, …) are in the supported set (`src/main/plugin/userPluginPackage.ts:335`).
- The relative docs link `../../../examples/user-plugins/baizhi-agent-toolkit/README.md` resolves correctly from `docs/features/user-plugins/`.
- `fflate` (used by the test to build the ZIP) is already a runtime dependency — no new dependency added.
- Ran `pnpm vitest run test/main/plugin/userPluginLifecycle.test.ts`: **8/8 pass**, including the new test.

### Standard checks

- **Over-engineering:** none. The package is a manifest, one remote MCP declaration, docs, and a license. No new abstractions or product-code changes.
- **Consistency:** follows the user-plugin conventions and the authoring doc structure, and the precedent of vendor-contributed MCP integrations (#2299). The README is unusually forthcoming: discloses the commercial nature, where tool inputs are sent, that calls may be billed, and that this is not a DeepChat endorsement.
- **Breaking changes:** none (0 deletions, no behavior changes).
- **Excess tests:** one test, not a suite — acceptable, see note 1.
- **Logic / security:** disabled by default, no hooks, no `autoApprove`, key handled via the env-var binding and SecretStore; the test asserts the raw key never appears in the plugin store, MCP settings, or service state. Good.

### Minor, non-blocking notes

1. The new test re-walks ground already covered by the generic lifecycle test ("installs disabled, preserves host MCP identities…"): install-disabled → enable → secret binding → rotation → uninstall cleanup. The genuinely new coverage is (a) round-tripping the *actual shipped example files* through inspect/install, and (b) secret-binding persistence + setup-variable clearing across a same-plugin replacement and a service reload. Dropping the duplicated rotation loop would cut ~30 lines without losing coverage.
2. The test hard-codes the vendor endpoint URL, so the test breaks if Baizhi moves the endpoint. For a shipped example that's arguably correct (docs, artifact, and test should move together) — just noting the coupling.

Reviewed at `ac583a9`.

ct-jaryn:
Thanks for the review and the approval. Both notes are addressed in 51d112f, along with the one actionable CodeRabbit comment.

**zh-CN credential wording (CodeRabbit inline)** — fixed. The Chinese line had its two nouns swapped relative to `README.md` ("no MCPRouter account, intermediary key or OAuth sign-in"): it asked the reader for an MCPRouter key and an intermediary account. It now reads `此流程不使用 MCPRouter 账号、中间代理 Key 或 OAuth 登录。` The other two MCPRouter mentions were already consistent with the English text.

**Note 1, duplicated test coverage** — done, with one deliberate deviation. I dropped the rotation loop: the generic lifecycle test already binds a secret, rebinds it, and re-checks the binding, so the second iteration added nothing. I kept the four `not.toContain` assertions that were inside that block, because they are not fully duplicated — the generic test checks the raw key against `store.read()`, the migration snapshot and the wrapped secrets, whereas this block also checks the plugin record (`service.get(id)`) and the MCP server configs. That is a smaller cut than your ~30-line estimate: the loop was 9 lines and only the second iteration was redundant. Happy to trim further if you would prefer the duplicated install → enable → uninstall sequence reduced as well — I stopped there because those assertions are what pins the shipped-ZIP round-trip.

**Note 2, hardcoded endpoint** — keeping it, for the reason you gave: it is an expected contract for a shipped example, and the docs, the artifact and the test should move together when the endpoint moves.

**Docstring coverage (CodeRabbit pre-merge check)** — not fixing, deliberately. The check is diff-scoped and reports "1 functions across 1 files": the new test in `userPluginLifecycle.test.ts`. That file, and the test suite around it, carry no docstrings at all, so adding one here would cut against the surrounding convention, and the 80% threshold is a CodeRabbit default rather than a DeepChat rule. I did not add a docstring or change test config to satisfy it. Glad to revisit if DeepChat does want docstrings on tests.

**What I ran** — `test/main/plugin/`: 8 files, 182 tests pass, including this file's 8. Caveat: my local toolchain is Node 22.22.2 with pnpm 10.34.5, while the repo pins Node >=24.18 <25, so please treat upstream CI as the authoritative gate. No other files changed.

ct-jaryn:
A short correction to my previous note. I said my local run was on Node 22 and that upstream CI should be treated as authoritative. I have since moved to the pinned toolchain and re-ran with a clean `--frozen-lockfile` install under Node 24.21.0 / pnpm 10.34.5:

- `test/main/plugin/`: 8 files, 182 tests pass (this file's 8 included).
- Full suite: 921 passed / 1 failed / 1 skipped; 11591 passed / 1 failed / 5 skipped.

The one failure is `test/main/scripts/buildCli.test.ts > CLI bundle > builds a standalone Node entry and Electron-hosted launchers`, failing with `Library not loaded: @rpath/libnode.137.dylib`. I reproduced it identically on the parent commit `ac583a9` in an isolated worktree, so it is pre-existing and unrelated to this PR. Mentioning it only in case it is a known local macOS packaging issue on your side.

Upstream CI is still the authoritative gate — the above is local evidence only.

ct-jaryn:
Handled both points from the re-review.

**Plugin directory removal (outside-diff comment)** — fixed in 3dc4455. You're right that `uninstall()` removes the plugin's own directory and that the test only checked MCP settings, the plugin store and the wrapped secrets. Added the filesystem assertion via `USER_PLUGIN_INSTALL_DIRECTORY`; it passes against current code.

**Changed replacement archive (inline comment on `userPluginLifecycle.test.ts:227`)** — I did not apply this one, and I would rather show the reason than just push it.

The mechanism you describe is correct: with an unchanged archive, `record.mcpDigests[key] === record.digest` holds, so `activate()` takes the preservation branch and the test only exercises that path. I tried the suggested change locally — building the replacement archive with one extra file so the digest differs — and the test goes red:

```
✗ expect(setup[key]).toEqual([])  →  received ["BAIZHI_API_KEY"]
```

Probing the state after that changed-digest replacement:

```
setup                  : ["BAIZHI_API_KEY"]   # variable awaiting configuration again
serverId               : unchanged
variable bindings      : {}                   # stored binding is gone
customHeaders          : Bearer ${BAIZHI_API_KEY}
raw key present anywhere: false
```

So on a changed-digest replacement — an ordinary plugin upgrade — the stored credential binding is dropped and the plugin asks for the key again, while an unchanged reinstall preserves it. That is current behaviour rather than something this PR introduces, which is why I did not encode it as an expectation in the test.

Flagging it as a question instead: is dropping the binding on upgrade intended? If it is, the changed-archive variant needs different assertions (and would pin that behaviour); if it is not, the test becomes a useful regression guard once it changes. The changed-digest path is already exercised elsewhere in this file by the tests that call `f.manifest(...)` before a replacement install. Happy to add whichever variant you prefer.

Two notes on verification, since upstream CI has not run for this branch: `test/main/plugin/` is 8 files / 182 tests green on Node 24.21.0 + pnpm 10.34.5, and `format:check`, `lint` and `typecheck:node` are clean. The first version of my new assertion was mis-formatted — `format:check` caught it and I fixed it before pushing rather than leaving it to CI.

zerob13 (approved):
**Re-review of ac583a9..3dc4455: approval holds. Re-approved.**

The delta is 2 commits, 2 files, +13/−11 — no product code, no manifest changes. Both commits address the notes from my earlier review plus CodeRabbit's follow-ups. I verified each against the implementation and re-ran the tests locally.

- `51d112f` (zh-CN wording + test dedupe): correct. The old README.zh-CN.md line had "MCPRouter key" and "intermediary account" swapped relative to README.md ("no MCPRouter account, intermediary key or OAuth sign-in"); it now matches. The dropped two-token rotation loop is genuinely redundant — the generic lifecycle test in the same file already covers rebinding a rotated secret (userPluginLifecycle.test.ts:275–280). The kept `not.toContain` assertions remain unique coverage: they check the plugin record and the MCP server configs, which the generic test (migration snapshot) does not.
- `3dc4455` (uninstall directory assertion): correct and proportionate. The assertion mirrors the implementation exactly: `USER_PLUGIN_INSTALL_DIRECTORY` is the shared constant, `id.slice(5)` matches the derivation in `root()` (userPlugins.ts:105), and the asserted path is precisely what `uninstall()` removes (userPlugins.ts:331). The `rm` is awaited before the check, so no flake risk. Four lines closing a real gap — with `force: true`, a directory-removal regression would otherwise have stayed green.

CodeRabbit's re-review: the outside-diff comment (assert directory removal) was valid and is addressed by `3dc4455`. On the inline suggestion (replacement with a changed archive) I agree with declining: with a changed digest, `activate()` skips the preservation branch (userPlugins.ts:387) and the stored credential binding is dropped, with setup recomputed (userPlugins.ts:521). That is pre-existing behavior outside this PR's scope; pinning it in an example-plugin test would ossify behavior that may itself be a bug. Keeping it as an open question: is dropping the bound secret on a changed-digest upgrade intended? Worth a separate issue either way.

CI note: no GitHub Actions have run on this PR (all runs sit in `action_required` — fork PR awaiting workflow approval), same as at the previously approved commit. Verified locally instead: `test/main/plugin` — 8 files, 182 tests, all pass on Node 24.18.0.

**Details / references**
- examples/user-plugins/baizhi-agent-toolkit/README.md:22 — English wording the zh-CN line now matches
- test/main/plugin/userPluginLifecycle.test.ts:203–209 — deduplicated secret-boundary assertions
- test/main/plugin/userPluginLifecycle.test.ts:233–235 — new directory-removal assertion
- src/main/plugin/userPlugins.ts:103–105 — pluginId format and `slice(5)` directory derivation
- src/main/plugin/userPlugins.ts:331–334 — uninstall removes the `user/<id>` directory
- src/main/plugin/userPlugins.ts:387, 521 — unchanged-digest preservation branch; setup recomputation
- src/shared/pluginPaths.ts:2 — `USER_PLUGIN_INSTALL_DIRECTORY = 'user'`

zerob13 (approved):
Incremental review of `ac583a9..3dc4455e` (the two commits pushed after the earlier approval). The delta is clean; verified locally.

**What changed**

1. `README.zh-CN.md` — fixes crossed credential wording. The old text read "MCPRouter Key、中间代理账号" (terms swapped); it now reads "MCPRouter 账号、中间代理 Key", which matches the English README: "There is no MCPRouter account, intermediary key or OAuth sign-in in this flow." Correct fix.

2. `test/main/plugin/userPluginLifecycle.test.ts` — two changes:
   - The secret-binding loop now runs a single token instead of two. The remaining assertions still cover the full path: configure → binding stored → token absent from store / plugin service / MCP settings / secret map → binding preserved across ZIP replacement. The only coverage given up is "re-configuring with a *different* token updates the binding", which was redundant with the replacement flow. Consistent with the repo's policy against duplicated test cases.
   - New post-uninstall assertion: the plugin's install directory is removed. It mirrors the production path exactly (`userPlugins.ts` builds `<root>/user/<uuid>/versions/<digest>`), and that directory provably exists mid-test, so this is a genuine regression check for uninstall cleanup — not a tautology that could pass forever.

**Verification**

- Ran `pnpm vitest run --config vitest.config.ts test/main/plugin/userPluginLifecycle.test.ts` at `3dc4455e`: 8/8 passed.

**Notes**

- This delta touches only example docs and one test file; no production code changed. No breaking changes, no over-engineering, no gratuitous new tests. The new assertion is exactly the kind of lifecycle coverage worth keeping.

Issue #2312, opened 2026-09-16
[Feature] 键盘输入时自动聚焦主输入框 (Type-to-focus)


xingkaixin:
### 需求背景与痛点 (Background & Problem)
在日常使用中，当用户翻阅聊天记录、选中并复制消息文本、点击界面空白处，或者刚切回窗口时，主输入框往往处于失焦（blur）状态。
此时若用户直接敲击键盘输入内容（例如开始打字提问或敲拼音），按键会被直接忽略或丢失，用户必须中断打字并用鼠标点击输入框才能重新激活输入，打断了打字与提问心流。

主流桌面 AI 客户端（如 Codex App、Claude Desktop）均具备全局键盘捕获能力，当处于聊天界面且没有其他控件抢占焦点时，敲击键盘即可自动聚焦到输入框并完成输入。

### 期望功能与行为 (Proposed Behavior)
1. **自动捕获并聚焦**：在主会话处于非只读状态且允许输入时，用户在窗口任意位置按下可打印字符、或启动输入法（IME Composition）时，焦点自动且平滑转移至主输入框（`ChatInputBox`）。
2. **首字符无缝保留**：
   - 英文及半角字符直接敲击时，触发聚焦的第一个字符需直接输入到输入框内，不丢字；
   - 中文/多语言输入法启动时，输入法候选浮层应自然挂载在输入框当前光标位置。
3. **已有草稿追加**：若输入框中已有未发送的草稿内容，光标直接定位至草稿末尾，新输入字符在末尾自然追加。
4. **空格键处理**：在聊天视口无焦点时按下空格键（Space），作为普通字符输入处理（聚焦并输入空格），不触发浏览器默认的空格向下滚屏行为。

### 边界与防冲突规则 (Constraints & Edge Cases)
- **已有活动输入控件时豁免**：若焦点当前在其他输入控件上（如顶部聊天搜索栏 `ChatSearchBar`、弹窗 input/textarea、设置项输入框等），绝不抢夺焦点。
- **系统/修饰组合键豁免**：按键包含修饰键（`Cmd/Ctrl + C/V/A/F/K`、`Alt`、单纯的 `Shift/Cmd/Ctrl/Alt` 单键）时，保持原生快捷键逻辑，不转移焦点。
- **导航与功能键豁免**：
  - 滚动键（`ArrowUp/ArrowDown`、`PageUp/PageDown`、`Home/End`）继续保留用于聊天列表滚动；
[...]

PR #2325, opened 2026-09-18
feat(chat): focus the composer on keyboard input

zhangmo8:
## Summary

Implements #2312. When the chat session is editable and no control holds the keyboard, typing anywhere in the view moves focus into the composer — and the character that triggered the focus lands in it, so nothing is dropped. An existing draft is appended to at its end. The new-thread composer gets the same behavior.

Follow-up to review: the dead-key observation is addressed in `093b9741c`.

## BEFORE / AFTER (key routing)

```
BEFORE  (focus on the message list / body, user presses "n" or Space)
  keydown ─► ChatPage.handleWindowKeydown
              ├─ restore window: Space → markKeyboardScrollIntent
              └─ handleSearchKeydown
  → nobody claims it → the keystroke is dropped; the user must click the
    composer first

AFTER   (same conditions)
  keydown ─► useComposerTypeToFocus (registered first, runs first)
              ├─ defaultPrevented / Cmd-Ctrl-Alt chords ......... pass through
              ├─ focus in an editable target or an interactive
              │  control (button, link, menu, modal) ........... pass through
              ├─ composer not focusable (read-only, inert) ..... pass through
              ├─ IME (isComposing | keyCode 229 | 'Process')
              │  or a dead key ('Dead') ....................... focusInput()
              └─ single printable char, incl. Space ............. preventDefault()
                                                                 + focusAndInsertText(key)
  ─► ChatPage.handleWindowKeydown still runs; with the event already
     default-prevented the restore scroll intent stands down
```

## Design notes

- **The routing rules are a pure function.** `resolveComposerTypeToFocusIntent` takes the event plus three booleans and returns ignore / focus-only / focus-and-insert. The exemption contract is the whole feature, so it is unit-tested without mounting a page or an editor.
- **Ownership is checked before intent.** The first draft put the IME test first, which meant a composition running in the chat search box (`keyCode === 229` bubbles to `window`) would have pulled focus into the composer mid-word. Editable target, interactive focus, and the enabled flag are now all checked ahead of the IME branch, and a test pins that order. The IME branch must still precede the printable-character test, because `'Process'` is seven code units long.
- **`compositionstart` is not the entry point**, contrary to the issue's suggestion: with focus on a non-editable element no composition ever starts, so there is no event to wait for. The entry point has to be `keydown`, identifying IME keystrokes by `isComposing` / `keyCode === 229` / `key === 'Process'`.
- **Dead keys take the same branch as the IME.** A dead key produces no text of its own, so ignoring it left the follow-up letter to be inserted as a bare character — "dead key + e" became `e` instead of `é`. Focus-only (no insertion, no `preventDefault`) leaves the platform's pending composition intact so the *next* keystroke reaches an already focused editor and composes natively. Raised in review; see the caveat under Known gaps.
- **Focus and insert are one call.** `focusAndInsertText` reuses `focusInput`'s `focus()` + `setCaretToEnd()` pair, so an existing draft is appended to at the end, then inserts a `{ type: 'text' }` node. A node rather than a string, because a lone space would otherwise be collapsed by the parser — and a separate method rather than `insertRecognizedText`, which trims and drops empty strings.
- **Space stops being a scroll key.** It was in `SESSION_RESTORE_SCROLL_INTENT_KEYS`; it is composer input now, and the claimed key is `preventDefault`ed so the transcript does not scroll. Making it explicit means the two mechanisms no longer depend on listener registration order.
- **No fourth copy of the focus guards.** The renderer had three near-duplicate "is the user typing somewhere" helpers. `lib/keyboardFocus.ts` is the new canonical pair, and `ChatPage`'s local copy now imports from it. `useChatSearch` and `useSidebarSessionShortcuts` keep theirs — different semantics, own tests — so this is a consolidation of one call site, not a sweep.
- **The interactive-control list deliberately omits a bare `[tabindex]`.** The message scroll container is `tabindex="0"`, and it is exactly where focus lands after clicking empty space — the main scenario for this feature.
- **Modal detection cannot require `aria-modal`.** reka-ui 2.10.4 writes `role="dialog"` / `role="alertdialog"` but never `aria-modal`, so the existing sidebar guard's `[role="dialog"][aria-modal="true"]` selector matches essentially nothing. The new predicate matches on role plus the shadcn `data-slot` markers and the popper wrappers.
- **The new-thread composer shares the composable.** It has no event bridge, so the composable owns its own `useEventListener(window, 'keydown')` rather than joining `useChatPageEventBridge`; both pages call it, and it detaches with the component scope.
- **Out of scope, as agreed:** clicking empty space to focus the composer. No pointer handler was added — but the keyboard path still works after such a click, which is the point.

## Verification

- `pnpm run format:check`, `pnpm run lint` (0 warnings, 0 errors), `pnpm run typecheck` (node + web), `pnpm run architecture:renderer-baseline:check` — all clean.
- **2554 tests passed** across the full renderer suite (281 files), including 38 new cases: the routing table (each exemption, the dead-key branch, and the two ordering constraints), the focus predicates against real DOM (scroll container is not interactive; dialog/popover descendants are; `contenteditable="false"` is not editable), and the composable's wiring (insert, space, IME and dead-key focus-only, editable/interactive/disabled exemption, fallback when the composer cannot insert, scope disposal), plus two component cases for `focusAndInsertText`.
- CI green on `093b9741c`: `build`, `static`, `test-main`, `test-renderer`, `test-native-memory`, `package-*`.
- The scroll-architecture guard required registering the new `scrollIntoView` in `ChatInputBox.vue` (2 → 3); the entry now carries a comment that all three scroll the composer itself, not the message list.
- Not added: an e2e spec. The chat smoke specs are gated behind `RUN_PROVIDER_INTEGRATION=true` and a live model, which does not fit a keyboard-routing behavior.

## Needs manual verification — **Windows and Linux keyboard input method testing**

Automated tests cannot cover this: whether an input method delivers its first keystroke to a window whose focus is on a non-editable element is a platform behavior, not something readable from the source. **Please verify on Windows and Linux (and macOS) before this merges.**

For each platform, with a Chinese (or other IME) input method active:

1. Open a session, click on the message list so focus leaves the composer.
2. Type the first letter of a pinyin syllable. Check two things:
   - does a stray Latin character appear in the composer (typing `n` for 你 leaves an `n` behind)?
   - does the candidate window anchor at the composer caret?
3. Record what `keydown` reports for that first keystroke — `key`, `keyCode`, `isComposing`. (`keyCode === 229` / `key === 'Process'` means the IME branch is taken and only focus happens; a plain Latin `key` means the character is inserted literally.)
4. On a layout that composes accents with a dead key — French, German, Spanish, Nordic, US-International, or a Linux compose key — press the dead key and then a letter. The composer should end up with `é`, not `e` and not nothing.

If a stray character shows up on any platform, the fix is a targeted one: remember the character inserted by type-to-focus, and if `compositionstart` arrives in the composer within ~300 ms, undo that single insertion. That mitigation is deliberately **not** in this PR — it adds complexity and a (small) risk of deleting a legitimate character, so it should only land if a platform actually needs it. Report the findings here and I will add it or leave it out.

Also worth a manual pass on the same platforms: plain English typing keeps the first character; Space types a space instead of scrolling; typing is not stolen while the chat search box or a dialog input has focus; `ArrowUp/Down`, `PageUp/Down`, `Home/End`, `Esc`, `Tab` and `Cmd/Ctrl+C/V/A/F` keep their native behavior.

## Known gaps

- **The dead-key fix rests on an unverified premise.** Routing `Dead` to focus-only only helps if the platform's pending composition survives the focus change — and TipTap's `focus()` lands on the next animation frame, not synchronously. If a platform drops the pending dead key instead, the outcome changes from "wrong letter" to "no letter", which is not obviously better. That is why item 4 above is in the manual list. If it fails there, the fix is to focus synchronously rather than to revert this.
- **AltGr characters** (Ctrl+Alt on layouts where that produces `@`, `€`, …) are rejected by the modifier exemption, so they do not trigger focus. If that matters, the exemption can special-case `event.getModifierState('AltGraph')`.
- **The `contenteditable` attribute literal ProseMirror injects was not runtime-verified.** The predicate uses the broadest form (`[contenteditable]:not([contenteditable="false"])`) plus `role="textbox"` as a second net, so either literal matches.
- The full local renderer suite took 951 s on the run behind the numbers above (vs ~250 s earlier on the same machine) with everything passing, i.e. the intermittent load slowdown that has been seen on this suite, not a hang in this change.
- No commit hooks were skipped and no force-push was used.

## Summary by CodeRabbit

- **New Features**
  - Typing anywhere in an eligible chat area now automatically focuses the composer and inserts the typed character, including spaces.
  - Added support for this behavior on chat and new-thread pages.
  - Preserves typing in text fields, interactive controls, shortcuts, navigation keys, and IME composition.

- **Bug Fixes**
  - Prevents the Space key from scrolling the page when it is intended as message input.

- **Tests**
  - Added coverage for typing, focus protection, IME input, disabled states, and special keyboard interactions.


zerob13 (approved):
# Review — type-to-focus for the composer

**Verdict:** Clean PR. No blocking or P1 issues. Approving — the notes below are informational, no action required.

## What this adds (for readers who haven't followed #2312)

When you're reading a conversation and just start typing, the keystroke used to be silently dropped unless the composer already had focus — you had to click it first. After this PR, typing any printable character anywhere in the chat view moves focus into the composer **and keeps the character you typed**, so nothing is lost. It works on the new-thread page too, and it doesn't fire when it shouldn't: shortcuts (Cmd/Ctrl/Alt), navigation and function keys, another open text field (e.g. the chat search box), focused buttons/links/menus, and open dialogs are all left alone. Space now goes into the composer instead of scrolling the transcript — exactly what the issue asked for.

## Why the design holds up

1. **The routing rules are a tested pure function.** `resolveComposerTypeToFocusIntent` takes the key event plus three booleans and returns ignore / focus-only / focus-and-insert. Every exemption is pinned by a unit test without mounting a page or an editor. This is the whole risk surface of the feature (stealing a keystroke from something else), and it's now a documented contract.
2. **Ownership is checked before intent.** The ordering matters and they got it right: an IME composition running in another field (keyCode 229 bubbles to `window`) must not pull focus mid-word, so the editable-target check precedes the IME branch. A test pins that order.
3. **IME handling is the conservative correct choice.** Input-method keystrokes (`isComposing` / keyCode 229 / `'Process'`) focus the composer without inserting anything and without `preventDefault` — the composition owns the text. Non-IME printable chars are inserted via a TipTap text **node** rather than a string, which is what makes a lone Space survive the parser. Good attention to detail.
4. **No over-design.** One pure resolver, one thin composable, one shared predicate module. The predicates (`lib/keyboardFocus.ts`) also replace ChatPage's pre-existing inline copy used by the scroll-intent guard — a consolidation, not new abstraction for its own sake. The scroll-architecture test whitelist is updated to match the one new `scrollIntoView` call.
5. **Enable-state wiring is consistent.** ChatPage gates on read-only / preparing / pending-tool-interaction; NewThreadPage gates on `isSubmittingInput`, which is the same signal that drives its `:editable` prop — so there's no window where a key is `preventDefault`ed and then swallowed by an inert composer. ChatPage never binds `editable`, so it's always true when the box is rendered.
6. **Tests are proportionate.** ~420 lines across a resolver test, a predicate test, a wiring/lifecycle test, and two small additions to existing suites. CI is green including `test-renderer`.

## Minor observations (no action needed)

- **Dead-key accents:** on layouts where an accented char is two keystrokes (dead key + letter), the first key is `'Dead'` → ignored (correct), but the second lands as a plain focus-and-insert of the bare letter — the user gets `e` instead of `é` on the very first keystroke after losing focus. Niche enough to leave; if it ever gets reported, routing `Dead` to focus-only would fix it.
- **Test overlap:** the composable test re-covers several routing cases the resolver test already pins. Mild duplication, justified by the wiring/scope-lifecycle coverage it adds; not worth trimming.
- **Space as a scroll key is gone** from the chat view for keyboard-only users; PageUp/PageDown and the arrows remain. This is the behavior the issue specified, just noting the trade-off for anyone who misses it.

## References

- `src/renderer/src/features/chat-page/model/typeToFocus.ts` — the routing contract (ignore / focus-only / focus-and-insert)
- `src/renderer/src/features/chat-page/composables/useComposerTypeToFocus.ts` — window keydown wiring, shared by both pages
- `src/renderer/src/lib/keyboardFocus.ts` — canonical editable/interactive predicates, reused by the scroll-intent guard
- `src/renderer/src/components/chat/ChatInputBox.vue` — `focusAndInsertText` (~L914), text-node insert
- `src/renderer/src/features/chat-page/ChatPage.vue` — enable-state computed (~L1390), Space removed from restore-scroll-intent keys
- `src/renderer/src/pages/NewThreadPage.vue` — same composable, `!isSubmittingInput` gate
- `test/renderer/features/chat-page/model/typeToFocus.test.ts`, `test/renderer/lib/keyboardFocus.test.ts`, `test/renderer/features/chat-page/composables/useComposerTypeToFocus.test.ts` — the exemption matrix

zerob13 (approved):
# Review — dead-key follow-up

**Verdict:** No blocking issues. Approving.

## What changed

This follow-up fixes the remaining dead-key case in the type-to-focus flow. When the browser reports `KeyboardEvent.key === 'Dead'`, the handler now focuses the composer without inserting text or cancelling the event. The pending dead-key state is therefore left to the platform, so the next letter can compose normally (for example, `Dead` + `e` → `é`).

## Review findings

- **Logic:** The new branch is in the right place: after editable/interactive-target ownership checks and before the printable-character fallback. It does not steal keystrokes from another input, shortcut, or interactive control.
- **Behavior:** The existing `focus-only` path does not call `preventDefault()` and does not insert the reported key, which is required for dead-key composition.
- **Scope/design:** The production change is minimal and reuses the existing IME handling path. No unnecessary abstraction or unrelated behavior change was added.
- **Tests:** The added resolver and composable tests cover both the returned intent and the absence of insertion/default prevention. The test additions are proportionate to the regression.
- **Breaking changes:** None found.

## Verification

- Reviewed the delta from `67880ae32c577eed21e23dda65352e3336c9dc01` to `093b9741c588f0e0376c3e3e73b3b702e35a4608`.
- `git diff --check` passed.
- Targeted Vitest could not be executed in the detached review worktree because its dependencies are not installed there; the root worktree's test runner does not include the detached worktree files. This is an environment limitation, not a code finding.

## References

- `src/renderer/src/features/chat-page/model/typeToFocus.ts:53-65`
- `src/renderer/src/features/chat-page/composables/useComposerTypeToFocus.ts:42-45`
- `test/renderer/features/chat-page/model/typeToFocus.test.ts`
- `test/renderer/features/chat-page/composables/useComposerTypeToFocus.test.ts`

PR #2332, opened 2026-09-20
docs(plugin): explain credential reset on update

ct-jaryn:
A changed user-plugin package digest drops saved credential bindings even when the endpoint and credential declaration stay the same. The Baizhi example currently mentions only destination changes, so a documentation-only package update can unexpectedly leave users needing to enter their key again.

Update both English and Chinese lifecycle instructions to describe this behavior, point users to DeepChat's native credential form, and leave the intended core update behavior open for maintainer discussion. This follows the preservation-path explanation in review 5244917446 on #2324; it changes only two README files and targets dev.

Validation on dev 51528f822fb0aa44b06529019ea37df6508a2aa1: format:check, i18n:validate, lint, both typechecks, plugin:validate, and the existing userPluginLifecycle suite (8 tests) passed. README files are excluded by the repository formatter; both texts were manually checked. No live service or Electron GUI run was performed for this documentation change.

## Summary by CodeRabbit

- **Documentation**
  - Updated plugin update instructions to require reviewing package changes and approving new snapshots.
  - Clarified that package updates may remove saved credential bindings, even when endpoint and authentication settings are unchanged.
  - Added guidance to check plugin setup status and re-enter the original key when requested.
  - Added corresponding updates to the Chinese documentation.

PR #2336, opened 2026-09-20
docs: clarify Baizhi update credential scope

ct-jaryn:
A documentation-only Baizhi plugin update can request credential setup again, while the stored secret and `serverId` remain intact. This corrects the “drops stored credential bindings” wording in both READMEs and follows up on the review in #2332.

The lifecycle test now changes only the ZIP's README, installs the changed snapshot and reloads the plugin. It confirms that the MCP declaration, server identity and wrapped secret are preserved, but the versioned `PLUGIN_ROOT` / `CLAUDE_PLUGIN_ROOT` environment paths change `mcpVariableBindingScope`. `getMcpVariableBindings` consequently returns no usable values, and the plugin setup status reports `BAIZHI_API_KEY` as missing again. Re-entering the synthetic key restores the binding. The existing identical-ZIP and uninstall checks remain covered.

No runtime or credential-scope behavior changes. The test uses the real `UserPlugins`, `McpSettings` and `SecretStore` classes with mocked storage/encryption and connection supervision; it does not verify OS encryption or contact the hosted service.

Validation on Node 24.21.0 and pnpm 10.34.5:

- `pnpm run format:check`
- `pnpm run i18n:validate`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm exec vitest run --config vitest.config.ts test/main/plugin/userPluginLifecycle.test.ts test/main/plugin/userPlugins.test.ts test/main/mcp/settings.test.ts` — 43 tests passed
- `pnpm run plugin:validate --plugin-root examples/user-plugins/baizhi-agent-toolkit` — no findings

Prepared with AI assistance as a follow-up to my Baizhi integration contribution.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that plugin updates may require credentials to be configured again, including updates that only change documentation.
  - Saved credentials remain stored locally, but users may need to check setup status and re-enter the original key if prompted.
  - Updated guidance in English and Chinese to explain the update, credential, and uninstall workflow.


zerob13 (approved):
## Review: docs: clarify Baizhi update credential scope

**Verdict: Approve.** Documentation-only PR (2 example READMEs + 1 test extension, +51/−14, no runtime changes). The new wording is factually correct against the current implementation, and the test now pins exactly what the docs claim. No blocking or should-fix findings; two observations below.

### What this PR does

The Baizhi example READMEs previously said a package update "drops stored credential bindings" and flagged this as an open upstream question. That description was wrong in one important way: the stored secret is **not** deleted. What actually happens is that the plugin's versioned install path (`PLUGIN_ROOT` / `CLAUDE_PLUGIN_ROOT`, which contain the package digest) changes on every update, and that path is part of the hash used to validate credential bindings (`mcpVariableBindingScope` in `src/main/mcp/environmentBindings.ts:16`). The stored secret still exists under the same `serverId`, but the scope check fails, so the plugin reports the credential as missing and asks the user to re-enter it.

This PR corrects the wording to describe that mechanism accurately ("the saved credential remains in local storage, but the updated plugin's configuration no longer matches its binding") and extends the existing lifecycle test to prove it: a documentation-only ZIP update preserves `serverId`, endpoint, headers, and the stored secret, but changes the binding scope, reports `BAIZHI_API_KEY` as missing, and recovers after re-entering the key.

### Verified

- Read the diff and traced the mechanism in `src/main/plugin/userPlugins.ts` (`activate`, the versioned-root `env` injection) and `src/main/mcp/settings.ts` (`getMcpVariableBindings` scope check). The README claims match the code path for path.
- Ran `pnpm exec vitest run --config vitest.config.ts test/main/plugin/userPluginLifecycle.test.ts` on this branch: 8/8 passed.

### Observations (non-blocking)

1. **The old "open upstream question" caveat is gone.** The previous wording explicitly noted that invalidating credentials on a docs-only update may be unintended. The new text describes the behavior neutrally. The mechanism is now precisely documented, so this is defensible — but the underlying design question (should the binding-scope hash include transient install paths at all, when every package update therefore forces credential re-entry?) is still unresolved upstream. Worth keeping on the radar as a possible future fix; if it is ever fixed, the test added here will need updating, which is fine for a characterization test.
2. **The test grew to cover three scenarios in one `it()`** (identical-ZIP replacement, docs-only update, uninstall). It stays cohesive and reuses the same fixture lifecycle, so I would not split it — just noting the consolidation is intentional and reasonable.

### Detailed references

- `examples/user-plugins/baizhi-agent-toolkit/README.md:88-100` — updated Update bullet, new wording.
- `examples/user-plugins/baizhi-agent-toolkit/README.zh-CN.md:68-72` — zh-CN counterpart, consistent with the English text.
- `test/main/plugin/userPluginLifecycle.test.ts:146` — renamed test; docs-only update block at `:230-266`.
- `src/main/mcp/environmentBindings.ts:16-33` — `mcpVariableBindingScope` hashes `config.env`, which includes the digest-versioned `PLUGIN_ROOT`.
- `src/main/plugin/userPlugins.ts:369-411` — `activate()` rebuilds the MCP config (new root paths) whenever `record.mcpDigests[key] !== record.digest`, i.e. on any package change.
- `src/main/mcp/settings.ts:972-994` — `getMcpVariableBindings` returns `{}` when the stored scope no longer matches; the secret itself remains stored under `mcpVariableBindings.${serverId}`.

